PR #3, opened 2015-04-09
Housekeeping

gkoz:


GuillaumeGomez:
I'm gonna split out pango this evening. We should also change headers of files (that's a lot of code...). Thanks for that !

gkoz:
The headers, right :)
But they shouldn't take a lot of work. Hopefully we can automate the process.

GuillaumeGomez:
Yep, but it remains annoying ^^'

gkoz:
As we're breaking everything anyway, this seems to be a good occasion to follow Rust's style and make feature flags lowercase.

GuillaumeGomez:
I agree !

PR #7, opened 2015-04-11
Add missing file

GuillaumeGomez:


src/timeout_func.rs
```diff
@@ -0,0 +1,32 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with rgtk.  If not, see <http://www.gnu.org/licenses/>.
+
+pub mod timeout {
+    use ffi;
+
+    pub fn add<T>(interval: u32, func: fn(&T) -> i32, data: &T) -> u32 {
+        let tmp = data as *const T;
```
gkoz: This pointer will in most cases become invalid very soon after `add` returns. You need a proper boxed closure here.

src/timeout_func.rs
```diff
@@ -0,0 +1,32 @@
...
+        let tmp_f = func as ffi::gpointer;
```
gkoz: `func` isn't `extern "C"` so you're asking for trouble here. There's a reason `signal.rs` in `gtk` uses trampolines.

GuillaumeGomez: The point here is to test something on a little project of mine, depending of the results, I'll change a lot of things. ;)

PR #13, opened 2015-04-14
Make Value own the GValue struct

gkoz:


glib-sys/src/lib.rs
```diff
@@ -17,11 +17,11 @@
 
 extern crate libc;
 
-use libc::{c_void, c_int, c_uint, c_float, c_double, c_char, c_uchar, c_long, c_ulong};
+use libc::{c_void, c_int, c_uint, c_float, c_double, c_char, c_uchar, c_long, c_ulong, size_t};
 
 pub type GQuark = u32;
 
-pub type GType = c_ulong;
+pub type GType = size_t;
```
GuillaumeGomez: Where did you find this information ?

glib-sys/src/lib.rs
```diff
@@ -34,6 +34,12 @@ pub type GCallback = extern "C" fn();
 pub type GClosureNotify = extern "C" fn(data: gpointer, closure: gpointer);
 
 #[repr(C)]
+pub struct C_GValue {
+    typ: GType,
```
GuillaumeGomez: typ ?

glib-sys/src/lib.rs
gkoz: This is my current understanding of
```C
#if     GLIB_SIZEOF_SIZE_T != GLIB_SIZEOF_LONG || !defined __cplusplus
typedef gsize                           GType;
#else   /* for historic reasons, C++ links against gulong GTypes */
typedef gulong                          GType;
#endif
```
This change is a noop for unix-like architectures which are LP64 but affects Windows x64 where `sizeof(long) == 4`, so we might want to make extra sure this works there before merging. ;)

glib-sys/src/lib.rs
gkoz: Can make this `type_` I guess. Prefixing underscores has a meaning in Rust so I'm against that.

glib-sys/src/lib.rs
gkoz: And `GType` indeed is 8 bytes on win64 with both GTK SDKs I have, so `ulong` would've been a problem there.

gkoz:
Fixed the field name.

GuillaumeGomez:
I'll merge it but I'd prefer `type_` than `typ`. Can you change that please ?

gkoz:
Yeah, it's done already.

GuillaumeGomez:
Perfect ! Thanks for your work !

PR #15, opened 2015-04-16
Update license headers and authors list

gkoz:


GuillaumeGomez:
It might seem a bit stupid but why `rust-GNOME` instead of `rust-gnome` ?

gkoz:
GNOME is their official project name and all lowercase seemed ugly in its own way...

gkoz:
We're capitalizing all the words in these cases ("The ... Project Developers") to make it look super important :). In a usual text it might just be 'rust-gnome'.

GuillaumeGomez:
I don't have any preference on that so let's say we'll keep it that way. Thanks for the "making it look super important" ! ;-)

PR #19, opened 2015-05-01
Make travis build the examples repo to check for intercrate breakage

gkoz:
cc https://github.com/rust-gnome/gtk/issues/50


GuillaumeGomez:
oh right ! We talked about that a long time ago...

gkoz:
I've just got around to log into travis for the first time and see what's what hehe

PR #21, opened 2015-05-07
Add from_raw API stability workaround

gkoz:


GuillaumeGomez:
Perfect. Is there anything else who need to be changed to be compatible with beta ?

gkoz:
A similar change in `gtk` and that's it, blue skies =)

PR #22, opened 2015-05-07
Travis: build on both nightly and beta

gkoz:
Now that we've attained beta compatibility let's keep it that way :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: rust
+rust:
+- nightly
+- beta
 sudo: true
 before_install:
   - yes | sudo add-apt-repository ppa:ricotz/testing


GuillaumeGomez:
On that point, I don't agree with you. I thin it's the moment to create a new branch which will target beta. Targetting both nightly and beta at the same time doesn't seem to be a good idea. What do you think ?

gkoz:
Let's talk costs and benefits.
Branching today gives us a theoretically more mature repo organization and almost nothing in practice. It will require twice the work of merging the PRs with occasional merging conflicts happy-fun-time.
Keeping master compatible currently takes the least work. The downside so far is limited to little mostly cosmetic `feature(alloc/unique)` workarounds and we don't depend on any unstable Rust features.
When a PR making meaningful use of unstable features is submitted, this equation would change. As it stands we're good with a single branch. And this isn't without precedent: a number of crates are supporting beta and nightly from one branch currently.

GuillaumeGomez:
That's right, I forgot again about the fact we don't have unstable stuff anymore... What a glorious day ! =D

For now I'll merge your PR, but we'll have to talk again about this when the 1.0 will be released.

gkoz:
I'll open the PRs on other repos too then?

GuillaumeGomez:
Please ;-)

PR #27, opened 2015-05-09
Object reform

gkoz:


src/translate.rs
```diff
@@ -61,26 +61,22 @@ pub trait Ptr: Copy {
 }
 
 impl <T> Ptr for *const T {
-    fn is_null(&self) -> bool {
-        (*self).is_null()
-    }
+    #[inline]
+    fn is_null(&self) -> bool { (*self).is_null() }
```
GuillaumeGomez: Not sure that was super useful to reduce to one line but I guess it doesn't really matter...

src/translate.rs
```diff
@@ -157,7 +163,7 @@ impl <'a> ToGlibPtr<'a, *const c_char> for str {
 impl <'a> ToGlibPtr<'a, *const c_char> for String {
     type Storage = CString;
 
-    fn borrow_to_glib(&'a self) -> Stash<*const c_char, String> {
+    fn to_glib_none(&self) -> Stash<'a, *const c_char, String> {
         let tmp = CString::new(&self[..]).unwrap();
```
GuillaumeGomez: &self wasn't enough here ?

src/translate.rs
gkoz: I guess, I've picked up this style from somewhere in the libstd or similar and it stuck.. :)

GuillaumeGomez:
If I understand well, my investigations about the TreeViewModel was totally useless... Very nice job !

src/translate.rs
GuillaumeGomez: Like I said it doesn't matter, I have a personal preference for expanded code but who cares ? ^^

GuillaumeGomez:
@gkoz: Tell me when you feel ready so that I start merging all your work.

src/translate.rs
gkoz: `String` is a simple case, it's converting from references to objects that gave me a headache. And for a long time I didn't know how to avoid this lifetime here.

gkoz:
This stuff compiles on my end, so I dunno. We could ask interested people to review the ownership transfers for correctness. Or just merge it.

> If I understand well, my investigations about the TreeViewModel was totally useless...

Well... the ownership transfer bugs likes that are still possible if we put `from_glib_full` instead of `from_glib_none` somewhere and until we start generating the `-sys` crates from GIR the compiler can't help us.

gkoz:
> This stuff compiles on my end

To be clear, it works only if you merge all four PRs.

GuillaumeGomez:
That's why I asked you. If you're ready then I merge. Here we go !

PR #28, opened 2015-05-09
Fix the 32-bit build by truncating the microseconds value

gkoz:


gkoz:
The truncation may easily cause bugs though...

GuillaumeGomez:
Hum... Yes it could since it's supposed to be 64 bits...

gkoz:
I think we don't need `g_usleep` at all and the add method could be implemented in Rust (BTW its argument should've been signed).

gkoz:
Another option is to implement `from/to_glib` for `TimeVal` <-> `Duration`.

PR #33, opened 2015-05-12
FFI types update

gkoz:


gkoz:
These PRs will probably have to be merged despite failing builds and after that one of the builds could be restarted to check if everything is still in order.

GuillaumeGomez:
Ok, let's merge everything then.

PR #43, opened 2015-06-04
Add documentation

GuillaumeGomez:


src/error.rs
```diff
@@ -37,12 +49,19 @@ impl Error {
         }
     }
 
+    /// Does nothing if self.pointer is NULL; if self.pointer is non-NULL, then *self.pointer must be NULL.
+    /// A new GError is created and assigned to *self.pointer .
```
gkoz: Discussions of pointers are likely to confuse. Unfortunately, simply copypasting the documentation doesn't always make sense.

src/date.rs
```diff
@@ -81,6 +81,8 @@ pub struct Date {
 }
 
 impl Date {
+    /// Allocates a GDate and initializes it to a sane state. The new date will be cleared
```
gkoz: Just an observation: ideally these types would not exist and the foreign ones would be converted to/from one of the Rust date/time types.

GuillaumeGomez: For now they do exist, so I guess it's better (for now) to keep them.

src/error.rs
GuillaumeGomez: I didn't but this one (and certainly others) escaped from my attention.

PR #49, opened 2015-06-16
Implemented Debug, Display, and Error for glib::Error

### LINKED ISSUE #48: Add getting the error message from a GError
I don't think there's a way to get the message from a GError in safe code right now. I need to add this and add a Debug implementation for `glib::Error` so I can use it in a `Result`.

I can implement a `message() -> &str` function to access the field, if that's all that needs to be done.

gsingh93:
Fixes #48.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,6 +2,10 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+use std::str;
+use std::ffi::CStr;
+use std::fmt::{self, Formatter, Debug, Display};
+use std::error;
 use ffi::{self, GQuark};
 use glib_container::GlibContainer;
 use translate::ToGlibPtr;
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -59,12 +63,18 @@ impl Error {
...
+    ///
     /// Note that self is no longer valid after this call. If you want to keep using the same Error
     /// struct, you need to set it to NULL after calling this function on it.
     pub fn propagate(&mut self, other: &Error) -> () {
         unsafe { ffi::g_propagate_error(&mut self.pointer, other.pointer) }
     }
+
+    /// Returns the error message stored in the wrapped GError
+    pub fn message(&self) -> &str {
+        let c_str = unsafe { CStr::from_ptr((*self.pointer).message) };
```
gkoz: We can't guarantee that the `message` lives as long as `&self`. For example, the `set` method will invalidate the pointer.

gkoz:
I believe now that the borrow checker will be able to enforce safety.
Let's merge this @GuillaumeGomez

GuillaumeGomez:
I merge it then ! Thanks for your work @gsingh93 !

PR #51, opened 2015-06-24
ffi bindings for GHashTable

nicokoch:
This commit introduces the native bindings for GHashTable.
Some functions have been omitted, namely the ones taking an "unsafe
extern "c" fn" as an argument, and those dealing with GHashTableIter.
Also added a rudimentary test case.

I can also try to implement the corresponding interface in the main glib crate (although, I'm not really sure how stable the HashMap API really is).

### DIFF
--- a/glib-sys/src/lib.rs
+++ b/glib-sys/src/lib.rs
@@ -24,6 +24,8 @@ pub type GSourceFunc = extern "C" fn(user_data: gpointer) -> gboolean;
 pub type GCallback = extern "C" fn();
 pub type GClosureNotify = extern "C" fn(data: gpointer, closure: gpointer);
 pub type GDestroyNotify = extern "C" fn(data: gpointer);
+pub type GHashFunc = unsafe extern "C" fn(v: gconstpointer) -> c_uint;
+pub type GEqualFunc = unsafe extern "C" fn(v1: gconstpointer, v2: gconstpointer) -> gboolean;
 
 #[repr(C)]
 pub struct GAppInfo;
@@ -98,6 +100,9 @@ pub struct GDate;
     /// the day of the day-month-year representation of the date
     pub year: u8
 }*/
+
+#[repr(C)]
+pub struct GHashTable;
 
 //=========================================================================
 // GType constants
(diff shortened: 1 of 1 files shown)


gkoz:
Can you share your use case for `HashTable`? It's possible that a binding in the main crate and most of these functions may not be needed because the container conversions from the `translate` module would be used anyway. It's like this with lists: there're broken bindings for them but they're not used and I plan to remove them.

nicokoch:
In another project I'm doing I needed GHashTable to bind to another c library.

I'm not really that familiar with glib and gobjects and totally fine with using a ```translate``` solution. (Could you provide me with a basic usage example?)

GuillaumeGomez:
@nicokoch: Just by curiosity: which project are you talking about ?

You can find usage examples by doing a research on `to_glib_none`.

PS: Sorry but if it's not needed in at least one of rust/gnome repository, we won't merge it.

nicokoch:
I'm currently writing rust bindings for libsecret as a personal project. It is nowhere near presentable and thus not on github. For example [secret_attributes_build](https://developer.gnome.org/libsecret/unstable/libsecret-Secret-Attributes.html#secret-attributes-build) returns a HashTable.
Since you are one of the gnome guys: Is there currently anyone working on a libsecret-rs type project?

GuillaumeGomez:
Not as far as I know. Maybe @gkoz has heard about it ?

gkoz:
I have not heard of any efforts to bind that library.

@nicokoch From a cursory look at the API you need a way to pass back and forth a `HashMap<String, String>`. This could be achieved by implementing `FromGlibPtr` and `ToGlibPtr`, hiding all the gory details of ownership management.
Binding the `secret_attributes_build` function in particular may not be needed at all.

gkoz:
The definitions in this PR would be helpful for that (not all of them though) but the utility of the included test is dubious. We should rather test the user facing APIs.

gkoz:
There's a speedbump though: I expect that implementation to need `CString`'s [`into_ptr`](http://doc.rust-lang.org/nightly/std/ffi/struct.CString.html#method.into_ptr)/[`from_ptr`](http://doc.rust-lang.org/nightly/std/ffi/struct.CString.html#method.from_ptr) which aren't getting stabilized sooner than in Rust 1.2.

gkoz:
> There's a speedbump though: I expect that implementation to need CString's into_ptr/from_ptr, which aren't getting stabilized sooner than in Rust 1.2.

A workaround is to copy the `CString` implementation into our crate (a have a patch for that now).

nicokoch:
Thanks for your suggestions! Very helpful.

Implementing this will take a little bit longer than I initially thought, since I'm pretty busy at the moment. Would you be willing to merge a `FromGlibPtr` and `ToGlibPtr` implementation for HashMap or do you prefer to keep stuff like that out of the library? (And if yes, should the unused API bindings be removed?)

gkoz:
If you remove the test from this PR we could merge it and then I'd implement the conversions between `HashTable` and `HashMap` -- and those could be tested (a test won't catch memory management mistakes though).

nicokoch:
Done

GuillaumeGomez:
Once the tests have passed, I merge (except if @gkoz see something else ?). Thanks for your work !

gkoz:
@nicokoch Could you squash the commits please?

nicokoch:
@gkoz done :thumbsup:

gkoz:
@nicokoch The `HashMap<String, String>` translations have been committed in #59.

nicokoch:
Very nice! My libsecret bindings are on [github](https://github.com/nicokoch/secret-rs) now, too. 
Currently I have a usecase where I need to convert a `Vec` of `Ref`s to a `GList`. Unfortunately, I cannot get it to work with the newest commits. Rustc tells me that GList does not implement Copy. Any ideas?

gkoz:
> Currently I have a usecase where I need to convert a `Vec` of `Ref`s to a `GList`.

The conversions will only work with slices for now, so you'll have to coerce the vector to a slice with `&*` [like here on line 31](https://github.com/rust-gnome/gtk/commit/91983ed1db40150371be5b9688d6ce26502aad41#diff-d9c837c484468651e832dadec268dfe4).

BTW you shouldn't need to use `object::Ref`s except when implementing `object::Wrapper`.

PR #52, opened 2015-07-02
Implement ToGlibPtr and FromGlibPtrContainer for HashMap<String, String>

gkoz:
Also remove `IterToGlibPtr`.


gkoz:
This and https://github.com/rust-gnome/gtk/pull/112 depend on each other for the tests to pass :(

GuillaumeGomez:
Erf, then I just merge, no need to wait. Thanks for your work !

PR #55, opened 2015-08-13
Fix Date on Windows

gkoz:
I expect the tests to be green now.


GuillaumeGomez:
We'll see ! ;)

gkoz:
Turns out there's still is a problem with gdk and windows that needs to be fixed in the generator.

GuillaumeGomez:
Arf...

PR #65, opened 2015-10-06
Fix the soundness warnings

gkoz:


gkoz:
Looks like Rust is gradually becoming more verbose...

PR #67, opened 2015-10-15
Add wrapper macros and extend glib::translate

gkoz:


gkoz:
Just putting this out there.
Don't merge yet, I'd like to publish the crates first.

gkoz:
Modified the macro syntax and added object types support. Examples;

```rust

glib_wrapper! {
    /// Text buffer iterator
    pub struct TextIter(Boxed<ffi::GtkTextIter>);

    match fn {
        copy => |ptr| ffi::gtk_text_iter_copy(ptr),
        free => |ptr| ffi::gtk_text_iter_free(ptr),
    }
}

glib_wrapper!{
    /// Object representing an input device
    pub struct Device(Object<ffi::GdkDevice>);

    match fn {
        get_type => || ffi::gdk_device_get_type(),
    }
}
```

gkoz:
Might also add some syntax for `Upcast` impls.

GuillaumeGomez:
The code is getting more and more difficult to understand but its usage is improving quite a lot. However, with this approach, we have to "redeclare" all the specialized method, right ?

gkoz:
You had to put these in somehow before as well but it was more boilerplate:
```rust
impl types::StaticType for Device {
    fn static_type() -> types::Type { unsafe { from_glib(ffi::gdk_device_get_type()) } }
}
```
Now the `glib` crate is taking care of most gory details. It has to be macros because the newtypes must be defined in the crates that implement them.

gkoz:
Expanded the object wrapper macro to support upcasts.

gkoz:
This can be merged now if there are no objections.

GuillaumeGomez:
There is no such things. Really nice work, thanks !

PR #74, opened 2015-11-17
Make FFIGObject::wrap_object unsafe

gkoz:


GuillaumeGomez:
Wo. That's a big move here. Why this change and why now ?

gkoz:
The reason is mostly laziness because marking the functions unsafe stops the `check_init_asserts` script in gtk from complaining :)
On a more serious note the `wrap` method can't possibly be safe. Anyway this trait's days are numbered, I'll be very happy to delete it after finishing the object reform.

GuillaumeGomez:
That's what I thought: it was supposed to be deleted, so why specifying it as unsafe ? I'm glad that it's still the case. :)

gkoz:
We should merge this to let the gtk PR pass the build.

GuillaumeGomez:
Merged.

PR #77, opened 2015-11-25
Wrappers evolution

gkoz:
* Add `Refcounted` wrapper.
 * Implement `Object` wrapper via `Refcounted`.
 * Change placement of the lifetime in `ToGlibPtr` for saner implementations.
 * Shift some implementations burden from generics to macros.
 * Upcasts are implicit now.

"Reformed" branches: https://github.com/gtk-rs/gdk/pull/84 and [gtk](https://github.com/gkoz/gtk-rs-gtk/tree/object_reform_gir_wrappers2)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - beta
+  - nightly
```
GuillaumeGomez: Maybe we should just add beta too once and for all. Like this, all three stages will be tested. That's not very useful but whatever...

GuillaumeGomez:
Except for build targets, all seems ok. That's a very impressive work, thanks a lot for this !

.travis.yml
GuillaumeGomez: Same here.

gkoz: Yeah I guess there's no harm in having beta too.

GuillaumeGomez:
Perfect, examples will need to be updated too. Wanna do it ?

gkoz:
I don't think examples are affected, the build breaks because the other two PRs and this one are mutually dependent.

GuillaumeGomez:
Ok, then let's merge ! :D

PR #82, opened 2015-11-27
Add `Send` bound and panic guard to `idle_add` and `timeout_add`

gkoz:


withoutboats:
What is the reason that these functions need a `Send` bound?

gkoz:
Correct. Sorry they aren't documented properly, this is on our agenda.

PR #87, opened 2016-01-13
Create `trait Cast` with generic `upcast` and `downcast` methods

gkoz:
Remove `Upcast::upcast` making it a marker trait.
`Downcast` is not considered a general purpose trait, it's an implementation detail now.


gkoz:
As usual, interlocked changes will fail the tests...
See https://github.com/gtk-rs/examples/pull/57 for half of motivation for the change (the same applies to `upcast`).

GuillaumeGomez:
I'm really in love of this new object reform.

GuillaumeGomez:
No problem detected, nice work @gkoz! I merge.

PR #93, opened 2016-01-19
Add `new_with` to `Value`

pablocouto:
Hi, Just wrote this for my own convenience. Submitting in case you find it useful.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -23,6 +23,13 @@ impl Value {
         Value { inner: mem::zeroed() }
     }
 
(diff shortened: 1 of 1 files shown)


gkoz:
I'd rather these bindings were redesigned in a safe manner.

GuillaumeGomez:
Hi @pablocouto. Unfortunately, we can't merge this because it isn't basically in GTK and doesn't bring something gtk-rs would really need (I mean, you can absolutely do this in your code). I'll close this PR.

Don't hesitate to come back if you want to help and contribute! :)

Issue #73, opened 2015-11-16
GValue bindings are completely broken


gkoz:
This isn't quite at the top of my todo list so I'm inclined to mark all methods as unsafe pending a rewrite.

GuillaumeGomez:
Said like this, it's so much better ! Haha. Well, it's not a priority for now. We can use them through other values if someone really need it (like I do for process-viewer).

PR #95, opened 2016-01-22
Overhaul `GValue` bindings

gkoz:
Here's the best compromise between type safety and ergonomics I've managed to achieve.
cc #73 

Examples:

```rust
    let mut value = glib::Value::from(Some("Hello world !"));
    println!("gvalue.get example: {}", value.typed::<String>().unwrap().get().unwrap());

    let mut int = glib::Value::from(1);
    int.typed_mut::<i32>().unwrap().set_some(&2);
    assert_eq!(int.get(), Some(2));

    for i in 0..10 {
        let iter = right_store.append(None);
        right_store.set_value(&iter, 0, &i.into());
    }
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,7 +58,7 @@ pub use self::permission::Permission;
 pub use self::source::{Continue, idle_add, timeout_add, timeout_add_seconds};
 pub use self::traits::FFIGObject;
 pub use types::Type;
-pub use value::{TypedValue, TypedValueMut, Value};
+pub use value::{TypedValue, Value};
 pub use self::date::{TimeVal, Time, Date, Year, Month, Weekday, Day};
 
 #[macro_use]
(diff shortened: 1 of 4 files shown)


gkoz:
Supporting gtk update: https://github.com/gkoz/gtk-rs-gtk/tree/value

GuillaumeGomez:
Missing re-exported value I guess.

gkoz:
I think it's done now. Can't easily run `cargo test` in travis because of older glib and `g_type_init` issue but added to appveyor.

gkoz:
All right. It's the [pthread linking errors](https://ci.appveyor.com/project/GuillaumeGomez/glib-l2j1a/build/1.0.73/job/khul6gxj5t72km19) again. I guess I'll remove the CI changes from this PR.

gkoz:
Writing the docs made me realize that borrowing the value to get `TypedValue` might not be the best idea. Instead `Value` could be converted to `TypedValue`, which would deref to `Value`.

gkoz:
Now `TypedValue` contains `Value` and derefs to it.

PR #103, opened 2016-02-08
Remove more remnants of old incomplete wrappers

gkoz:


GuillaumeGomez:
Thanks! Any others before I merge?

gkoz:
Ah, I forgot to actually delete all of the removed modules. Should be complete now.

PR #110, opened 2016-03-04
Document Continue and idle/timeout_add, export CallbackGuard

gkoz:


GuillaumeGomez:
Nice idea! I wait for checks and I merge.

PR #111, opened 2016-03-04
Implement ToGlib and FromGlib traits to cmp::Ordering

GuillaumeGomez:


src/translate.rs
```diff
@@ -213,6 +214,19 @@ impl ToGlib for Option<char> {
...
 
+impl ToGlib for Ordering {
+    type GlibType = i32;
+
+    #[inline]
+    fn to_glib(&self) -> i32 {
+        match *self {
+            Ordering::Less => -1,
+            Ordering::Equal => 0,
+            _ => 1,
```
gkoz: Um, could you spell out the last variant please? They deserve equal treatment ;)

GuillaumeGomez: I'll update the commit's comment as well to try to get redemption haha.

src/translate.rs
```diff
@@ -644,6 +658,19 @@ impl FromGlib<u32> for char {
...
+    #[inline]
+    fn from_glib(val: i32) -> Ordering {
+        if val < 0 {
+            Ordering::Less
+        } else if val > 0 {
+            Ordering::Greater
+        } else {
+            Ordering::Equal
+        }
+    }
```
gkoz: Want a clever hack? ;) You can also say `val.partial_cmp(&0)`

GuillaumeGomez: Nice to know but I prefer to keep the current code. :) (it's slower)

gkoz: Actually now that I've thought a bit more it's not really a hack: `<` and `>` are sugar for `partial_cmp` anyway :)

GuillaumeGomez: Yes. ^^'

gkoz: I forgot `unwrap` though, `partial_cmp` returns `Option`. It wouldn't be that pretty after all.

GuillaumeGomez: Ah! :D

PR #118, opened 2016-03-21
Add extended glib_wrapper syntax for cross-crate inheritance

gkoz:


src/object.rs
```diff
@@ -232,9 +232,73 @@ macro_rules! glib_object_wrapper {
...
+
+        impl $crate::object::IsA<$super_name> for $name { }
+    };
+
+    (@munch_impls $name:ident, $super_name:path, $($implements:tt)*) => {
+        glib_object_wrapper!(@munch_impls $name, $super_name);
+        glib_object_wrapper!(@munch_impls $name, $($implements)*);
+    };
+
+    (@munch_impls $name:ident, $super_name:path => $super_ffi:path, $($implements:tt)*) => {
```
EPashkin: Not safer place it before L288?

gkoz: I don't expect any ambiguity: in L288 `$super_name` is followed by `,` and here by `=>`.

EPashkin: Missed that.

src/wrapper.rs
```diff
@@ -123,26 +139,25 @@ macro_rules! glib_wrapper {
...
     ) => {
-        glib_shared_wrapper!([$($attr)*] $name, $ffi_name, @ref $ref_arg $ref_expr,
-            @unref $unref_arg $unref_expr);
+        glib_object_wrapper!([$($attr)*] $name, $ffi_name, @get_type $get_type_expr, []);
     };
 
     (
         $(#[$attr:meta])*
-        pub struct $name:ident(Object<$ffi_name:path>);
+        pub struct $name:ident(Object<$ffi_name:path>): [$($implements:tt)+];
```
EPashkin: [] really needed?
May be change to something like:
```
pub struct Application(Object<ffi::GtkApplication>): 
         gio::Application(gio_ffi::GApplication),
         gio::ActionGroup(gio_ffi::GActionGroup),
         gio::ActionMap(gio_ffi::GActionMap);
```

gkoz: `[]` lets me capture its insides as a set of `tt`, that can be processed later. It's difficult if not impossible to work with a variable number of tokens somewhere other than at the right side of a token tree. `[]` provides a separate tree, without `[]` I'd get ambiguity with `;`.

So, without `[]` I'd have to make each entry use the extended syntax. Maybe that's not a big downside. Also note that `path` must be followed by either of `=>` `,` `=` `|` `;` `:` `>` `[` `{` `as` `where`.

EPashkin: Thank for explanation.

EPashkin: If I not wrong again then  `pub struct Bin(Object<ffi::GtkBin>): Container, Widget, Buildable;` not usable now and
example in need be fixed also.

gkoz: Since travis is green I assume the old syntax keeps working.

gkoz: I could add a case to gir so it wouldn't generate `[]` if they aren't needed.

EPashkin:
Thanks.

EPashkin:
Merge then.

GuillaumeGomez:
Good for me as well. Thanks @gkoz!

PR #128, opened 2016-06-11
Add stop_emission()

Susurrus:
Adds the stop_emission() function. I believe the function signature and code is correct, but seems very straightforward to me, so may be wrong.

I haven't gotten this integrated into my code yet as I don't know how to get the `signal_id` or the `detail` argument yet, but I figured I'd post it now and report back once I've tried it out a bit.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -8,6 +8,7 @@ use libc::{c_void, c_uint};
 
 use gobject_ffi::{self, GCallback};
 use glib_ffi::GQuark;
+use object::{IsA, Object};
 use source::CallbackGuard;
 use translate::ToGlibPtr;
 
@@ -23,6 +24,12 @@ pub unsafe fn connect(receiver: *mut gobject_ffi::GObject, signal_name: &str, tr
 pub fn signal_stop_emission<T: IsA<Object>>(instance: &T, signal_id: u32, detail: GQuark) {
     unsafe {
         gobject_ffi::g_signal_stop_emission(instance.to_glib_none().0, signal_id as c_uint, detail);
+    }
+}
+
+pub fn signal_stop_emission_by_name<T: IsA<Object>>(instance: &T, signal_name: &str) {
+    unsafe {
+        gobject_ffi::g_signal_stop_emission_by_name(instance.to_glib_none().0, signal_name.to_glib_none().0);
     }
 }
 
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -4,9 +4,10 @@
 
 //! `IMPL` Low level signal support.
 
-use libc::c_void;
+use libc::{c_void, c_uint};
 
 use gobject_ffi::{self, GCallback};
+use glib_ffi::GQuark;
 use source::CallbackGuard;
 use translate::ToGlibPtr;
 
@@ -17,6 +18,12 @@ pub unsafe fn connect(receiver: *mut gobject_ffi::GObject, signal_name: &str, tr
         gobject_ffi::GConnectFlags::empty()) as u64;
     assert!(handle > 0);
     handle
+}
+
+pub fn signal_stop_emission<T: IsA<Object>>(instance: &T, signal_id: u32, detail: GQuark) {
+    unsafe {
+        gobject_ffi::g_signal_stop_emission(instance.to_glib_none().0, signal_id as c_uint, detail);
+    }
 }
 
 unsafe extern "C" fn destroy_closure(ptr: *mut c_void, _: *mut gobject_ffi::GClosure) {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,6 +85,10 @@ pub use object::{
     Object,
     ObjectExt,
 };
+pub use signal::{
+    signal_stop_emission,
+    signal_stop_emission_by_name
+};
 pub use source::{
     CallbackGuard,
     Continue,
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -1,4 +1,4 @@
-//! Traits and essential types inteded for blanket imports.
+//! Traits and essential types intended for blanket imports.
 
 pub use {
     Cast,


EPashkin:
IMHO better implement `g_signal_stop_emission_by_name`.

Susurrus:
@EPashkin I've started using `stop_emission_by_name()`, which I think is supposed to be called in a callback and isn't a global call done outside of the callback, but the Gtk docs aren't clear on this. Maybe you could clear this up?

Regardless I've tried to run this on my code, but it fails when I do the following:
```rust
extern crate glib;
extern crate gtk;

use glib::object::Object;
use gtk::prelude::*;
use glib::signal::stop_emission_by_name;

let text_view = gtk::TextView::new();
let buffer = text_view.get_buffer().unwrap();
stop_emission_by_name(buffer.upcast::<Object>(), "GtkTextBuffer::insert-text");
```

And now I end up with errors like:
```
$ cargo build
   Compiling gattii v0.1.0
src/main.rs:196:34: 196:40 error: the trait bound `glib::Object: gtk::StaticType` is not satisfied [E0277]
src/main.rs:196     stop_emission_by_name(buffer.upcast::<Object>(), "GtkTextBuffer::insert-text");
                                                 ^~~~~~
src/main.rs:196:34: 196:40 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:196:34: 196:40 error: the trait bound `glib::Object: glib::wrapper::UnsafeFrom<glib::object::ObjectRef>` is not satisfied [E0277]
src/main.rs:196     stop_emission_by_name(buffer.upcast::<Object>(), "GtkTextBuffer::insert-text");
                                                 ^~~~~~
src/main.rs:196:34: 196:40 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:196:34: 196:40 error: the trait bound `glib::Object: glib::wrapper::Wrapper` is not satisfied [E0277]
src/main.rs:196     stop_emission_by_name(buffer.upcast::<Object>(), "GtkTextBuffer::insert-text");
                                                 ^~~~~~
src/main.rs:196:34: 196:40 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:196:34: 196:40 error: the trait bound `gtk::TextBuffer: gtk::IsA<glib::Object>` is not satisfied [E0277]
src/main.rs:196     stop_emission_by_name(buffer.upcast::<Object>(), "GtkTextBuffer::insert-text");
                                                 ^~~~~~
src/main.rs:196:34: 196:40 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:196:34: 196:40 help: the following implementations were found:
src/main.rs:196:34: 196:40 help:   <gtk::TextBuffer as gtk::IsA<gtk::Object>>
src/main.rs:196:27: 196:52 error: mismatched types:
 expected `*mut glib::object::GObject`,
    found `glib::Object`
(expected *-ptr,
    found struct `glib::Object`) [E0308]
src/main.rs:196     stop_emission_by_name(buffer.upcast::<Object>(), "GtkTextBuffer::insert-text");
                                          ^~~~~~~~~~~~~~~~~~~~~~~~~
src/main.rs:196:27: 196:52 help: run `rustc --explain E0308` to see a detailed explanation
error: aborting due to 5 previous errors
error: Could not compile `gattii`.

To learn more, run the command again with --verbose.
```

I think I'm calling it correctly. Though using a string to identify the signal name that isn't checked at compile time is very not-rusty. Qt supports compile-time signal checking, which I found invaluable as having signal connection errors is very easy to do without compile-time checking. Though maybe this improvement should come after this initial implementation.

Susurrus:
Also, `appveyor` is running tests on Rust 1.6 that seem to fail unrelated to my implementation (something about menu.rs using an unstable feature). Maybe the minimum rust version needs to be upped and appveyor updated accordingly? Travis is only checking on 1.9+, and gtk only checks on 1.9+ as well.

EPashkin:
Instead upcast try `.to_glib_none().0`
Yes we need update same in Gtk. Can you add separate PR for this?

EPashkin:
Also IMHO `stop_emmission_by_name` don't need be `unsafe`. What you think, @gkoz ?

Susurrus:
@EPashkin `.to_glib_none()` isn't implemented for TextBuffer. Am I misunderstanding what you said?

gkoz:
No I don't think they need to be unsafe. Generally, since these functions are intended for the users, their APIs should be high-level, like the rest. E.g.

```
pub fn stop_emission_by_name<T: IsA<Object>>(instance: &T, signal_name: &str)
```

For the other function this means adding a conversion from `str` to `GQuark` and creating an opaque type for the signal id. But is it worth the effort when there's a `by_name` alternative?

EPashkin:
Agree that `stop_emission` unneeded if `stop_emission_by_name` working with just signal names without details.

EPashkin:
Also agree that function need generalization.
If we remove `stop_emission` then `stop_emission_by_name` better take its name

gkoz:
Well it's possible someone will want the real `stop_emission` in the future so I'd rather not rename.

Susurrus:
Here's an updated version that provides a little bit nicer interface. I'm still getting the same errors as before about the trait bound gtk::TextBuffer: glib::IsA<glib::Object> not being satisfied with the following code example:
```rust
extern crate glib;
extern crate gtk;

use glib::object::Object;
use gtk::prelude::*;
use glib::signal::stop_emission_by_name;

let text_view = gtk::TextView::new();
let buffer = text_view.get_buffer().unwrap();
stop_emission_by_name(&buffer, "GtkTextBuffer::insert-text");
```

According to the docs TextBuffer implements the `IsA<Object>` trait. Am I missing something here?

gkoz:
I can't reproduce the error, seems all right to me. Could you please also reexport the functions in `lib.rs`, we only keep the "implementation detail" ones in modules.

src/signal.rs
```diff
@@ -4,9 +4,11 @@
 
 //! `IMPL` Low level signal support.
 
-use libc::c_void;
+use libc::{c_void, c_uint};
 
 use gobject_ffi::{self, GCallback};
+use glib_ffi::GQuark;
+use ::object::{IsA, Object};
```
gkoz: Leading `::` is redundant in imports, they always are absolute paths.

Susurrus:
Fixed the import path, but I don't understand what you mean about re-exporting in lib.rs. It's already exported through the `signal` module, or did you want me to actually move the implementation for these two functions into lib.rs directly?

gkoz:
Oh I've actually missed another thing. Look at [this piece of code](https://github.com/gtk-rs/glib/blob/master/src/lib.rs#L88-L94). All actual bindings functions we define keep their whole names except for the the `g_`, `gtk_` etc. prefix (e.g. `signal_stop_emission_by_name`) and are reexported (`pub use`) at the top level of their crates. Organizing naming hierarchically like it's customary in Rust would've complicated code generation and confused experienced GTK+ users.

Susurrus:
@gkoz I think I've fixed it. Can you tell me if this now looks right? Right now if I import `gtk::prelude::*` then I still can't use `signal_stop_emission_by_name()` without an additional import statement. And you can't use `glib::prelude::*` as that fails due to conflicting with `gtk::prelude::*`, but I can import it directly. Is that the correct/ideal behavior?

Susurrus:
But I also have been unable to use this function, here's a complete project:

Cargo.toml:
```toml
[dependencies]
gtk = { git = "https://github.com/susurrus/gtk" }
glib = { git = "https://github.com/susurrus/glib" }
```

src/main.rs:
```rust
extern crate glib;
extern crate gtk;

use gtk::prelude::*;
use glib::signal_stop_emission_by_name;

fn main() {
    let text_view = gtk::TextView::new();
    let buffer = text_view.get_buffer().unwrap();
    signal_stop_emission_by_name(&buffer, "insert-text");
}
```

The output of this simple project is:
```
$ cargo build
   Compiling test_signal v0.1.0 (~/Projects/test_signal)
src/main.rs:10:5: 10:33 error: the trait bound `gtk::TextBuffer: glib::IsA<glib::Object>` is not satisfied [E0277]
src/main.rs:10     signal_stop_emission_by_name(&buffer, "insert-text");
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/main.rs:10:5: 10:33 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:10:5: 10:33 note: required by `glib::signal_stop_emission_by_name`
error: Could not compile `test_signal`.

To learn more, run the command again with --verbose.
```

EPashkin:
I confirmed that error present in new project, will look at it this evening.
Strange, but if you add this code in `examples\gtktest.rs` it compiles fine.
May be problem with many `[[package]] name = "glib"` in `Cargo.lock`
and you need `replace` all gtk modules.
Or just get pull it all in same superfolder as test project and place `.cargo\config` in it with something like
```
paths = [
 "d:/eap/rust/0/sys/glib-sys",
 "d:/eap/rust/0/sys/gobject-sys",
 "d:/eap/rust/0/sys/gio-sys",
 "D:/eap/rust/0/gio",
 "D:/eap/rust/0/glib",
 "D:/eap/rust/0/cairo",
 "D:/eap/rust/0/cairo/cairo-sys-rs",
 "d:/eap/rust/0/sys/pango-sys",
 "D:/eap/rust/0/pango",
 "d:/eap/rust/0/sys/gdk-sys",
 "d:/eap/rust/0/sys/gdk-pixbuf-sys",
 "D:/eap/rust/0/gdk-pixbuf",
 "D:/eap/rust/0/gdk",
 "d:/eap/rust/0/sys/gtk-sys",
 "D:/eap/rust/0/gtk",
]
```

EPashkin:
May be just one path "D:/eap/rust/0/glib" do work.

EPashkin:
It really compiles with only `glib` path override in `.cargo\config`
```
.cargo
  config
glib (from github.com/susurrus/glib)
test
```
with deps in `Cargo.Toml`
```
[dependencies]
gtk = { git = "https://github.com/gtk-rs/gtk" }
glib = { git = "https://github.com/gtk-rs/glib" }
```

Susurrus:
Is that because gtk pulls in glib as well? So once this PR is pulled this shouldn't be a problem anymore, correct?

EPashkin:
@Susurrus, yes

gkoz:
@GuillaumeGomez r+?

GuillaumeGomez:
Sorry, I wasn't following. Let me catch up before.

GuillaumeGomez:
No doc on it?

GuillaumeGomez:
Ignore my previous question, I got my answer. It's good for me. Really nice work @Susurrus!

Issue #135, opened 2016-08-19
GKeyFile is missing


hfiguiere:
GKeyFile is missing from the bindings, so a bit hard to use with Rust.

Should these be auto generated or it supposed to be done by hand?

EPashkin:
Currently glib contains no generated code, even if we add it only one function for loading files will be `load_from_file`, as others `load_from_*` can't be generated due current generator limitations:
- `load_from_file` and `load_from_data_dirs` return array
- `load_from_data` working with memory buffer.

If you want add it manually I can show generated file which can be used as template.

hfiguiere:
I have started doing it. Lot of manual work though.

https://github.com/hfiguiere/glib/tree/keyfile

EPashkin:
Good, I prepare generated file in 3 hour.
Minor mistakes:
- `struct KeyFile` by convention go to "key_file.rs"
- if you pass  `full_path=NULL` to `g_key_file_load_from_data_dirs`, function not return anything to it, so `from_glib_full(full_path)` just wrong

hfiguiere:
The fullpath argument is a char** I pass "&NULL" (pointer to NULL) so it will return a newly allocated string.

GuillaumeGomez:
&NULL is a constant pointer. Not sure this is a good idea.

hfiguiere:
````rust
let mut full_path: *mut libc::c_char = ptr::null_mut();
````

and then I pass ````&mut full_path```` (which is ````*mut *mut i8````)

Anyway, I could test the actual binding because it seems that just adding the git repo to my crate Cargo.toml I hit an error "error: native library `gobject` is being linked to by more than one package, and can only be linked to by one package". I'll figure that one later, I'm already late.

hfiguiere:
what should be modified is the ````*mut i8```` by being a new pointer to the string (it is ````null_ptr```` on input)

EPashkin:
I was wrong about  `full_path`.

Also not better make `load_from_data_dirs` return type `Result<PathBuf, Error>`?

EPashkin:
Generated code: https://gist.github.com/EPashkin/069e929501f37adbe31144fb927f6eb3

hfiguiere:
where do we find that generator?

EPashkin:
https://github.com/gtk-rs/gir

hfiguiere:
I have updated my branch with @EPashkin patches and mine.

Just "make gir" and cargo build.

[...]

PR #136, opened 2016-08-20
Add GKeyFile to the API

hfiguiere:


hfiguiere:
This should fix issue #135. See the discussion over there.

hfiguiere:
I currently use a non HEAD/master commit for gir as there is a PR gtk-rs/gir#265

hfiguiere:
This is causing the failure on TravisCI. When that PR lands, I'll change the gitsubmodule commit to address this.

GuillaumeGomez:
All this Makefile and submodule things aren't very encouraging. Also, we won't merge anything as long as CIs don't say 'ok'.

src/source.rs
```diff
@@ -6,7 +6,8 @@ use std::cell::RefCell;
 use std::mem::transmute;
 use std::process;
 use std::thread;
-use glib_ffi::{self, gboolean, gpointer};
+use ffi as glib_ffi;
+use ffi::{self, gboolean, gpointer};
```
EPashkin:
```
warning: unused import, #[warn(unused_imports)] on by default
  --> src\source.rs:10:11
   |
10 | use ffi::{self, gboolean, gpointer};
   |           ^^^^
```

Gir.toml
```diff
@@ -0,0 +1,35 @@
...
+]
+
+[[object]]
+name = "GLib.KeyFile"
+status = "generate"
+       [[object.function]]
+       name = "load_from_data_dirs"
+       ignore = true
+       [[object.function]]
+       name = "load_from_dirs"
```
EPashkin: Its not in manual part?
Compiling example looks like this:
```rust
   pub fn load_from_dirs<T: AsRef<std::path::Path>, U: AsRef<std::path::Path>>(&self, file: T, search_dirs: &[U], flags: KeyFileFlags) -> Result<path::PathBuf, Error> {
        unsafe {
            let search_dirs: Vec<&std::path::Path> = search_dirs.iter().map(|p| p.as_ref()).collect();
            let mut error = ptr::null_mut();
            let mut full_path: *mut libc::c_char = ptr::null_mut();
            let _ = ffi::g_key_file_load_from_dirs(self.to_glib_none().0,
                                                        file.as_ref().to_glib_none().0,
                                                        search_dirs.to_glib_none().0,
                                                        &mut full_path,
                                                        flags.to_glib(), &mut error);
            if error.is_null() {
                let path: String = from_glib_full(full_path);
                Ok(path::PathBuf::from(path))
            } else {
                Err(from_glib_full(error))
            }
        }
   }
```

src/key_file.rs
```diff
@@ -0,0 +1,38 @@
+
```
EPashkin: Add copyright lines, please

hfiguiere:
As I said, the CI will pass for you merge gtk-rs/gir#265, which is fine to wait on.

Gir.toml
hfiguiere: I was meaning to remove it from the Gir.toml, but then I'll do quite the opposite. Will update the PR

Makefile
```diff
@@ -0,0 +1,17 @@
+GIR = gir/target/bin/gir
+GIR_SRC = gir/Cargo.toml gir/Cargo.lock gir/build.rs $(shell find gir/src -name '*.rs')
+GIR_FILES = gir-files/Gtk-3.0.gir
```
gkoz: This would be `gir-files/GLib-2.0.gir`.

Gir.toml
```diff
@@ -0,0 +1,35 @@
+[options]
+girs_dir = "gir-files"
+library = "GLib"
+version = "2.0"
+min_cfg_version = "2.32"
+target_path = "."
+work_mode = "normal"
+
+generate = [
+    "GLib.KeyFileFlags",
```
gkoz: I think `GLib.KeyFileError` ought to be here as well.

Gir.toml
```diff
@@ -0,0 +1,35 @@
...
+status = "generate"
+       [[object.function]]
+       name = "load_from_data_dirs"
+       ignore = true
+       [[object.function]]
+       name = "load_from_dirs"
+       ignore = true
+       [[object.function]]
+       name = "error_quark"
+       ignore = true
```
gkoz: As long as the error enum is generated this bit should not be necessary.

Makefile
hfiguiere: doh. ! Will fix too

EPashkin:
Gir part merged

Gir.toml
gkoz: Oh, is the function filed under `GKeyFile` instead of `GKeyFileError`? This might need some workaround in gir.

hfiguiere: generated binding return a "Quark" whose import fail unless I reexport it from ffi::GQuark, but @EPashkin told me not to. So I did it manually.... To me is is the generator that should be fixed...

Gir.toml
hfiguiere: This is now in the PR

src/key_file.rs
hfiguiere: done too

Gir.toml
hfiguiere: Good catch, I added this one.

Gir.toml
gkoz: Since the error enum is generated correctly never mind, this `ignore` seems the least hairy solution.

hfiguiere:
My bad I'll fix the CI.

hfiguiere:
it is actually a codegen bug :-/

hfiguiere:
It need this change in gir
````
diff --git a/src/codegen/enums.rs b/src/codegen/enums.rs
index 2ef5175..efdad75 100644
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -134,10 +134,16 @@ impl FromGlib<ffi::{ffi_name}> for {name} {{
     if let Some(ref get_quark) = enum_.error_domain {
         let get_quark = get_quark.replace("-", "_");
         let has_failed_member = members.iter().any(|m| m.name == "Failed");
+        let glib_sys_module: &str;
+        if env.namespaces.glib_ns_id == namespaces::MAIN {
+            glib_sys_module = "ffi";
+        } else {
+            glib_sys_module = "glib_ffi";
+        }
 
         try!(version_condition(w, env, enum_.version, false, 0));
         try!(writeln!(w, "impl ErrorDomain for {name} {{
-    fn domain() -> glib_ffi::GQuark {{
+    fn domain() -> {glib_sys_module}::GQuark {{
         {assert}unsafe {{ ffi::{get_quark}() }}
     }}
 
@@ -146,7 +152,8 @@ impl FromGlib<ffi::{ffi_name}> for {name} {{
     }}
 
     fn from(code: i32) -> Option<Self> {{
-        {assert}match code {{", name = enum_.name, get_quark = get_quark, assert = assert));
+        {assert}match code {{", name = enum_.name, get_quark = get_quark, assert = assert,
+                      glib_sys_module = glib_sys_module ));
 
         for member in &members {
             try!(version_condition(w, env, member.version, false, 3));

````

So here is what I propose:

- I remove ````GLib.KeyFileError```` from the generate Gir.toml
- we get this merge
- I create a new for the above, and one for gir that will be needed to solve the above issue.

**Comment by EPashkin** [comment]
Agree with merge without `GLib.KeyFileError`.
`Gir` can also fixed in enum.rs. Not sure what better.
```diff
diff --git a/src/codegen/enums.rs b/src/codegen/enums.rs
index 2ef5175..e0c0f1c 100644
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -19,7 +19,8 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
         try!(writeln!(w, ""));
         try!(writeln!(w, "use ffi;"));
         if env.namespaces.glib_ns_id == namespaces::MAIN {
-            try!(writeln!(w, "use error::ErrorDomain;
+            try!(writeln!(w, "use ffi as glib_ffi;
+use error::ErrorDomain;
 use translate::*;"));
         } else {
             try!(writeln!(w, "use glib_ffi;
```

Gir.toml
hfiguiere: Actually see the comments below. It needs another fix in gir.

gkoz:
Well, I vote for the smaller diff ;)

hfiguiere:
Ok, I'll check that.

hfiguiere:
ok this is the updated PR

src/key_file.rs
```diff
@@ -0,0 +1,61 @@
...
+use KeyFile;
+
+impl KeyFile {
+
+    pub fn error_quark() -> ffi::GQuark {
+        unsafe {
+            ffi::g_key_file_error_quark()
+        }
+    }
```
gkoz: Oh, I guess we've overlooked this bit. The function doesn't need to be exposed, the generated enum whenever it gets added covers that.

src/lib.rs
```diff
@@ -130,9 +132,13 @@ pub mod shared;
 #[macro_use]
 pub mod object;
 
+pub use auto::*;
+mod auto;
+
 mod bytes;
 pub mod error;
 mod file_error;
+pub mod key_file;
```
gkoz: We keep the modules private unless they provide some extensions (implementation details) needed by other `gtk-rs` crates not reexported in `lib.rs`.

src/key_file.rs
hfiguiere: No. The problem is that gir generate code that doesn't build:
````
src/auto/key_file.rs:6:5: 6:10 error: unresolved import `Quark`. There is no `Quark` in the crate root [E0432]
src/auto/key_file.rs:6 use Quark;
                           ^~~~~
src/auto/key_file.rs:6:5: 6:10 help: run `rustc --explain E0432` to see a detailed explanation
````

Fixing gir should help getting rid of that manual wrapper.

File: src/lib.rs (hunk 8 above)
Comment (hfiguiere) [inline src/lib.rs]: this is fixed now.

**Comment by EPashkin** [comment]
:+1: for this state.
Complete remove `KeyFile::error_quark` can be done with adding `GLib.KeyFileError` and `KEY_FILE_DESKTOP_*`.
Or we can merge PR266 and do it in this PR.

**Comment by hfiguiere** [comment]
Updated PR with the constant.
Wrapped manually a function to take a path instead of a string.

File: Cargo.toml [hunk 9]
```diff
@@ -27,3 +28,7 @@ version = "0.3.1"
 [dependencies.gobject-sys]
 git = "https://github.com/gtk-rs/sys"
 version = "0.3.1"
+
+[features]
+v2_38 = [ "glib-sys/v2_38", "gio-sys/v2_38", "gobject-sys/v2_38" ]
+v2_40 = [ "v2_38", "glib-sys/v2_40", "gio-sys/v2_40" ]
```
EPashkin: Empty line at end, pls

Gir.toml
```diff
@@ -0,0 +1,39 @@
...
+       [[object.function]]
+       name = "save_to_file"
+       # file parameter needs to be a PathBuf
+       ignore = true
```
EPashkin: Empty line, pls

hfiguiere:
addressed the last two comments.

EPashkin:
@hfiguiere, as @gkoz or @GuillaumeGomez still not merged this PR, can you add `GLib.KeyFileFlags` and remove manual `error_quark` here?

hfiguiere:
no, because it is not solved by the current gir.... ie it doesn't build.

EPashkin:
Strange, PR266 merged and all builds fine for me.

gkoz:
I believe updating the gir submodule to the latest master fixes the `KeyFileError` issue.

hfiguiere:
I tried that before commenting. If it has changed since, maybe. But 13 hours ago it didn't fix it.

gkoz:
At least it's worked for me right now. But submodules can be a bit confusing sometimes.

hfiguiere:
I did pull master, rebuild the generated filed. Still not fixing the problem. Problem still in Gir and I'm not familiar with it, so I haven't dug into the problem yet. I thinks perfect is the enemy of good in that case as we are still better than last time.

I have already pushed by code to use the AP this PR provides.

gkoz:
Thanks for contributing! I'll sort out the minor things later then.
@GuillaumeGomez any objections?

hfiguiere:
Closed by way of #137

PR #139, opened 2016-11-04
Add signal::Inhibit

EPashkin:
We need `Inhibit` in other crates, see https://github.com/gtk-rs/gio/blob/master/Gir.toml#L53-L59


EPashkin:
@GuillaumeGomez, @gkoz,  builds passed, take look at this. I add PR's for Gtk and Gio after merge.

GuillaumeGomez:
Well, since it doesn't modify anything, I'll just merge it. Thanks!

PR #145, opened 2017-02-16
Added glib::source_remove()

hfiguiere:
I could do `glib::idle_add()` but not remove it, so I wrapped `glib::source_remove()`.

Actually I wonder if that one couldn't be automated, in wish case this PR would totally be wrong.

Thanks.

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -111,3 +111,16 @@ where F: FnMut() -> Continue + Send + 'static {
             Some(trampoline), into_raw(func), Some(destroy_closure))
     }
 }
+
+/// Removes the source with the given id `tag` from the default main context.
+///
+/// It is a programmer error to attempt to remove a non-existent source.
+/// Note: source id are reused.
+///
+/// For historical reasons, the native function always returns true, so we
+/// ignore it here.
+pub fn source_remove(tag: u32) {
+    unsafe {
+        glib_ffi::g_source_remove(tag);
+    }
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Then it might be interesting to wrap the return of `glib::idle_add()` into an id or something which doesn't provide access to the `u32` and make the `glib::source_remove()` function takes it.

EPashkin:
@hfiguiere instead make Id field public can you add `FromGlib` implementation?

hfiguiere:
Good idea. done.

src/source.rs
```diff
@@ -8,7 +8,26 @@ use std::process;
 use std::thread;
 use ffi as glib_ffi;
 use ffi::{gboolean, gpointer};
-use translate::ToGlib;
+use translate::{FromGlib, ToGlib};
```
EPashkin: `use translate::{from_glib, FromGlib, ToGlib};` and use `from_glib` without `Id::`

hfiguiere:
done.

also this will break gtk-rs. I need to submit a PR for that.

hfiguiere:
Not break. Just consistency.

EPashkin:
Thanks, @hfiguiere 
👍 from me

GuillaumeGomez:
Excellent, thanks for creating the new Id type. It'll help for clarity!

PR #146, opened 2017-02-26
Standard derive for SourceId.

hfiguiere:
I forgot the standard derivce on SourceId. Which is making it harder to use.

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -11,6 +11,9 @@ use ffi::{gboolean, gpointer};
 use translate::{from_glib, FromGlib, ToGlib};
 
 /// The id of a source that is returned by `idle_add` and `timeout_add`.
+///
+/// A value of 0 is a good default as it is never a valid source ID.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub struct Id(u32);
 
 impl ToGlib for Id {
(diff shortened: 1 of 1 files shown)


src/source.rs
```diff
@@ -11,6 +11,7 @@ use ffi::{gboolean, gpointer};
 use translate::{from_glib, FromGlib, ToGlib};
 
 /// The id of a source that is returned by `idle_add` and `timeout_add`.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
```
EPashkin: Default really needed?

hfiguiere: it is usefull. A default value of zero is good because it is an invalid source and used that way in glib.

EPashkin: Ok.

GuillaumeGomez: Then you should write it to avoid any further confusion.

hfiguiere:
Added comment.

Also move `from_glib()` into `unsafe{}`  as per https://github.com/gtk-rs/gtk/pull/455#discussion_r102380965

GuillaumeGomez:
Thanks! Please squash your commits now. ;)

@EPashkin: anything to add?

EPashkin:
All seems good for me.

GuillaumeGomez:
Perfect then! Just remaining the squash. If you don't know how to do it, take a look [here](https://blog.guillaume-gomez.fr/Git-tips/1/2).

hfiguiere:
commit squashed.

GuillaumeGomez:
Perfect, thanks!

PR #153, opened 2017-03-15
Fix array of string full transfer

EPashkin:
Fix https://github.com/gtk-rs/gtk/issues/469

@Susurrus check this please.

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -923,12 +923,15 @@ for Vec<T> {
         if num == 0 || ptr.is_null() {
             return Vec::new()
         }
+        let orig_ptr = ptr;
         let mut res = Vec::with_capacity(num);
         while !(*ptr).is_null() {
             res.push(from_glib_full(*ptr));
             ptr = ptr.offset(1);
         }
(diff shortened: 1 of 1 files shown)


EPashkin:
@GuillaumeGomez, please look https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L1017-L1019
and https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L1077-L1079
I'm wrong or no way condition will be false?

GuillaumeGomez:
Well:

```rust
if num == 0 || ptr.is_null() {
    return Vec::new()
}
```

So impossible to be null, indeed.

EPashkin:
Fixed

GuillaumeGomez:
Thanks!

EPashkin:
Strange, appveyour do 2 builds per "action".

GuillaumeGomez:
Travis too some times. Dark magic and stuff.

Susurrus:
Looks like this fixed my code example. Thanks!

PR #160, opened 2017-04-29
Fix FromGlibContainer double freeing

EPashkin:
Revert one change from #153
Was wrong replacing g_free with g_strfreev by two reason:
1) g_strfreev frees inner string that already freed
2) g_strfreev must used only for string arrays, but this implementation for any arrays

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -911,9 +911,7 @@ for Vec<T> {
             res.push(from_glib_full(*ptr));
             ptr = ptr.offset(1);
         }
-        //See g_resource_enumerate_children
-        //glib_ffi::g_free(orig_ptr as *mut _);
-        glib_ffi::g_strfreev(orig_ptr as *mut _);
+        glib_ffi::g_free(orig_ptr as *mut _);
         res
     }
 }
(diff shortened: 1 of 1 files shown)


EPashkin:
@Susurrus please check that this don't broke https://github.com/gtk-rs/gtk/issues/469

sdroege:
Might make sense to add a simple test for this, looks good and correct to me apart from that

EPashkin:
No idea how to test it :(

GuillaumeGomez:
How did you get the bug?

EPashkin:
See https://github.com/gtk-rs/examples/pull/119#issuecomment-296235239

sdroege:
@EPashkin You could just manually create such a NULL-terminated array of e.g. strings with glib_sys::g_malloc() and ::g_strdup(). When passing that in there it should crash.

EPashkin:
@sdroege, Thanks I don't even think to test it (forgot that it not Gtk and can be tested).

Susurrus:
gtk-rs/gtk#469 is still fixed with this code, but this doesn't seem to fix gtk-rs/examples#119. I'm getting new errors now, like:

  *  `(.:12663): Gtk-CRITICAL **: GtkBox 0x7f8893b191a0 returned a widget path for type (null), but child is GtkScrolledWindow`
  * `(.:12663): Gtk-CRITICAL **: gtk_style_provider_get_style_property: assertion 'g_type_is_a (gtk_widget_path_get_object_type (path), pspec->owner_type)' failed`
  * `(.:12809): GLib-GObject-WARNING **: g_object_weak_unref: couldn't find weak ref 0x7ff1eb45c5e0(0x7ff1ece22ed8)`

But I don't get hard segfaults anymore with no warnings or errors, so I guess that's an improvement.

EPashkin:
@GuillaumeGomez ready for merge

GuillaumeGomez:
Thanks!

PR #162, opened 2017-05-02
Add support for getting a &str from a GValue without copying

sdroege:
Needs some discussion if the lifetime parameter is good or not here. For G_TYPE_STRING the copying or not might not be that important, but I'll also have to implement this for more complex types in the future where copying is expensive.

Without the added lifetime parameter, the following would compile just fine (while it shouldn't):
```rust
    let s = {
        let v = Value::from("test");
        assert_eq!(v.get::<&str>(), Some("test"));
        v.get::<&str>()
    };

    assert_eq!(s, Some("test"));
```

(Note that using get::<String>() instead above still works, as it should)

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -75,6 +75,7 @@ use std::fmt;
 use std::marker::PhantomData;
 use std::mem;
 use std::ops::{Deref, DerefMut};
+use std::ffi::CStr;
 
 use object::{Downcast, IsA, Object};
 use translate::*;
@@ -430,6 +431,17 @@ pub trait SetValue: StaticType {
 impl FromValueOptional for String {
     unsafe fn from_value_optional(value: &Value) -> Option<Self> {
         from_glib_none(gobject_ffi::g_value_get_string(value.to_glib_none().0))
+    }
+}
+
+impl<'a> FromValueOptional for &'a str {
+    unsafe fn from_value_optional(value: &Value) -> Option<Self> {
+        let cstr = gobject_ffi::g_value_get_string(value.to_glib_none().0);
+        if cstr.is_null() {
+            None
+        } else {
+            CStr::from_ptr(cstr).to_str().ok()
+        }
     }
 }
 
--- a/src/value.rs
+++ b/src/value.rs
@@ -349,6 +349,18 @@ impl<'a> From<&'a str> for TypedValue<String> {
     }
 }
 
+impl<'a> From<TypedValue<&'a str>> for TypedValue<String> {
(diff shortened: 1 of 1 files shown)


sdroege:
If this is accepted, I would also add something that can convert from a TypedValue<T: ToOwned> to a TypedValue<T::Owned> and back to allow easy switching between e.g. TypedValue<String> and TypedValue<&str>.

sdroege:
@GuillaumeGomez wrote on IRC that this seems ok to him. So next step: 1) fixing all implementations so the tests succeed, 2) implement something to convert between e.g. TypedValue<String> and TypedValue<&str> easily.

Any suggestions for the API for 2)? Or maybe just always work with TypedValue<String> but have a get_ref() on it if FromValueOptional is implemented for the types involved, and one implements Borrow for the other?

EPashkin:
For first part only gdk need fix.
We don't have other implementation currently. and we only need it in gdk as I unable to define implementations for some `BoxedValue` parallel with `IsA<Object>` implementations
👍 for merge this ignoring build error.

sdroege:
For 2) I don't see any useful way of doing that generically, because ToOwned/Borrowed work with "str" as type instead of "&str". We could define our own mapping traits here, which seems like quite some overhead, or just implement From<_> for the conversion between TypedValue<&str> and TypedValue<String> in both directions.

@GuillaumeGomez what do you think?

src/value.rs
```diff
@@ -427,12 +440,23 @@ pub trait SetValue: StaticType {
...
 }
 
+impl<'a> FromValueOptional<'a> for &'a str {
+    unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
+        let cstr = gobject_ffi::g_value_get_string(value.to_glib_none().0);
+        if cstr.is_null() {
+            return None;
+        }
+
+        CStr::from_ptr(cstr).to_str().ok()
```
GuillaumeGomez: Why not just an `else` condition and remove the `return` statement above?

GuillaumeGomez:
Just a small nit/suggestion but otherwise it's all good.

src/value.rs
sdroege: Done :) Just a matter of taste I guess, I prefer to have the "error" code paths outside the normal code path.

GuillaumeGomez:
@EPashkin: Ok for me, tell me when it's ok for you as well so I can merge.

EPashkin:
@GuillaumeGomez yes, IMHO in can be merged

GuillaumeGomez:
Great! Thanks!

PR #164, opened 2017-05-06
Crate

GuillaumeGomez:


EPashkin:
IMHO before this need update version in master branch

GuillaumeGomez:
You should add this into the `release_process.md` file so that master is always a version ahead of crate. So for this one, I think we'll just ignore this and make the PR on master afterwards.

EPashkin:
Ok.
Note that we ignored it on previous release too :( Master not be 2 version apart.

EPashkin:
Also IMHO version need be same and changed in same time.

GuillaumeGomez:
Well, if someone wants to override a version by specifying a git, I think it'll be simpler if master if always a version ahead. But otherwise I don't care much, we have to update the version number everytime anyway.

EPashkin:
👍

PR #167, opened 2017-05-11
Use g_object_ref_sink() instead of g_object_ref() everywhere

sdroege:
https://github.com/gtk-rs/gir/issues/367


EPashkin:
@GuillaumeGomez please take look at this PR and related issue.
IMHO it can be merged without problem

GuillaumeGomez:
The famous floating reference. Already spoke a bit about it with @sdroege. I agree with the change as well so let's merge it.

Thanks!

PR #169, opened 2017-05-17
Add glib::BoolError for use as return value of possibly failing functions returning booleans

sdroege:
https://github.com/gtk-rs/gir/issues/368

And example usage in utils::setenv(). Code generator changes come tomorrow when this is all ok.

How do you handle API changes?

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -113,3 +113,28 @@ pub trait ErrorDomain: Copy {
     /// i.e. any unrecognized codes map to it.
     fn from(code: i32) -> Option<Self> where Self: Sized;
 }
+
+/// Generic error used for functions that fail without any further information
+#[derive(Debug)]
+pub struct BoolError(&'static str);
+
+impl BoolError {
+    pub fn from_glib(b: glib_ffi::gboolean, s: &'static str) -> Result<(), Self> {
+        match b {
+            glib_ffi::GFALSE => Err(BoolError(s)),
+            _ => Ok(()),
+        }
+    }
+}
+
+impl fmt::Display for BoolError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl error::Error for BoolError {
+    fn description(&self) -> &str {
+        self.0
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,7 +81,7 @@ extern crate gobject_sys as gobject_ffi;
 
 use std::ffi::CStr;
 pub use bytes::Bytes;
-pub use error::Error;
+pub use error::{Error, BoolError};
 pub use file_error::FileError;
 pub use object::{
     Cast,
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -101,7 +101,7 @@ pub fn get_user_name() -> Option<String> {
 #[cfg(not(windows))]
 pub fn get_current_dir() -> Option<String> {
     unsafe {
-        from_glib_none(ffi::g_get_current_dir())
+        from_glib_full(ffi::g_get_current_dir())
     }
 }
 
@@ -113,6 +113,6 @@ pub fn get_current_dir() -> Option<String> {
     }
 
     unsafe {
-        from_glib_none(g_get_current_dir_utf8())
+        from_glib_full(g_get_current_dir_utf8())
     }
 }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -5,6 +5,7 @@
 use ffi;
 use translate::*;
 use std::path::PathBuf;
+use error::BoolError;
 
 /// Same as [`get_prgname()`].
 ///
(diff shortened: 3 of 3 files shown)


sdroege:
@EPashkin Is this also what you meant? Could make the struct field public, and discuss names for the constructor, but apart from that there don't seem to be many other alternatives.

FromGlib can't be implemented properly here because we need the additional &str argument.

EPashkin:
Yes I meant same, thanks
Also think that from_glib is not good.
Maybe rename it for 'check` or `try` ?

We have time to think about naming and usage anyway,
I prefer apply https://github.com/gtk-rs/gir/pull/371 and it effects first and it can take days.

sdroege:
`check` and `try` seem equally confusing. IMHO it should be some `from_XXX` for consistency with other API, as we create something *from* the arguments.

EPashkin:
`from` IMHO also confusing as it don't produce `BoolError`.
Maybe `to_result` ? But by convention it not static member :(

EPashkin:
@GuillaumeGomez that you think about this change?
If you accept, I merge https://github.com/gtk-rs/gir/pull/373.
@sdroege When Gir's PR merged, please generate all *env that currently ignored

GuillaumeGomez:
I have to admit that I don't see the point of having a `BoolError` type. It doesn't bring anything except for the function name. If you really want to not return a `bool` in case of failure, then let return `Option<bool>` or `Result<bool, ()>` directly.

If you think I missed something, I'm open to discussion so don't hesitate to bring the explanations. :p

EPashkin:
@GuillaumeGomez Main reason possibility of use `try`, see https://github.com/gtk-rs/gir/issues/368

GuillaumeGomez:
Then why create a whole new type? Just returning a `Result<bool, ()>` would be enough, no?

sdroege:
Just returning Result<(), ()> would work, yes. But it seems quite ugly, instead of some actual error type.

Also it's not just the try macro, but also the compiler warning about unchecked results.

sdroege:
Also with the function name in there you at least have some idea where it comes from when propagating the error further up and wrapping it in more meaningful error types

EPashkin:
Main problem that () not implement `Error`

EPashkin:
() don't implement `Error`

GuillaumeGomez:
Ok, your arguments are valid so now I agree. Thanks a lot for this PR!

@EPashkin: I let you merge gir then I guess this code will need to be regenerated?

EPashkin:
@GuillaumeGomez Yes

sdroege:
> @sdroege When Gir's PR merged, please generate all *env that currently ignored

Please merge the first commit (adding the BoolError type) at the same time, otherwise GIR will generate code that can't compile. I removed the g_setenv() commit from this pull request for now.

I'll open a new pull request that updates g_setenv() and others after the other parts are merged.

For g_*env(), you'd like to have them autogenerated instead of manually implemented? I can do that, I was wondering anyway why you implemented them manually (especially because they don't work only on Unix-like OS, but work just fine on Windows too).

EPashkin:
Just add commit here, with gir.toml, gir submodule and regen.

I get `src/utils.rs:39: undefined reference to `g_getenv'` when attempt of using it on Windows + MSYS2

sdroege:
I'll double check later, maybe it's a macro on Windows. I definitely used it on Windows from C code though and even the docs mention the Windows behaviour of it.

EPashkin:
On windows it just named differently: `g_getenv_utf8`.
Code was in https://github.com/GNOME/glib/blob/master/glib/genviron.h but seems it removed  this spring https://github.com/GNOME/glib/commit/b67d071321264cc4a7bb9df47a4381918232279a
So currently it better stays unix only.

EPashkin:
There many functions like that. We planed support naming difference sometime.
But if it replaced now situation worsened: we have old libraries that use "_utf8" suffix and new that don't :cry:

sdroege:
I'll investigate at home, going through code on the phone is not very pleasant

Glib changing API is a problem in any case, will talk to them. Thanks for pointing that out

sdroege:
@EPashkin You were misreading, it's all there still:
https://github.com/GNOME/glib/blob/master/glib/genviron.h#L36
https://github.com/GNOME/glib/blob/master/glib/genviron.c#L416

It's not just always going to the normal function and using UTF8 instead of the system's codepage. All good here

sdroege:
So what should happen about this? For GLib < 2.53 there is g_getenv() for Unix-like OS, g_getenv_utf8() for Windows (and g_getenv() on Windows that returns strings in the system codepage, which we don't want). For GLib >= 2.53, there's only g_getenv() in the headers and it always gives UTF8 (e.g. is equivalent to g_getenv_utf8() on Windows), and g_getenv_utf8() still exists on Windows (just not in the headers).

Maybe for the time being the following: Always expose a manually implemented g_getenv(), which on Windows calls g_getenv_utf8() (which is always available for the time being, maybe just not in the headers) and g_getenv() elsewhere. And in some far away future, we just call g_getenv() always once glib-rs depends on GLib >= 2.53

EPashkin:
Thanks for pointing that we still can use _utf8 even in new version.

sdroege:
This is now on top of https://github.com/gtk-rs/glib/pull/171

EPashkin:
Last commit IMHO unneeded as likely don't need regen

sdroege:
Yes, removed then :)

EPashkin:
@GuillaumeGomez CI passed, please do final check this PR

GuillaumeGomez:
All good for me. Thanks!

PR #170, opened 2017-05-18
Generate global functions

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/295
Apply https://github.com/gtk-rs/gir/pull/371


GuillaumeGomez:
Waiting for appveyor to finish and then I merge.

EPashkin:
Don't forget gtk's PR ;)

GuillaumeGomez:
Ah damn, you're right!

src/utils.rs
```diff
@@ -4,19 +4,6 @@
 
 use ffi;
 use translate::*;
-use UserDirectory;
-
-pub fn get_application_name() -> Option<String> {
```
sdroege: While all these functions are in auto/functions.rs, I don't see them re-exported from lib.rs anywhere

EPashkin: It can be used by long name

sdroege: But previously it was available as e.g. utils::get_application_name(). Now it would be auto::functions::get_application_name() (and the auto module is not pub-re-exported so that won't work)

EPashkin: Hm, you right. I missed fact that auto non public in all crates :(
What better do: add to lib.rs `pub mod functions` with reexports or something else?
@GuillaumeGomez maybe you have better idea?

EPashkin: Or just make `auto` public?

EPashkin: I also missed duplicate `gio::(un)resources_register` :(

GuillaumeGomez: I think we should do as follow: all standalone functions (so not methods of enums/structs) should be usable directly. As such, we should directly reexport `auto/functions`. (To be sure what I said was clear :p :

```rust
pub use functions::*;
```

So that we can call the functions like this:

```rust
gtk::some_func();
```
)

GuillaumeGomez: > I also missed duplicate `gio::(un)resources_register` :(

Arf. :'(

EPashkin: Added reexport to glib. Will add it to other crates tomorrow

EPashkin: Bad that rust don't even warn about duplicates on glob import :(

sdroege: I would (for now at least) just re-export the functions as they were before from src/utils.rs, e.g. `pub use auto::functions::get_application_name` in there.

GuillaumeGomez: @sdroege: Next release will be a breaking one. :p

sdroege: Fine with me then :) then my changes to convert bool returns to Result<(), BoolError> could also get in without headaches.
Are there other breaking changes on your list?

GuillaumeGomez: No, but just throw in whatever you have in mind (this is the good time!).

sdroege: Only https://github.com/gtk-rs/gir/issues/368 currently, nothing else I'm aware of at this point. We'll see. What do you target as release date?

Other API breaking change was mine to add a lifetime parameter to FromValue

GuillaumeGomez: We never target a specific date. When we judge we have enough changes or someone really want/need a release to publish his/her crate, we do it.

sdroege: Makes sense :) before a big release, it might be a good idea to do a small release with the floating reference bugfix, just of glib.

GuillaumeGomez: There is no such things as a small release (because it takes a lot of time to make one).

sdroege: It's not possible to just "cargo publish" glib with a new version number and that patch, and keep everything else the same?

GuillaumeGomez: Hum, might be actually.

EPashkin:
This is last for reexport

GuillaumeGomez:
And thanks!

PR #171, opened 2017-05-21
Add Windows implementations for g_setenv, g_getenv, g_unsetenv and g_get_current_dir

sdroege:
There are equivalent functions with the _utf8() prefix on Windows, to which macros with the "normal" names are pointing.

See https://github.com/gtk-rs/glib/pull/169#issuecomment-302931512 and below

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -101,7 +101,7 @@ pub fn get_user_name() -> Option<String> {
 #[cfg(not(windows))]
 pub fn get_current_dir() -> Option<String> {
     unsafe {
-        from_glib_none(ffi::g_get_current_dir())
+        from_glib_full(ffi::g_get_current_dir())
     }
 }
 
@@ -113,6 +113,6 @@ pub fn get_current_dir() -> Option<String> {
     }
 
     unsafe {
-        from_glib_none(g_get_current_dir_utf8())
+        from_glib_full(g_get_current_dir_utf8())
     }
 }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -31,14 +31,26 @@ pub fn set_prgname(name: Option<&str>) {
     }
 }
 
-#[cfg(unix)]
+#[cfg(not(windows))]
 pub fn getenv(variable_name: &str) -> Option<String> {
     unsafe {
         from_glib_none(ffi::g_getenv(variable_name.to_glib_none().0))
     }
 }
 
-#[cfg(unix)]
+#[cfg(windows)]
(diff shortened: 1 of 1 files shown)


sdroege:
@EPashkin Would probably make sense to merge this before https://github.com/gtk-rs/glib/pull/169 , or do https://github.com/gtk-rs/glib/pull/169 only with the new BoolError. And then afterwards setenv() can be changed to return a BoolError in another pull request. Too many interdependencies.

src/utils.rs
```diff
@@ -63,6 +101,19 @@ pub fn get_user_name() -> Option<String> {
 #[cfg(unix)]
```
EPashkin: something like not(cfg(windows))

sdroege: done

EPashkin:
@sdroege I don't like define ff-function in normal way, and prefer doing it later with gir starting with sys crate.
Thanks for pointing error in `get_current_dir` I missed this when checked generated code (also it better return `Option<std::path::PathBuf>`)

@GuillaumeGomez What you think?

sdroege:
>  don't like define ff-function in normal way, and prefer doing it later with gir starting with sys crate.

The problem is that you won't have it in the .gir file, unless you generate it on Windows. And then you won't have the non-UTF8 one. That's why I thought declaring it here would be best.

EPashkin:
As we don't have separate gir files for windows (using ubuntus),
we only can add for Gir's config parameter that add these function in sys mode,
and use it in normal mode.
But your can stays if GG agreed with it.

EPashkin:
Checked that `getenv` work on Windows with local (Cyrillic) characters

src/utils.rs
```diff
@@ -47,22 +60,61 @@ pub fn setenv(variable_name: &str, value: &str, overwrite: bool) -> bool {
...
+    use libc::c_char;
+    extern "C" {
+        fn g_get_current_dir_utf8() -> *mut c_char;
+    }
+
+    unsafe {
+        from_glib_full(g_get_current_dir_utf8())
+    }
+}
+
```
GuillaumeGomez: Extra empty line.

sdroege: Thanks, fixed

GuillaumeGomez:
It's a bit strange to have these C functions declared directly into the rust binding. However, it seems quite difficult to be able to do it otherwise (if it's not provided by the gir file, we have to do it by hand). So for me it's ok, I'm just a bit disappointed and hope that this won't be too much of a burden to maintain...

PR #172, opened 2017-05-21
Implement FromGlibPtrNone and FromGlibPtrFull for Value

sdroege:
There are some functions that need that in GStreamer.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -165,10 +165,8 @@ impl Value {
 impl Clone for Value {
     fn clone(&self) -> Self {
         unsafe {
-            // FIXME: make this safer by making GValue::g_type public
-            let type_ = *(&self.0 as *const gobject_ffi::GValue as *const glib_ffi::GType);
             let mut ret = Value::uninitialized();
-            gobject_ffi::g_value_init(ret.to_glib_none_mut().0, type_);
+            gobject_ffi::g_value_init(ret.to_glib_none_mut().0, self.0.g_type);
             gobject_ffi::g_value_copy(self.to_glib_none().0, ret.to_glib_none_mut().0);
             ret
         }
@@ -258,7 +256,6 @@ impl FromGlibPtrFull<*mut gobject_ffi::GValue> for Value {
         let mut ret = Value::uninitialized();
         ptr::swap(&mut ret.0, ptr);
         glib_ffi::g_free(ptr as *mut c_void);
-
         ret
     }
 }
--- a/src/value.rs
+++ b/src/value.rs
@@ -76,6 +76,8 @@ use std::marker::PhantomData;
 use std::mem;
 use std::ops::{Deref, DerefMut};
 use std::ffi::CStr;
+use std::ptr;
+use libc::c_void;
 
 use object::{Downcast, IsA, Object};
 use translate::*;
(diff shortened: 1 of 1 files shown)


EPashkin:
Maybe better add 2 other options but IMHO it can wait (don't like add unused functions)
@GuillaumeGomez 👍 for this

sdroege:
I can also add the other trait impls, might be good for consistency

src/value.rs
```diff
@@ -242,6 +242,25 @@ impl<'a> ToGlibPtr<'a, *mut gobject_ffi::GValue> for &'a [&'a ToValue] {
...
+        ret
+    }
+}
+
+impl FromGlibPtrFull<*mut gobject_ffi::GValue> for Value {
+    unsafe fn from_glib_full(ptr: *mut gobject_ffi::GValue) -> Self {
+        let mut ret = Value::uninitialized();
+        ptr::swap(&mut ret.0, ptr);
+        glib_ffi::g_free(ptr as *mut c_void);
+
```
GuillaumeGomez: I don't think this line is necessary. ;)

sdroege: True :)

sdroege:
So should I implement impls for the other traits (which?) or we just wait until someone needs it? :)

EPashkin:
IMHO it good as is.

antoyo:
I also implemented these traits in [my pull request](https://github.com/gtk-rs/glib/pull/173/files#diff-dd26c2766e65aca43e4228e64ebd0cf8R220).
I'm not sure which way is better but mine doesn't involve a copy (not sure if it is correct, thought).

By the way, are some implementions missing in this pull request?
I've got 4 implementations in mine.

sdroege:
@antoyo Sorry, I should've checked before. Didn't expect another pull request to do this as it's rather uncommon for C code to give you GValue pointers.

See my review comments on your code. I could add the `impl FromGlibPtrNone<*mut gobject_ffi::GValue> for Value` as that one is technically correct, but the `impl FromGlibPtrFull<*const gobject_ffi::GValue> for Value` is not. Anyway, see comments there :)

Should we finish your pull request and get that one merged because you were first, or just merge this one?

sdroege:
I added `impl FromGlibPtrNone<*mut gobject_ffi::GValue> for Value` too here. I guess let's get in my pull request for now, and @antoyo 's then only for the GClosure stuff (thanks btw, I will also need that soonish)

sdroege:
@antoyo in your case you probably also want to prevent the copying to happen (you just want to borrow the values from the arguments to the `marshal` function), for which it would make sense to add a lifetime parameter to FromGlibPtrNone (and maybe Full too?), and then implement it for `&'a Value` too

antoyo:
I don't have any problem on basing my pull request on this work (thanks for you work too ;) ).

I read about memory management in glib and I wonder if `FromGlibPtrNone` from glib is actually correct.
From what I understand, when there's no ownership transfer, it is similar to `&T` in Rust, so I believe `from_glib_none` should actually return a reference.
But I can see in the documentation that `FromGlibPtrNone<*const c_char> for String` is implemented and that copies the `String`.

I don't understand yet how all these traits work.
Do I need `FromGlibPtrBorrow` to get a reference to a value that I get with transfert none?
If yes, I would need the implementation of this trait in this pull request.

sdroege:
Basically what we would need for borrowing is `impl<'a> FromGlibPtrNone<*const gobject_ffi::GValue> for &'a Value` for example. Currently all implementations of `FromGlibPtrNone` are doing a copy (or another `ref()`), which is also correct. "transfer none" in GLib just means that you, as the caller of that function, don't own the return value. So taking a copy is a valid thing to do in that case, while with "transfer full" we could just take the return value without copying (but often can't because of different allocators, etc).

sdroege:
See also https://github.com/gtk-rs/glib/pull/162 for a similar case

sdroege:
@GuillaumeGomez Anything blocking this from being merged?

GuillaumeGomez:
No, we're all good. Thanks!

sdroege:
Great, thanks :)

PR #173, opened 2017-05-23
Add support for gobject closures

antoyo:
Hi.
I started to create the `Closure` type so that we can generate the code of methods taking a `GClosure` as a parameter.
This is a work in progress, but I open this pull request right now, which is working (even thought it contains a hack) so that you can comment on it.
I'll have to fix the hack for `size_of` by writing a C function to get the size of `GClosure`, unless someone has a better idea.

I think that marshallers support should be done in another pull request since I believe that doing something both safe and efficient will be non trivial.

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -1,0 +1,136 @@
+// Copyright 2013-2017, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+// TODO: support marshaller.
+
+use std::mem;
+use std::ptr;
+use std::slice;
+
+use libc::{c_uint, c_void};
+
+use gobject_ffi;
+use translate::{ToGlibPtr, ToGlibPtrMut, Uninitialized, from_glib_none};
+use types::Type;
+use Value;
+use ToValue;
+
+glib_wrapper! {
+    pub struct Closure(Shared<gobject_ffi::GClosure>);
+
+    match fn {
+        ref => |ptr| {
+            gobject_ffi::g_closure_ref(ptr);
+            gobject_ffi::g_closure_sink(ptr);
+        },
+        unref => |ptr| gobject_ffi::g_closure_unref(ptr),
+    }
+}
+
+impl Closure {
+    pub fn new<F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static>(callback: F) -> Self {
+        unsafe extern "C" fn marshal<F>(_closure: *mut gobject_ffi::GClosure, return_value: *mut gobject_ffi::GValue,
+            n_param_values: c_uint, param_values: *mut gobject_ffi::GValue, _invocation_hint: *mut c_void,
+            marshal_data: *mut c_void)
+            where F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static
+        {
+            let values = slice::from_raw_parts(param_values as *const _, n_param_values as usize);
+            let callback: Box<F> = Box::from_raw(marshal_data as *mut _);
+            let result = callback(values);
+            if !return_value.is_null() {
+                match result {
+                    Some(result) => *return_value = result.into_raw(),
+                    None => {
+                        let result = Value::uninitialized();
+                        *return_value = result.into_raw();
+                    },
+                }
+            }
+            mem::forget(callback);
+        }
+
+        unsafe extern "C" fn finalize<F>(notify_data: *mut c_void, _closure: *mut gobject_ffi::GClosure)
+            where F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static
+        {
+            let _callback: Box<F> = Box::from_raw(notify_data as *mut _);
+            // callback is dropped here.
+        }
+
+        unsafe {
+            let size = 4 + 4 + 3 * mem::size_of::<*mut c_void>() as u32;
+            let closure = gobject_ffi::g_closure_new_simple(size, ptr::null_mut());
+            assert_ne!(closure, ptr::null_mut());
+            let callback = Box::new(callback);
+            let ptr: *mut F = Box::into_raw(callback);
+            let ptr: *mut c_void = ptr as *mut _;
+            gobject_ffi::g_closure_set_meta_marshal(closure, ptr, Some(marshal::<F>));
+            gobject_ffi::g_closure_add_finalize_notifier(closure, ptr, Some(finalize::<F>));
+            from_glib_none(closure)
+        }
+    }
+
+    pub fn invoke(&self, values: &[&ToValue]) -> Option<Value> {
+        let mut result = unsafe { Value::uninitialized() };
+        let mut values: Vec<_> = values.iter()
+            .map(|v| v.to_value())
+            .collect();
+        let gvalues = values.as_mut_ptr() as *mut _;
+        unsafe {
+            gobject_ffi::g_closure_invoke(self.to_glib_none().0 as *mut _, result.to_glib_none_mut().0,
+                values.len() as u32, gvalues, ptr::null_mut());
+        }
+        if result.type_() == Type::Invalid {
+            None
+        } else {
+            Some(result)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    use super::Closure;
+    use ToValue;
+    use Value;
+
+    fn closure_fn(values: &[Value]) -> Option<Value> {
+        assert_eq!(values.len(), 2);
+        let string: Option<String> = values[0].get();
+        assert_eq!(string, Some("test".to_string()));
+        let int: Option<i32> = values[1].get();
+        assert_eq!(int, Some(42));
+        Some(24.to_value())
+    }
+
+    #[test]
+    fn test_closure() {
+        let call_count = Arc::new(AtomicUsize::new(0));
+
+        let count = call_count.clone();
+        let closure = Closure::new(move |values| {
+            count.fetch_add(1, Ordering::Relaxed);
+            assert_eq!(values.len(), 2);
+            let string: Option<String> = values[0].get();
+            assert_eq!(string, Some("test".to_string()));
+            let int: Option<i32> = values[1].get();
+            assert_eq!(int, Some(42));
+            None
+        });
+        let result = closure.invoke(&[&"test".to_string(), &42]);
+        assert!(result.is_none());
+        assert_eq!(call_count.load(Ordering::Relaxed), 1);
+
+        let result = closure.invoke(&[&"test".to_string(), &42]);
+        assert!(result.is_none());
+        assert_eq!(call_count.load(Ordering::Relaxed), 2);
+
+        let closure = Closure::new(closure_fn);
+        let result = closure.invoke(&[&"test".to_string(), &42]);
+        let int: Option<i32> = result.and_then(|result| result.get());
+        assert_eq!(int, Some(24));
+    }
+}
--- a/src/value.rs
+++ b/src/value.rs
@@ -153,7 +153,8 @@ impl Value {
         }
     }
 
-    fn into_raw(mut self) -> gobject_ffi::GValue {
+    #[doc(hidden)]
+    pub fn into_raw(mut self) -> gobject_ffi::GValue {
         unsafe {
             let ret = mem::replace(&mut self.0, mem::uninitialized());
             mem::forget(self);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -175,6 +175,7 @@ pub use auto::functions::*;
 mod auto;
 
 mod bytes;
+pub mod closure;
 pub mod error;
 mod enums;
 mod file_error;


src/closure.rs
```diff
@@ -0,0 +1,127 @@
...
+use Value;
+use ToValue;
+
+pub struct Closure {
+    closure: *mut gobject_ffi::GClosure,
+}
+
+impl Closure {
+    pub fn new<F: Fn(&[Value]) -> Option<Value> + 'static>(callback: F) -> Self {
+        unsafe extern "C" fn marshal<F: Fn(&[Value]) -> Option<Value> + 'static>(closure: *mut gobject_ffi::GClosure,
```
EPashkin: Rename parameter to `_closure` if it stay unused and `_invocation_hint` too

src/closure.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+            for value in values {
+                mem::forget(value);
+            }
+        }
+
+        unsafe {
+            // TODO: get the appropriate size with size_of (not working since the generated
+            // GClosure does not contain every field).
+            let closure = gobject_ffi::g_closure_new_simple(1000, ptr::null_mut());
```
EPashkin: IMHO `g_closure_new` need used with something like https://github.com/gtk-rs/glib/blob/master/src/signal.rs#L64-L68

EPashkin:
Just added minor tips, as I don't fully understand closures and don't sure how to test it in wild ;)

src/value.rs
```diff
@@ -215,6 +217,40 @@ impl Uninitialized for Value {
     }
 }
 
+impl FromGlibPtrNone<*mut gobject_ffi::GValue> for Value {
+    #[inline]
+    unsafe fn from_glib_none(ptr: *mut gobject_ffi::GValue) -> Self {
+        assert!(!ptr.is_null());
+        Value(ptr::read(ptr))
```
sdroege: This causes two owners of the same memory if the GValue is some kind of pointer (object, string, boxed type, ...). Both your new Value and the C GValue will contain the same pointer, and both think they own it and will try to free it later, leaving the other one with a pointer to a now invalid memory address.

src/value.rs
```diff
@@ -215,6 +217,40 @@ impl Uninitialized for Value {
...
+    }
+}
+
+impl FromGlibPtrFull<*mut gobject_ffi::GValue> for Value {
+    #[inline]
+    unsafe fn from_glib_full(ptr: *mut gobject_ffi::GValue) -> Self {
+        let value = Value(ptr::read(ptr));
+        gobject_ffi::g_value_unset(ptr);
+        value
```
sdroege: This one doesn't have the above problem, because the C code passed you ownership of whatever the GValue contains, but it also passed you ownership of the GValue itself. You need to free it afterwards, or leak memory.

src/value.rs
```diff
@@ -215,6 +217,40 @@ impl Uninitialized for Value {
...
+    #[inline]
+    unsafe fn from_glib_none(ptr: *const gobject_ffi::GValue) -> Self {
+        assert!(!ptr.is_null());
+        Value(ptr::read(ptr))
+    }
+}
+
+impl FromGlibPtrFull<*const gobject_ffi::GValue> for Value {
+    #[inline]
+    unsafe fn from_glib_full(ptr: *const gobject_ffi::GValue) -> Self {
```
sdroege: A `*const GValue` seems wrong here. C code is never going to mark a pointer a `const` if you're going to own it, as otherwise all mutating functions (and free(), g_value_unset(), etc) would cause a compiler warning (you can't pass const pointers to them).

src/value.rs
sdroege: Also it has actually the same problem, just slightly different. If you didn't have the g_value_unset(), it would be fine. But because of the g_value_unset(), for things stored as a pointer you will copy over the pointer and then immediately free it, leaving you with a pointer to a now invalid memory address.

src/closure.rs
```diff
@@ -0,0 +1,127 @@
...
+                values.push(value);
+                param_values = param_values.offset(1);
+            }
+            let callback: Box<Box<F>> = Box::from_raw(marshal_data as *mut _);
+            let result = callback(&values);
+            if let Some(result) = result {
+                *return_value = result.into_raw();
+            }
+            for value in values {
+                mem::forget(value);
```
sdroege: And this is why the FromGlib* impls work for you. Once fixed, you would have to remove the mem::forget() here and let the Drop impls take care of everything else

src/closure.rs
sdroege: No, that's fine as is. But the size of GClosure is 4 + 3 * sizeof(void*)

src/closure.rs
```diff
@@ -0,0 +1,127 @@
...
+            .map(|v| v.to_value())
+            .collect();
+        let mut gvalues: Vec<gobject_ffi::GValue> = values.iter()
+            .map(|v| unsafe { ptr::read(v.to_glib_none().0) })
+            .collect();
+        unsafe {
+            gobject_ffi::g_closure_invoke(self.to_glib_none_mut().0, value.to_glib_none_mut().0, gvalues.len() as u32,
+                gvalues.as_mut_ptr(), ptr::null_mut());
+        }
+        Some(value)
```
sdroege: This might be unsafe. I think you should a) initialize value with mem::zeroed(), and b) check here if value.g_type == G_TYPE_NONE and in that case return None. What you do here can AFAIU return a Value full of uninitialized memory.

src/closure.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+impl Drop for Closure {
```
sdroege: You can (and probably should) also implement Clone around g_closure_ref() here

src/closure.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        unsafe {
+            // TODO: get the appropriate size with size_of (not working since the generated
+            // GClosure does not contain every field).
+            let closure = gobject_ffi::g_closure_new_simple(1000, ptr::null_mut());
+            assert_ne!(closure, ptr::null_mut());
+            let callback = Box::new(Box::new(callback));
+            let ptr: *mut Box<F> = Box::into_raw(callback);
+            let ptr: *mut c_void = ptr as *mut _;
+            gobject_ffi::g_closure_set_meta_marshal(closure, ptr, Some(marshal::<F>));
```
sdroege: This will either leak the callback Box, or free it multiple times. g_closure_invoke() can be called multiple times.
What you could do is to install a finalize notifier (g_closure_add_finalize_notifier()) with the box, and destroy the box in there. And inside the marshaller you would always "forget" the box again at the end so it does not get destroyed.

sdroege: FWIW, your implementation frees it multiple times and uses the freed box

antoyo:
I think I fixed all mentionned issues (thanks for your review).
I'll use the work of pull request #172 when it's merged.

antoyo:
I made a quick test with @sdroege implementation and my closure implementation with `webkit2gtk-webextension` and it does not seem to work well:
the first time, the callback is called, but then, the webview freeze and obviously, the callback is not called anymore after that.
Perhaps it would be easier to debug with a function from gobject/glib.

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+            gobject_ffi::g_closure_add_finalize_notifier(closure, ptr, Some(finalize::<F>));
+            Closure {
+                closure: closure,
+            }
+        }
+    }
+
+    pub fn invoke(&mut self, values: &[&ToValue]) -> Option<Value> {
```
sdroege: This does not need &mut btw. It's safe (as long as your closure is!) to call invoke() from multiple threads at once

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+        let mut result: gobject_ffi::GValue = unsafe { mem::zeroed() };
+        let values: Vec<_> = values.iter()
+            .map(|v| v.to_value())
+            .collect();
+        let mut gvalues: Vec<gobject_ffi::GValue> = values.iter()
+            .map(|v| unsafe { ptr::read(v.to_glib_none().0) })
```
sdroege: To prevent copying, here it would be safe to just ptr::read(&v.0)

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl Closure {
+    pub fn new<F: Fn(&[Value]) -> Option<Value> + 'static>(callback: F) -> Self {
+        unsafe extern "C" fn marshal<F: Fn(&[Value]) -> Option<Value> + 'static>(_closure: *mut gobject_ffi::GClosure,
+            return_value: *mut gobject_ffi::GValue, n_param_values: c_uint, mut param_values: *mut gobject_ffi::GValue,
+            _invocation_hint: *mut c_void, marshal_data: *mut c_void)
+        {
+            let mut values = vec![];
+            for _ in 0 .. n_param_values {
+                let value: Value = FromGlibPtrNone::from_glib_none(param_values);
```
sdroege: And here to prevent copying, you could convert the `*mut GValue` to a `&Value` via pointer casting and dereference

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            let _callback: Box<Box<F>> = Box::from_raw(notify_data as *mut _);
+            // callback is dropped here.
+        }
+
+        unsafe {
+            let size = 4 + 4 + 3 * mem::size_of::<*mut c_void>() as u32;
+            let closure = gobject_ffi::g_closure_new_simple(size, ptr::null_mut());
+            assert_ne!(closure, ptr::null_mut());
+            let callback = Box::new(Box::new(callback));
```
sdroege: Why do you Box twice btw?

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+            return_value: *mut gobject_ffi::GValue, n_param_values: c_uint, mut param_values: *mut gobject_ffi::GValue,
+            _invocation_hint: *mut c_void, marshal_data: *mut c_void)
+        {
+            let mut values = vec![];
+            for _ in 0 .. n_param_values {
+                let value: Value = FromGlibPtrNone::from_glib_none(param_values);
+                values.push(value);
+                param_values = param_values.offset(1);
+            }
+            let callback: &Box<F> = mem::transmute(marshal_data as *mut _);
```
sdroege: You can also use Box::from_raw() here and just mem::forget() it before returning.

You should also create a glib::CallbackGuard() here to catch panics and exit on panics

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+            let result = callback(&values);
+            match result {
+                Some(result) => *return_value = result.into_raw(),
+                None => {
+                    let result = Value::uninitialized();
+                    let mut gvalue = result.into_raw();
+                    gvalue.g_type = gobject_ffi::G_TYPE_NONE;
```
sdroege: Value::uninitialized() already sets the type to G_TYPE_NONE

antoyo:
@sdroege You can test the `main` example in [the webkit2gtk-rs repository](https://github.com/antoyo/webkit2gtk-rs) with `cargo run --example --features v2_6`.
It requires that the example in [this repository](https://github.com/antoyo/webkit2gtk-webextension-rs/tree/feature/closure) to be built in debug mode (be sure to put both repositories at the same directory level so that [this path](https://github.com/antoyo/webkit2gtk-rs/blob/master/examples/main.rs#L36) works.
This will make the webview freeze after you clicked on the page: no more scroll or click working after that.

Since I used your `ToGlibPtr` implementations, this freeze cannot be caused by my implementation being unsound.

sdroege:
Seems to work fine for me, no freezes or anything

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl<'a> ToGlibPtr<'a, *const gobject_ffi::GClosure> for Closure {
+    type Storage = &'a Closure;
+
+    fn to_glib_none(&'a self) -> Stash<'a, *const gobject_ffi::GClosure, Self> {
+        Stash(self.closure, self)
+    }
+}
+
+impl<'a> ToGlibPtrMut<'a, *mut gobject_ffi::GClosure> for Closure {
```
sdroege: You probably also want to implement the FromGlibPtr* traits

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+            .collect();
+        unsafe {
+            gobject_ffi::g_closure_invoke(self.to_glib_none_mut().0, &mut result, gvalues.len() as u32,
+                gvalues.as_mut_ptr(), ptr::null_mut());
+        }
+        if result.g_type == gobject_ffi::G_TYPE_NONE {
+            None
+        }
+        else {
+            Some(unsafe { Value::from_glib_full(&mut result as *mut _) })
```
sdroege: This should work:

let mut result = Value::uninitialized();
...
gobject_ffi::g_closure_invoke(..., result.to_glib_none_mut().0, ...)
...
Some(value)

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+            .map(|v| v.to_value())
+            .collect();
+        let mut gvalues: Vec<gobject_ffi::GValue> = values.iter()
+            .map(|v| unsafe { ptr::read(v.to_glib_none().0) })
+            .collect();
+        unsafe {
+            gobject_ffi::g_closure_invoke(self.to_glib_none_mut().0, &mut result, gvalues.len() as u32,
+                gvalues.as_mut_ptr(), ptr::null_mut());
+        }
+        if result.g_type == gobject_ffi::G_TYPE_NONE {
```
sdroege: G_TYPE_INVALID, not G_TYPE_NONE

src/closure.rs
sdroege: Or alternatively something like `into_raw` in reverse (`from_raw`)

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+            let callback: &Box<F> = mem::transmute(marshal_data as *mut _);
+            let result = callback(&values);
+            match result {
+                Some(result) => *return_value = result.into_raw(),
+                None => {
+                    let result = Value::uninitialized();
+                    let mut gvalue = result.into_raw();
+                    gvalue.g_type = gobject_ffi::G_TYPE_NONE;
+                    *return_value = gvalue;
```
sdroege: You also want to check first `if !return_value.is_null()`

src/closure.rs
```diff
@@ -0,0 +1,163 @@
...
+        unsafe extern "C" fn finalize<F: Fn(&[Value]) -> Option<Value> + 'static>(notify_data: *mut c_void,
+            _closure: *mut gobject_ffi::GClosure)
+        {
+            let _callback: Box<Box<F>> = Box::from_raw(notify_data as *mut _);
+            // callback is dropped here.
+        }
+
+        unsafe {
+            let size = 4 + 4 + 3 * mem::size_of::<*mut c_void>() as u32;
+            let closure = gobject_ffi::g_closure_new_simple(size, ptr::null_mut());
```
sdroege: And you want to call g_closure_sink() after this. Floating references \o/  (fixes crashes, testcase coming in a minute)

sdroege:
Standalone testcase here, with fixed up (hopefully) Closure bindings: https://gist.github.com/sdroege/d47508b39f55bf454f90673150ccef20
Is still missing the optimizations to prevent unneeded copying though

src/closure.rs
EPashkin: Note: Instead `result.to_glib_none_mut().0`  better use `mut_override(result.to_glib_none().0)` in most cases

src/closure.rs
antoyo: Isn't it better to transmute to a reference to avoid having to `mem::forget`?
I believe both options has some drawbacks.

src/closure.rs
antoyo: I followed the style used by the code generated by `gir`, for instance in [`gtk::button`](https://github.com/gtk-rs/gtk/blob/master/src/auto/button.rs#L294).
I don't know the reason to double box: can @GuillaumeGomez or @EPashkin explain, please?

EPashkin: I don't remember clearly :(
Maybe for "safety" if outer box restored to broken object we get AV when try deref inner

sdroege:
Maybe use the glib_wrapper!() macro here with the ref/unref functions, it would automatically give you most of the trait impls. And instead of calling g_closure_ref() & g_closure_sink only in the constructor, the best would be to make that the ref function. I.e.
```Rust
glib_wrapper! {
     pub struct Closure(Shared<ffi::Closure>);

     match fn {
         ref => |ptr| {
             g_closure_ref(ptr);
             g_closure_sink(ptr);
         },
         unref => |ptr| ffi::g_closure_unref(ptr),
     }
 }
```

src/closure.rs
```diff
@@ -23,31 +23,31 @@ impl Closure {
             _invocation_hint: *mut c_void, marshal_data: *mut c_void)
         {
             let values = slice::from_raw_parts(param_values as *const _, n_param_values as usize);
-            let callback: &Box<F> = mem::transmute(marshal_data as *mut _);
+            let callback: Box<F> = Box::from_raw(marshal_data as *mut _);
```
sdroege: If you want to omit the mem::forget() below, you could also do `let callback: &F = mem::transmute(marshal_data)`

antoyo:
I've just tested the webkit example with the latest version of this pull request and it now works fine.

Cargo.toml
```diff
@@ -24,7 +24,7 @@ version = "0.3.4"
 
 [dependencies.gobject-sys]
 git = "https://github.com/gtk-rs/sys"
-version = "0.3.4"
+#version = "0.3.4"
```
EPashkin: Forgot to remove temporary change?

src/closure.rs
```diff
@@ -0,0 +1,157 @@
...
+
+use gobject_ffi;
+use translate::{Stash, StashMut, ToGlibPtr, ToGlibPtrMut, Uninitialized, mut_override};
+use types::Type;
+use Value;
+use ToValue;
+
+pub struct Closure {
+    closure: *mut gobject_ffi::GClosure,
+}
```
sdroege: glib_wrapper! {
     pub struct Closure(Shared<ffi::Closure>);

     match fn {
         ref => |ptr| {
             g_closure_ref(ptr);
             g_closure_sink(ptr);
         },
         unref => |ptr| ffi::g_closure_unref(ptr),
     }
 }

src/closure.rs
```diff
@@ -0,0 +1,157 @@
...
+            let callback = Box::new(callback);
+            let ptr: *mut F = Box::into_raw(callback);
+            let ptr: *mut c_void = ptr as *mut _;
+            gobject_ffi::g_closure_set_meta_marshal(closure, ptr, Some(marshal::<F>));
+            gobject_ffi::g_closure_add_finalize_notifier(closure, ptr, Some(finalize::<F>));
+            gobject_ffi::g_closure_ref(closure);
+            gobject_ffi::g_closure_sink(closure);
+            Closure {
+                closure: closure,
+            }
```
sdroege: from_glib_none(closure) (and remove the manual ref/sink calls)

src/closure.rs
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    pub fn invoke(&self, values: &[&ToValue]) -> Option<Value> {
+        let result = unsafe { Value::uninitialized() };
+        let mut values: Vec<_> = values.iter()
+            .map(|v| v.to_value())
+            .collect();
```
sdroege: If we would take &[Value] here, we could also prevent the last copy

src/closure.rs
```diff
@@ -0,0 +1,157 @@
...
+        let gvalues = values.as_mut_ptr() as *mut _;
+        unsafe {
+            gobject_ffi::g_closure_invoke(self.to_glib_none().0 as *mut _, mut_override(result.to_glib_none().0),
```
sdroege: Why mut_override() instead of result.to_glib_none_mut()?

src/closure.rs
sdroege: Drop, Clone, ToGlib*, FromGlib* would all be automatically implemented by using `glib_wrapper!`

sdroege:
Generally looks good to me now, just some further comments

src/closure.rs
EPashkin: Agree, mut_override don't need here as result is output parameter

src/closure.rs
```diff
@@ -0,0 +1,134 @@
+// TODO: support marshaller.
```
EPashkin: Please, don't forget add standard header

src/closure.rs
antoyo: @sdroege: You mean that the parametr `values` should be `&[Value]` instead of `&[&ToValue]`?

src/closure.rs
antoyo: I think I need to remove these calls, now that I use `glib_wrapper`, right?

src/closure.rs
```diff
@@ -0,0 +1,138 @@
...
+            let size = 4 + 4 + 3 * mem::size_of::<*mut c_void>() as u32;
+            let closure = gobject_ffi::g_closure_new_simple(size, ptr::null_mut());
+            assert_ne!(closure, ptr::null_mut());
+            let callback = Box::new(callback);
+            let ptr: *mut F = Box::into_raw(callback);
+            let ptr: *mut c_void = ptr as *mut _;
+            gobject_ffi::g_closure_set_meta_marshal(closure, ptr, Some(marshal::<F>));
+            gobject_ffi::g_closure_add_finalize_notifier(closure, ptr, Some(finalize::<F>));
+            gobject_ffi::g_closure_ref(closure);
+            gobject_ffi::g_closure_sink(closure);
```
sdroege: You can remove the manual calls to `ref` and `sink` here now

src/closure.rs
sdroege: Yes

src/closure.rs
sdroege: Not sure, but if it was you wouldn't have to allocate the Vec and could do everything without any copying or allocations. But with `&[&ToValue]` it's nicer for the user to call it.

sdroege:
No further comments from me 👍

antoyo:
All the tests passed, is there any other comment you want to make?

src/closure.rs
```diff
@@ -0,0 +1,136 @@
...
+            let ptr: *mut c_void = ptr as *mut _;
+            gobject_ffi::g_closure_set_meta_marshal(closure, ptr, Some(marshal::<F>));
+            gobject_ffi::g_closure_add_finalize_notifier(closure, ptr, Some(finalize::<F>));
+            from_glib_none(closure)
+        }
+    }
+
+    pub fn invoke(&self, values: &[&ToValue]) -> Option<Value> {
+        let mut result = unsafe { Value::uninitialized() };
+        let mut values: Vec<_> = values.iter()
```
GuillaumeGomez: Ii's more of a style issue, so not mandatory, but I'd prefer it to be like this:

```rust
let mut values: Vec<_> = values.iter()
                                .map(|v| v.to_value())
                                .collect();
```

src/closure.rs
```diff
@@ -0,0 +1,136 @@
...
+            .collect();
+        let gvalues = values.as_mut_ptr() as *mut _;
+        unsafe {
+            gobject_ffi::g_closure_invoke(self.to_glib_none().0 as *mut _, result.to_glib_none_mut().0,
+                values.len() as u32, gvalues, ptr::null_mut());
+        }
+        if result.type_() == Type::Invalid {
+            None
+        }
+        else {
```
GuillaumeGomez: Still a style issue (so still not mandatory): could you put the `else {` on the same line as `}` please?

GuillaumeGomez:
I made some but otherwise seems all good to me.

src/closure.rs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            let values = slice::from_raw_parts(param_values as *const _, n_param_values as usize);
+            let callback: Box<F> = Box::from_raw(marshal_data as *mut _);
+            let result = callback(values);
+            if !return_value.is_null() {
+                match result {
+                    Some(result) => *return_value = result.into_raw(),
+                    None => {
+                        let result = Value::uninitialized();
+                        *return_value = result.into_raw();
```
EPashkin: Why don't call g_value_init(return_value)?

sdroege: There's no type it could be initialized with here. We just need to fill the return value with something here, just in case. And that something is a GValue initialized with all zeroes (i.e. G_TYPE_INVALID), something we can detect later and handle appropriately.

EPashkin: You right, Type::Invalid don't accepted by g_value_init

sdroege: Value::uninitialized() is a bit misleading IMHO, it's not uninitialized but all-zeroes :)

src/closure.rs
antoyo: I believe it is best to be consistent, whatever style we choose.
For instance, I used the same style as [this](https://github.com/gtk-rs/glib/blob/master/src/value.rs#L238-L240).

src/closure.rs
```diff
@@ -0,0 +1,136 @@
...
+        ref => |ptr| {
+            gobject_ffi::g_closure_ref(ptr);
+            gobject_ffi::g_closure_sink(ptr);
+        },
+        unref => |ptr| gobject_ffi::g_closure_unref(ptr),
+    }
+}
+
+impl Closure {
+    pub fn new<F: Fn(&[Value]) -> Option<Value> + 'static>(callback: F) -> Self {
```
sdroege: One problem I forgot about is that GClosure is `Sync+Send`. You have no guarantees from which thread your callback is called, you have no guarantees that it isn't called simultaneously from multiple threads at once. As such this should probably be `F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static`

src/closure.rs
```diff
@@ -0,0 +1,136 @@
+// Copyright 2013-2017, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+// TODO: support marshaller.
```
sdroege: What does `TODO: support marshaller` mean? What do you want to support with marshallers?

antoyo: I would like to have something like `GCClosure`, i.e. being able to use a function with Rust types.
I believe it should be done in another pull request since it looks non-trivial to do something both safe and efficient.

sdroege: That would indeed be very interesting and is also on my list. For dynamically using signals/action signals
Separate pull request, yes

GuillaumeGomez:
Ok, it seems good to me.

@EPashkin @sdroege: For you?

EPashkin:
👍 Thanks, @antoyo 
And thanks @sdroege too for good ideas

GuillaumeGomez:
Let's merge then. Thanks to everyone for your work in here! :)

PR #179, opened 2017-06-06
Make BoolError constructor public

sdroege:
Allows retrieving the string by matching or constructing BoolErrors
without the from_glib() helper function


EPashkin:
IMHO better add getter

sdroege:
> IMHO better add getter

See https://github.com/gtk-rs/gtk/pull/523 for another reason. Alternatively we could also add a getter and new() function, but that seems like it has no advantage and just adds noise.

GuillaumeGomez:
Well, it doesn't harm to have it. Thanks!

PR #183, opened 2017-06-18
Generate functions that uses gsize/gusize

EPashkin:
Part of https://github.com/gtk-rs/gir/pull/385

Only ignored functions that causes compiler errors

cc @sdroege


EPashkin:
@GuillaumeGomez, take look at this. CI passed.

src/auto/functions.rs
```diff
@@ -40,13 +41,17 @@ pub fn ascii_strcasecmp(s1: &str, s2: &str) -> i32 {
     }
 }
 
-//pub fn ascii_strdown(str: &str, len: /*Unimplemented*/Fundamental: SSize) -> Option<String> {
-//    unsafe { TODO: call ffi::g_ascii_strdown() }
-//}
+pub fn ascii_strdown(str: &str, len: isize) -> Option<String> {
+    unsafe {
```
sdroege: This function is not safe as-is, but with my array-length change it will be.

Best to ignore all such functions for now

EPashkin:
Ignored functions, added comment `#gsize`.

sdroege:
👍

src/auto/key_file.rs
```diff
@@ -283,7 +293,12 @@ impl KeyFile {
         }
     }
 
-    //pub fn to_data(&self) -> Result<(String, /*Unimplemented*/Fundamental: Size), Error> {
-    //    unsafe { TODO: call ffi::g_key_file_to_data() }
-    //}
+    pub fn to_data(&self) -> Result<(String, usize), Error> {
```
sdroege: This one is also suboptimal. The usize is the length of the string, it should just return Result<String, Error>. Also going to be fixed with my array-length changes.

But this one here is at least not unsafe, only suboptimal :)

src/auto/key_file.rs
```diff
@@ -116,9 +133,14 @@ impl KeyFile {
         }
     }
 
-    //pub fn get_string_list(&self, group_name: &str, key: &str) -> Result<(Vec<String>, /*Unimplemented*/Fundamental: Size), Error> {
-    //    unsafe { TODO: call ffi::g_key_file_get_string_list() }
-    //}
+    pub fn get_string_list(&self, group_name: &str, key: &str) -> Result<(Vec<String>, usize), Error> {
```
sdroege: Same as g_key_file_to_data()

src/auto/key_file.rs
EPashkin: Yes, that why I don't ignore it and other returning functions.

sdroege: Yeah, just saying :) In my local branch this ends up being `pub fn to_data(&self) -> Result<String, Error>` already btw. Hope to be able to clean that up a bit for an initial review later today.

GuillaumeGomez:
I'm fine with this and @sdroege seems to be fine too so let's merge. Thanks!

src/auto/key_file.rs
sdroege: This one actually not because it misses the array-length annotation for the return value in GLib. We should go through all these things and report them in Bugzilla...

EPashkin:
Opps, forgot to check git submodule version: it stay old, in other PR's seems too.

GuillaumeGomez:
Oups, forgot to check as well. Good catch @EPashkin!

PR #189, opened 2017-07-05
Implement public ObjectExt::set_property() and ::get_property()

sdroege:
This is useful for using properties (or properties of types) that are
not part of any bindings in a still safe way.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -9,6 +9,8 @@ use types::{self, StaticType};
 use wrapper::{UnsafeFrom, Wrapper};
 use gobject_ffi;
 
+use Value;
+
 /// Upcasting and downcasting support.
 ///
 /// Provides conversions up and down the class hierarchy tree.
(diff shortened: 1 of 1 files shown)


sdroege:
For some background, I need this for the GStreamer bindings but could also add it to gst::Object instead. In GStreamer, plugins are providing new GObject subclasses that don't exist in any (public) headers and thus don't have any GObject-Introspection data around them (and can't easily be part of any bindings as such).

The API to be used on such types is the standard GObject API: setting/getting properties, connecting to signals, etc.


If this is accepted here, I'll also submit further pull requests that allow connecting (safely!) to arbitrary signals (via @antoyo's GClosure work) or emitting them. These will also be based on GValues.

GuillaumeGomez:
Wo, great, thanks!

EPashkin:
Thanks, @sdroege.
I agree that these functions is actually safe, but as warning maybe better mark unsafe?

sdroege:
The functions are not unsafe though, they are as safe as putting things in a `HashMap<String, Value>`, and for example safer than current generated property setters for `i32` properties that have a limited range (setting a value outside the range will give a warning but it's still set).

In addition to this I'd also like to add `ObjectExt::has_property(property_name: &str, type_: Option<Type>) -> bool` later, if we agree that this is acceptable as-is.

EPashkin:
Ok, then 👍 from me.
Thanks again.

sdroege:
Thanks, preparing the other pieces later :)

Issue #187, opened 2017-07-03
Add support for get_type() functions for boxed types, enums, flags


sdroege:
Currently glib_wrapper! only allows them for GObjects. Boxed types also have get_type() functions though, and they are important for using them in GValues.

Same for flags and enums.

EPashkin:
That would be good.
There problem with boxed that all value traits (ex. SetValueOptional) generalized on `IsA<Object>` and difficult to make same for boxed.
Seems better add macros for defining these traits using [RGBA](https://github.com/gtk-rs/gdk/blob/master/src/rgba.rs#L182-L200) as example.
Defining `glib::StaticType` also can be macro, but IMHO it unneeded.
Also until we have special glib_wrapper! for "direct" types (RGBA, Rectangle) macro for type better have separate version.

sdroege:
Together with my plan from https://github.com/gtk-rs/gir/issues/388 , all these related traits would be directly implemented by the macro instead of implementing them generically. That way you get rid of all the overlapping impls problems. The generic impls we have right now are too generic.

sdroege:
Could make the get_type() function another "special" function also, for the implementation that probably makes most sense

EPashkin:
Only problem that it mostly not defined as function, only as `glib:get-type` attribute to main object none

sdroege:
https://github.com/gtk-rs/gir/pull/410
https://github.com/gtk-rs/gir/pull/409
https://github.com/gtk-rs/glib/pull/199

sdroege:
That's all there now

GuillaumeGomez:
🎊

PR #190, opened 2017-07-05
Various GObject property changes

sdroege:


EPashkin:
👍, thought it currently not works with enums.

sdroege:
> thought it currently not works with enums.

Correct unfortunately: https://github.com/gtk-rs/glib/issues/187

My plan here is to get rid of the generic trait impls for all the conversion traits and get them via macros, and then we get GValue support for boxed types, enums, etc automatically with some additional effort.

sdroege:
@GuillaumeGomez Let's get this in then? :)

GuillaumeGomez:
Indeed! (Started to look at it but forgot meantime...)

Thanks!

PR #191, opened 2017-07-06
Implement ObjectExt::connect() for connecting to arbitrary signals

sdroege:


src/object.rs
```diff
@@ -414,4 +419,60 @@ impl<T: IsA<Object>> ObjectExt for T {
...
+            // This is actually G_SIGNAL_TYPE_STATIC_SCOPE
+            let return_type = details.return_type & (!gobject_ffi::G_TYPE_FLAG_RESERVED_ID_BIT);
+            let closure = Closure::new(move |values| {
+                let ret = callback(values);
+
+                if return_type == gobject_ffi::G_TYPE_NONE {
+                    // Silently drop return value
+                    None
+                } else {
+                    // Silently create empty return value
```
sdroege: Main question here is whether we should really silently drop/introduce return values, or if this should result in a panic!() or exit() instead, or just printing some warning

src/object.rs
```diff
@@ -414,4 +419,60 @@ impl<T: IsA<Object>> ObjectExt for T {
...
     }
+
+    fn connect<'a, N, F>(&self, signal_name: N, after: bool, callback: F) -> Result<u64, BoolError>
+        where N: Into<&'a str>, F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static {
+        let signal_name: &str = signal_name.into();
+
+        unsafe {
+            let obj = self.to_glib_none().0;
+            let klass = (*obj).g_type_instance.g_class as *mut gobject_ffi::GTypeClass;
+            let type_ = (*klass).g_type;
```
EPashkin: Why not use `let type_ = T::static_type().to_glib();` instead?

sdroege: Because this gives you the static type of whatever was passed in here. So if you have a GtkButton and cast it to a GtkWidget, and pass it in here then, you would get the type of GtkWidget (which does not have e.g. the "clicked" signal).

Or more problematic, in case of GStreamer the types of objects from plugins don't even have bindings and you always pass them around as GstElements.

sdroege: That said, it might make sense to extract the above code into a new ObjectExt::get_type() function. Or what do you think?

src/object.rs
```diff
@@ -414,4 +419,60 @@ impl<T: IsA<Object>> ObjectExt for T {
...
+                    // Silently drop return value
+                    None
+                } else {
+                    // Silently create empty return value
+                    if let None = ret {
+                        let mut value = Value::uninitialized();
+                        gobject_ffi::g_value_init(value.to_glib_none_mut().0, return_type);
+                        Some(value)
+                    } else {
+                        ret
```
sdroege: Also in this case we want to check if ret's type_ == return_type too, and if not do the same as if None was returned

src/object.rs
EPashkin: You right, signal need be added to actual type.

src/object.rs
EPashkin: Reasonable

src/object.rs
EPashkin: `get_type` also good idea, but name better be others to not conflict with cairo's Surface and Font.

EPashkin: Hm, missed that both of it don't Object, so ignore my word about rename.

src/object.rs
sdroege: @EPashkin What do you think?

sdroege: All but this one solved now

EPashkin: I agree with you current decision.

EPashkin:
Maybe comment on call `g_object_set_property` can be removed already?

sdroege:
> Maybe comment on call g_object_set_property can be removed already?

Which?

EPashkin:
There only one call of g_object_set_property :wink:
As you added type check beforehand this comment IMHO already unneeded.
```
// Using property names that don't exist or using a GValue of a wrong type
// causes a warning printed at runtime by GObject and is considered a
// programming error. It is however memory-safe to do so
```

sdroege:
Ah, that's unrelated to this pull request though :) Can do, sure

EPashkin:
👍 
Big thanks, @sdroege ,now we have generic signals and will have way to emit it ;)

sdroege:
Now we only need to be able to put enums/flags and all that into GValues, but that comes next then (I actually need that working for GStreamer too).

GuillaumeGomez:
I just have a small question/concern: shouldn't the handler be put inside a type so users don't confuse it with something else?

sdroege:
> I just have a small question/concern: shouldn't the handler be put inside a type so users don't confuse it with something else?

It's the same as with all the other signal related functions, like the ones generated from gir :) I agree with you though, but let's do that in another step.

sdroege:
And not just `type` but a newtype wrapper imho (`struct SignalId(u64)`)

GuillaumeGomez:
That's what I meant when I wrote "put inside a type". But you're right: I should have said newtype, my bad.

Except for this point, everything seems good to me. @EPashkin Good for you too?

EPashkin:
Yes.

GuillaumeGomez:
Then I merge. Thanks!

PR #192, opened 2017-07-06
Implement ObjectExt::emit() for emitting arbitrary signals

sdroege:
Depends on, and includes, https://github.com/gtk-rs/glib/pull/191

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -9,6 +9,7 @@ use types::{self, StaticType};
 use wrapper::{UnsafeFrom, Wrapper};
 use gobject_ffi;
 use std::mem;
+use std::ptr;
 
 use Value;
 use Type;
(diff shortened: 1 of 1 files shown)


sdroege:
Before you ask, this is mostly useful for G_SIGNAL_ACTION signals. E.g. GtkWindow's activate-focus signal.

Action signals are basically function calls. They are emitted/called from the outside on the object, and the object handles them internally.

sdroege:
And updated so that it only includes the relevant commit here now

EPashkin:
👍 Thanks again

GuillaumeGomez:
Thanks! Waiting for CIs.

GuillaumeGomez:
Indeed. :)

Issue #168, opened 2017-05-12
GMainLoop, GMainContext, GSource bindings


sdroege:
Currently only the static variants around the global default main context are exposed. This is useful to have more control

EPashkin:
All 3 can be generated.
First problem in generated `g_main_context_prepare`: `priority` don't marked as out parameter so gir use it wrong. But it can be implemented manually.
```
pub fn prepare(&self, priority: i32) -> bool {
     unsafe {
          from_glib(ffi::g_main_context_prepare(self.to_glib_none().0, priority))
      }
}
```
Also these objects many functions that need PoolFD and other currently unsupported types,
so I don't sure that generated objects will be usable and don't miss something necessary.

sdroege:
For PollFD you can write manual code, which then exposes the underlying file descriptor. That could then be used with other Rust functions using file descriptors and potentially even things like mio.
Apart from actually finding time to write the code, there don't seem to be any problems :)

antoyo:
These were implemented manually during the HackFest. You can see them in [futures-glib](https://github.com/antoyo/futures-glib-rs).

sdroege:
Having non-futures versions would be useful too though (with callbacks), and then implement futures versions on top of that. I have an implementation for those somewhere from a couple of years back, have to find it :)

antoyo:
Yeah, that would be nice.

sdroege:
I think for the GSource API we don't necessarily want to autogenerate that but rather write it manually with the builder pattern, or a single constructor taking all relevant parts. The setter functions on GSource are not thread-safe and not supposed to be used after attachment to a main context, unlike all the other API on it.

Also need to carefully place Send and Sync bounds/impls everywhere.

sdroege:
@antoyo did you already try switching over futures-glib?

antoyo:
You mean using these instead of the ones in `futures-glib`?
No, I haven't done it yet.
I'm not sure I'll have everything I need, though.

sdroege:
Shout if something is missing for you

PR #194, opened 2017-07-11
Initial GMainLoop/GMainContext bindings

sdroege:
https://github.com/gtk-rs/glib/issues/168


src/source.rs
```diff
@@ -146,3 +195,137 @@ pub fn source_remove(source_id: Id) {
...
+pub const PRIORITY_DEFAULT: Priority = Priority(glib_ffi::G_PRIORITY_DEFAULT);
+pub const PRIORITY_HIGH_IDLE: Priority = Priority(glib_ffi::G_PRIORITY_HIGH_IDLE);
+pub const PRIORITY_DEFAULT_IDLE: Priority = Priority(glib_ffi::G_PRIORITY_DEFAULT_IDLE);
+pub const PRIORITY_LOW: Priority = Priority(glib_ffi::G_PRIORITY_LOW);
+
+/// Adds a closure to be called by the main loop the return `Source` is attached to when it's idle.
+///
+/// `func` will be called repeatedly until it returns `Continue(false)`.
+pub fn idle_source_new<'a, N: Into<Option<&'a str>>, F>(name: N, priority: Priority, func: F) -> Source
+where F: FnMut() -> Continue + Send + 'static {
```
sdroege: The question here is mostly whether we want single functions for creating Sources, or create something with a builder pattern here. Single functions seems more simple to me

sdroege:
For MainContext::invoke() this could in theory be this, but I'm not sure how to handle `FnOnce` inside a `Box<_>`. What am I missing?
This fails with `error[E0161]: cannot move a value of type std::ops::FnOnce() + std::marker::Send: the size of std::ops::FnOnce() + std::marker::Send cannot be statically determined`

```Rust
impl MainContext {
    pub fn invoke<F>(&self, func: F)
    where F: FnOnce() + Send + 'static {
        unsafe {
            glib_ffi::g_main_context_invoke_full(self.to_glib_none().0, glib_ffi::G_PRIORITY_DEFAULT_IDLE, Some(trampoline),
                into_raw(func), Some(destroy_closure))
        }
    }

    pub fn invoke_with_priority<F>(&self, priority: Priority, func: F)
    where F: FnOnce() + Send + 'static {
        unsafe {
            glib_ffi::g_main_context_invoke_full(self.to_glib_none().0, priority.to_glib(), Some(trampoline),
                into_raw(func), Some(destroy_closure))
        }
    }
}

unsafe extern "C" fn trampoline(func: gpointer) -> gboolean {
    let _guard = CallbackGuard::new();
    let func: &RefCell<Option<Box<FnOnce() + Send + 'static>>> = transmute(func);

    if let Some(func) = func.borrow_mut().take() {
        func();
    }

    glib_ffi::G_SOURCE_REMOVE
}

unsafe extern "C" fn destroy_closure(ptr: gpointer) {
    let _guard = CallbackGuard::new();
    Box::<RefCell<Option<Box<FnOnce() + Send + 'static>>>>::from_raw(ptr as *mut _);
}

fn into_raw<F: FnOnce() + Send + 'static>(func: F) -> gpointer {
    let func: Box<RefCell<Option<Box<FnOnce() + Send + 'static>>>> =
        Box::new(RefCell::new(Some(Box::new(func))));
    Box::into_raw(func) as gpointer
}
```

sdroege:
Problem is that we need something like `FnBox` https://doc.rust-lang.org/alloc/boxed/trait.FnBox.html

sdroege:
Good for review then IMHO, @GuillaumeGomez @EPashkin :) No further additions planned to this pull request

src/main_context.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    // FIXME: These can actually be FnOnce but require FnBox to
+    // stabilize, or Box<FnOnce()> to be callable otherwise
+    pub fn invoke<F>(&self, func: F)
+    where F: FnMut() + Send + 'static {
+        unsafe {
+            glib_ffi::g_main_context_invoke_full(self.to_glib_none().0, glib_ffi::G_PRIORITY_DEFAULT_IDLE, Some(trampoline),
+                into_raw(func), Some(destroy_closure))
+        }
+    }
```
EPashkin: We always pass to FFI boxed closure. And `destroy_closure` working with boxed.

EPashkin: Hm. Missed that you added your version `destory_closure`

src/lib.rs
```diff
@@ -183,7 +190,7 @@ mod file_error;
 mod key_file;
 pub mod prelude;
 pub mod signal;
-mod source;
+pub mod source;
```
EPashkin: Part of source reexported, while part is not.
Not better do `pub use source::*;` ?

EPashkin:
Code looks good to me, but PR missing update Gir submodule and regen to correct version.

src/auto/functions.rs
```diff
@@ -572,10 +569,6 @@ pub fn hostname_to_unicode(hostname: &str) -> Option<String> {
 //    unsafe { TODO: call ffi::g_idle_remove_by_data() }
 //}
 
-//pub fn idle_source_new() -> /*Ignored*/Option<Source> {
```
GuillaumeGomez: Since we bound `Source`, why is this ignored?

src/auto/functions.rs
```diff
@@ -142,10 +143,6 @@ pub fn check_version(required_major: u32, required_minor: u32, required_micro: u
 //    unsafe { TODO: call ffi::g_child_watch_add_full() }
 //}
 
-//pub fn child_watch_source_new(pid: Pid) -> /*Ignored*/Option<Source> {
```
GuillaumeGomez: Since we bound `Source`, why is this ignored? (We don't have `Pid`?)

src/auto/functions.rs
GuillaumeGomez: Ah nevermind...

src/lib.rs
sdroege: Your call :)

src/source.rs
GuillaumeGomez: Either way seems fine to me so do whatever seems better to you.

src/lib.rs
GuillaumeGomez: No preference for me so do whatever you want.

GuillaumeGomez:
Looks good to me as well. Fix @EPashkin's issues and let's merge!

sdroege:
@EPashkin What do you want to have fixed? Should source::* be re-exported at the crate level?

EPashkin:
@sdroege Yes + gir submodule update +  regen

sdroege:
OK, will do in an hour or two

sdroege:
Done :)

sdroege:
Ah I guess that's the reason for not re-exporting source::*. source::Id should be re-exported as SourceId. What do you suggest? Rename Id to SourceId?

sdroege:
Let's try that...

EPashkin:
👍
Thanks @sdroege

GuillaumeGomez:
The "changes PR" is getting really huge. Next release will be certainly a worldwide ovation. People will throw money at us. Gods will sing our names and we'll be taken as examples by future generations. Can't wait to be at the release. :sob:

sdroege:
I'd like to get some more things into the next release though, most importantly removing generic impls for the various conversions traits (and move them into the macros) and implement them for enums/flags/boxed/etc types (which is not possible right now. so we can have those in GValues, for example)

GuillaumeGomez:
I don't intend to make the release before at least another month. I still need to work on gtk-rs/release anyway.

sdroege:
🎉

GuillaumeGomez:
Thanks a lot!

PR #196, opened 2017-07-12
Implement Value traits inside glib_wrapper!() instead of doing it gen…

sdroege:
…erically

This is needed to be able to create implementations for other pointer
types, e.g. boxed types, without creating overlapping implementations.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -12,6 +12,7 @@ use std::mem;
 use std::ptr;
 
 use Value;
+use value::SetValue;
 use Type;
 use BoolError;
 use Closure;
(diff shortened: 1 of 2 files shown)


src/auto/enums.rs
```diff
@@ -1,4 +1,4 @@
-// This file was generated by gir (7504fb6) from gir-files (71d73f0)
+// This file was generated by gir (93fb71d) from gir-files (71d73f0)
```
EPashkin: This change really needed?

EPashkin:
Seems Gir need changed before this to add `use gobject_ffi`

EPashkin:
Oh, missed that you already added it.

src/object.rs
```diff
@@ -12,6 +12,7 @@ use std::mem;
 use std::ptr;
 
 use Value;
+use value::{FromValueOptional, SetValueOptional, SetValue};
```
EPashkin: Seems only SetValue needed

sdroege: The macro also implements SetValueOptional

sdroege: Oh. Yes, only SetValue is needed indeed :)

EPashkin:
👍 after after fixes

sdroege:
Done. And you need to merge all the others ones at the same time to keep things working

EPashkin:
Checked all linked PRs and all works (only need update gir submodule).
My only worry was, that now we need add `gobject_sys` to dependence to each lib with object.
And I thought that better reexport `g_value_dup_object`, `g_value_set_object` and `g_value_set_object` in glib and use reexported in macros?
But this method will result reexport 3 boxed functions `g_value_set_boxed` too and maybe others,
so currently I prefer this PR as is.
@sdroege, @GuillaumeGomez, what you think?

sdroege:
All PRs updated to also have the gir submodule updated.

About requiring `gobject_sys`, this was already required before for example whenever you had an object with properties. I don't think that's much of a problem really. (And yes, others too in a bit: `g_value_set_enum`, `g_value_set_flags`, etc)

EPashkin:
Thanks, @sdroege 
@GuillaumeGomez, I vote for merge this PR and all related regens.

GuillaumeGomez:
Thanks for all these PRs (took me a bit of time to read all this code ;) )!

PR #198, opened 2017-07-14
Remove generic pointer array impls of ToGlibPtr and FromGlibContainer

sdroege:
This introduces new traits:
- ToGlibContainerFromSlice: Conversion from a &[Self] into some GLib
  container (array, GList, etc).

- FromGlibContainerAsVec: Conversion from a GLib container (array,
  GList, etc) to a Vec<Self>. This always requires the length to be
  given and can be used for non-pointer containers (e.g. uint8_t[]).

- FromGlibPtrContainerAsVec: Some as FromGlibContainerAsVec but does not
  require the length, and instead requires pointers as elements and the
  container to be NULL-terminated

They are all implemented for T and not [T]/Vec<T>, which is required for this to
be possible to implement in crates other than the glib crate.

impls for Strings and fundamental types exist in GLib, glib_wrapper!()
is automatically implementing them too.

Based on these there are generic implementations for ToGlibPtr for [T]
and FromGlibContainer for Vec<T>.

Also split FromGlibPtrContainer into FromGlibContainer and FromGlibPtrContainer,
the first requiring a length to be given (and thus is usable for
non-NULL terminated or non-pointer containers), the second requires
pointers and NULL-termination.

With all this in place it is now possible to implement conversions
from/to containers of fundamental types, enums, flags, etc.


GuillaumeGomez:
I already said it earlier but for the records: didn't see anything shocking. Thanks for the huge update!

sdroege:
It might make sense to generate the impls in the macros by yet another macro, they're basically copy&paste right now. But that would mean that GLib exports yet another macro, so I'm not sure about that. Your decision

GuillaumeGomez:
It depends on the final outcome: if we still just have to call one macro (which will call whatever macros), it's fine. I'd just prefer that we don't have to directly call multiple macros. (If that makes sense?)

sdroege:
Hm, the new traits should probably have different function names. Otherwise they conflict with the other ones and have to be explicitly specified. Doing so later, but for an initial review this should be ok.

EPashkin:
Looks good for fast check, will check more tomorrow.
Thanks, @sdroege

sdroege:
> It depends on the final outcome: if we still just have to call one macro (which will call whatever macros), it's fine. I'd just prefer that we don't have to directly call multiple macros. (If that makes sense?)

Yes it would still just be glib_wrapper!() that is called... but internally it would call some new macro

sdroege:
Actually can't use macros for this because of different mutability, and that's not easy to abstract over

EPashkin:
Cairo now need adding feature like "use_glib" instead "glib" to include both "glib" and "glib_sys" crates

sdroege:
All complete and compiling from my side now

sdroege:
I'd prefer for this (and the related PRs) to be merged before my other PRs btw. I'll rebase the others as needed then (they just add features, don't break things).

EPashkin:
Thanks, @sdroege all PRs look good, only last update to last gir when needed.

sdroege:
Updated gir

EPashkin:
@GuillaumeGomez, 👍 for merge this and all linked PRs

GuillaumeGomez:
Taking a look in a few hours.

GuillaumeGomez:
Thanks a lot!

sdroege:
Thanks :) now only gio and gdk-pixbuf are missing

PR #200, opened 2017-07-19
Fix transfer full conversion of pointer arrays

sdroege:
Instead of the original pointer, the pointer at the end of the array was
being freed.

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -161,15 +161,14 @@ macro_rules! glib_boxed_wrapper {
 
         #[doc(hidden)]
         impl $crate::translate::FromGlibContainerAsVec<*mut $ffi_name, *mut *mut $ffi_name> for $name {
-            unsafe fn from_glib_none_num_as_vec(mut ptr: *mut *mut $ffi_name, num: usize) -> Vec<Self> {
+            unsafe fn from_glib_none_num_as_vec(ptr: *mut *mut $ffi_name, num: usize) -> Vec<Self> {
                 if num == 0 || ptr.is_null() {
                     return Vec::new();
                 }
 
                 let mut res = Vec::with_capacity(num);
-                for _ in 0..num {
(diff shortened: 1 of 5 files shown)


src/boxed.rs
```diff
@@ -180,15 +179,14 @@ macro_rules! glib_boxed_wrapper {
...
                 }
 
                 let mut res = Vec::with_capacity(num);
-                for _ in 0..num {
-                    res.push($crate::translate::from_glib_full(ptr::read(ptr)));
-                    ptr = ptr.offset(1);
+                for i in 0..num {
+                    res.push($crate::translate::from_glib_full(ptr::read(ptr.offset(i as isize))));
                 }
                 glib_ffi::g_free(ptr as *mut _);
```
EPashkin: IMHO old case was better. For full version enough just copy ptr to variable.

sdroege: The point here is to also remove the mut variable to prevent such sillyness in the future. It's equivalent code-wise apart from that.

EPashkin: But a little slower.
Ok as you both think that it is good.

sdroege: We're copying everything here anyway (which we don't really need if we had more infrastructure) :) And it shouldn't be slower really: first of all, a sufficiently smart compiler should otherwise be able to optimize it that way, and more importantly: the difference is between +1, and +1 for the iterator and +n for the pointer. That's one extra instruction (register +1) and one changed instruction (pointer addition by register instead of immediate value). That hardly matters.

I didn't check the assembly output though, it doesn't seem worth thinking about this here.

GuillaumeGomez:
@EPashkin: Are you fine with this change or do you want to think about it more? If you have good arguments against it, I'm open to discussion. :)

EPashkin:
Yes, I fine with this change.

PR #202, opened 2017-07-20
Implement public ObjectExt::get_property_type()

sdroege:
It was private before.


src/object.rs
```diff
@@ -597,6 +585,22 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
         }
     }
 
+
```
EPashkin: Extra line

sdroege:
Use case for this is that I need to get the type of flags/enums that are only defined in a plugin. And then use the GEnum/GFlags API to do something meaningful with them (patches for that following in a bit).

EPashkin:
👍 
Thanks, @sdroege

GuillaumeGomez:
Thanks!

PR #204, opened 2017-07-23
Add WeakRef support to glib::Object

sdroege:
Another thing for which I had custom bindings in my GStreamer code.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -11,6 +11,7 @@ use ffi as glib_ffi;
 use gobject_ffi;
 use std::mem;
 use std::ptr;
+use std::marker::PhantomData;
 
 use Value;
 use value::SetValue;
@@ -515,7 +516,7 @@ glib_object_wrapper! {
     Object, GObject, @get_type gobject_ffi::g_object_get_type()
 }
 
-pub trait ObjectExt {
+pub trait ObjectExt: IsA<Object> {
     fn get_type(&self) -> Type;
 
     fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &Value) -> Result<(), BoolError>;
@@ -526,6 +527,8 @@ pub trait ObjectExt {
     fn connect<'a, N, F>(&self, signal_name: N, after: bool, callback: F) -> Result<u64, BoolError>
         where N: Into<&'a str>, F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static;
     fn emit<'a, N: Into<&'a str>>(&self, signal_name: N, args: &[&Value]) -> Result<Option<Value>, BoolError>;
+
+    fn downgrade(&self) -> WeakRef<Self>;
 }
 
 impl<T: IsA<Object> + SetValue> ObjectExt for T {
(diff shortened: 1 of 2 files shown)


src/object.rs
```diff
@@ -714,4 +717,56 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+        unsafe {
+            let w = WeakRef(Box::new(mem::zeroed()), PhantomData);
+            gobject_ffi::g_weak_ref_init(mut_override(&*w.0), self.to_glib_none().0);
+            w
+        }
+
+    }
+}
+
+pub struct WeakRef<T: IsA<Object> + ?Sized>(Box<gobject_ffi::GWeakRef>, PhantomData<*const T>);
```
sdroege: Has to be a Box because GWeakRef remembers the address of this, and otherwise it would remember an "unstable" address (it is changing when moving the WeakRef around) and use it later (when the real object is destroyed).

sdroege:
Naming of the functions comes from `Arc`

sdroege:
@GuillaumeGomez Do you have any suggestions for the function names, or does it sound good to you like this?

src/object.rs
```diff
@@ -714,4 +717,56 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+    }
+}
+
+impl<T: IsA<Object> + ?Sized> Clone for WeakRef<T> {
+    fn clone(&self) -> Self {
+        unsafe {
+            let c = WeakRef(Box::new(mem::uninitialized()), PhantomData);
+
+            let o = gobject_ffi::g_weak_ref_get(mut_override(&*self.0));
+            gobject_ffi::g_weak_ref_init(mut_override(&*self.0), o);
```
EPashkin: Seems `c` won't used in this function and always empty

src/object.rs
```diff
@@ -714,4 +717,56 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+                gobject_ffi::g_object_unref(o);
+            }
+
+            c
+        }
+    }
+}
+
+unsafe impl<T: IsA<Object> + Send + Sync + ?Sized> Send for WeakRef<T> {}
+unsafe impl<T: IsA<Object> + Send + Sync + ?Sized> Sync for WeakRef<T> {}
```
EPashkin: Maybe +Send => Send, +Sync => Sync

EPashkin: Also `T: IsA<Object>` can be unsized?

GuillaumeGomez:
I'll take a look in ~1 hour.

src/object.rs
sdroege: Yes, bug. Thanks! The self should be c.

src/object.rs
sdroege: You need both to have each when reference counting is involved. We ensure that for Object anyway in gir, but better play safe here too.

For unsized, as a trait object maybe? Otherwise not really I guess. But doesn't hurt. I can remove that if you prefer that.

src/object.rs
sdroege: I had no Clone in my code before and only added that for consistency with Arc now. So that part was untested. But easy to fix, will do so later

GuillaumeGomez:
What's wrong with `upgrade` and `downgrade`?

src/object.rs
EPashkin: Ok. It was just note about unsized.

sdroege:
I thought they might be confusing with upcast and downcast. If you're fine with the names, great :) will fix it up later then. Thanks for looking

sdroege:
Should be good to go now

GuillaumeGomez:
I'll look at it tomorrow. Please ping me if I forget. ;)

sdroege:
@GuillaumeGomez ping :)

GuillaumeGomez:
And as promised it got out of my mind. Since I have nothing to say on this, I merge. Thanks!

sdroege:
Thanks!

Issue #205, opened 2017-07-26
Cast/Downcast traits are unnecessarily restrictive


sdroege:
Two problems here:

- Currently `upcast()` works exclusively on the `IsA<_>` trait. That is, you can only upcast from `A` to `B`, if `A: IsA<B>`. Upcasting is used (1) for going from a subclass to a parent class, or (2) for going from some class to an interface implemented by it.
Case (1) should always work fine, as it will be statically known that `A: IsA<B>` if both types are represented in Rust.
Case (2) however will fail easily: Consider a function that returns an object of type `A` according to its signature, but actually returns an object of a subclass `X` of `A` and `X` is not defined anywhere in Rust (might not even be in C headers or anywhere). This is the case for various things in GIO, and GStreamer elements.

The same also happens if you have an object of type `A`, stored in a variable of type `B` (`B` is the parent class of `A`). `A` implements interface `C` (and this is statically known) but `B` does not. You can't cast from `B` to `C` currently.

- Similarly, `downcast()` currently only works if there is a `IsA` relationship statically known between the types. Which fails in the same cases as above.

I would suggest that we deprecate/remove these traits and instead just have
- a `cast()` function that can cast between anything and checks at runtime (similar to downcast right now). This could be in `ObjectExt` also. The distinction between upcast and downcast is not really relevant here as you always have to check at runtime anyway for interfaces at least, and it can always fail.
- optionally a `static_upcast()` function (equivalent to the old `upcast()`) which only looks at the `IsA<_>` traits, and can because of that only support statically known upcasting (and never downcasting). This could be a free function, a new trait that is implemented for everything that implements `IsA<_>`, ...
- `static_downcast` (equivalent to the old `downcast`) is not really relevant or useful. It would have to check at runtime anyway, like `cast()` would, and it would only give very minor (and easily wrong, too restrictive) safety guards by the compiler.

Opinions?

GuillaumeGomez:
I have to admit that having most of checks performed at compile-time is a great thing from my point of view. Even if in the facts we can completely break it, it's still a nice base to build upon. So except if the new system brings really huge improvements (I'm not a big user of `GIO` or `GStreamer` so I don't have your use case in mind even though I have a little idea), I think we'll need to discuss this a lot more. This isn't an easy move and I want to be sure to understand all up and down sides before deciding anything!

EPashkin:
IMHO better don't change current functions of `trait Cast`
I don't against adding `cast` function without `IsA` bounds,
[...]

PR #206, opened 2017-07-27
Add Cast::dynamic_cast()

sdroege:
This does all type-checking at runtime and is required because it's not
always statically known at compile-time whether an object implements an
interface or not.

Also add notes about this limitation to Cast::downcast() and Cast::upcast().

And change Cast::is() to only perform runtime checking. It was always
doing runtime checking anyway, but in addition did compile-time checking
which is more restricted. If compile-time checking is wanted in
addition, Downcast::can_downcast() can be used for the old behaviour.

https://github.com/gtk-rs/glib/issues/205

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -25,6 +25,11 @@ use Closure;
 pub trait Cast: IsA<Object> {
     /// Upcasts an object to a superclass or interface `T`.
     ///
+    /// *NOTE*: This statically checks at compile-time if casting is possible. It is not always
+    /// known at compile-time, whether a specific object implements an interface or not, in which case
+    /// `upcast` would fail to compile. `dynamic_cast` can be used in these circumstances, which
+    /// is checking the types at runtime.
+    ///
     /// # Example
     ///
     /// ```ignore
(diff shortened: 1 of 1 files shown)


sdroege:
This is basically the same as downcast but without compile-time constraints. Confirmed to work well here for casting a GstElement to the GstURIHandler interface (which some plugin-defined GstElement subclasses implement, others don't).

EPashkin:
👍

src/object.rs
```diff
@@ -56,9 +66,38 @@ pub trait Cast: IsA<Object> {
         Downcast::downcast(self)
     }
 
-    /// Returns `true` if the object is an instance of (can be downcast to) `T`.
-    fn is<T>(&self) -> bool where Self: Downcast<T> {
-        Downcast::can_downcast(self)
+    /// Returns `true` if the object is an instance of (can be cast to) `T`.
+    fn is<T>(&self) -> bool where T: StaticType {
```
GuillaumeGomez: Can you put the where clause on the next line please?

GuillaumeGomez:
Just one nit and it's good for me.

sdroege:
Done, thank

GuillaumeGomez:
Thanks!

PR #207, opened 2017-07-30
Some helper API to dynamically work with GLib enums/flags

sdroege:
See commits

### DIFF
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -43,10 +43,15 @@ impl ToGlib for UserDirectory {
     }
 }
 
+/// Representation of an `enum` for dynamically, at runtime, querying the values of the enum and
+/// using them.
 #[derive(Debug)]
 pub struct EnumClass(*mut gobject_ffi::GEnumClass);
 
 impl EnumClass {
+    /// Create a new `EnumClass` from a `Type`.
+    ///
+    /// Returns `None` if `type_` is not representing an enum.
     pub fn new(type_: Type) -> Option<Self> {
         unsafe {
             let is_enum: bool = from_glib(gobject_ffi::g_type_is_a(type_.to_glib(), gobject_ffi::G_TYPE_ENUM));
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
We have a big lack of documentation in here!

sdroege:
That's very true :) I can write documentation when you tell me that the API is good as-is

src/enums.rs
```diff
@@ -37,3 +41,505 @@ impl ToGlib for UserDirectory {
...
     }
 }
+
+#[derive(Debug)]
+pub struct EnumClass(*mut gobject_ffi::GEnumClass);
+
+impl EnumClass {
+    pub fn new(type_: Type) -> Option<Self> {
+        unsafe {
+            let is_enum: bool = from_glib(gobject_ffi::g_type_is_a(type_.to_glib(), gobject_ffi::G_TYPE_ENUM));
```
GuillaumeGomez: Why not:

```rust
if from_glib::<bool>(gobject_ffi::g_type_is_a(type_.to_glib(), gobject_ffi::G_TYPE_ENUM)) {
    Some(EnumClass(gobject_ffi::g_type_class_ref(type_.to_glib()) as *mut _))
} else {
    None
}
```

?

src/enums.rs
```diff
@@ -37,3 +41,505 @@ impl ToGlib for UserDirectory {
...
+}
+
+
+#[derive(Debug)]
+pub struct FlagsClass(*mut gobject_ffi::GFlagsClass);
+
+impl FlagsClass {
+    pub fn new(type_: Type) -> Option<Self> {
+        unsafe {
+            let is_flags: bool = from_glib(gobject_ffi::g_type_is_a(type_.to_glib(), gobject_ffi::G_TYPE_FLAGS));
```
GuillaumeGomez: Same as above.

GuillaumeGomez:
So, it seems good for me as is. My only concern is that I miss something because I don't have idea where to use this code for the moment.

EPashkin:
@sdroege IMHO API is good.

sdroege:
> My only concern is that I miss something because I don't have idea where to use this code for the moment.

https://gist.github.com/sdroege/46868640122e373c814b593652878f79#file-playbin-rs-L24 would be an example of flags usage

sdroege:
Thanks for looking and the very fast feedback, I'll get docs written tomorrow then :)

GuillaumeGomez:
Thanks! :)

sdroege:
Docs added

sdroege:
Anything you want to have added/clarified in the docs or does it look ok?

GuillaumeGomez:
No, that's all good. Since @EPashkin already gave his approval, we can merge. Thanks again!

PR #208, opened 2017-07-30
Add GDate & GDateTime (and related) bindings

sdroege:


sdroege:
src/date.rs was autogenerated with gir by cheating: I edited the .gir file to add a copy function, and then implemented that one correctly afterwards. I don't think we can teach gir to handle this correctly, as without any copy/free functions we don't know if this is with reference counting or copying.

GuillaumeGomez:
I'm a bit worry about this: if you edited the .gir file, it means only you can generate this for the moment, right?

sdroege:
That's why it is in src/date.rs and not src/auto/date.rs. The problem here is that we'll never be able to autogenerate this properly unless GLib is adding a copy function.

GuillaumeGomez:
Ah I see, my bad. (it's awesome!)

src/date.rs
```diff
@@ -0,0 +1,330 @@
+// This file was generated by gir (a01311c+) from gir-files (0bcaef9+)
+// DO NOT EDIT
```
EPashkin: This need other header

EPashkin:
Code looks good, but I prefer that it wait for https://github.com/gtk-rs/glib/pull/209 and then rebased and regened with rights hashes

src/date.rs
sdroege: Done

sdroege:
This probably applies fine once #209 is merged, or it's a trivial rebase. Will check then

sdroege:
This should be good to go now. Rebased

GuillaumeGomez:
Later on, it'd be nice to have conversions with rust std equivalents. But that's completely ok to do it in another PR. Just waiting for @EPashkin since he said he wanted to check again. :)

EPashkin:
@sdroege Thanks.
Only nit is that `generate` array in `Gir.toml` need sort.

sdroege:
Done :)

GuillaumeGomez:
Thanks!

PR #209, opened 2017-07-31
Regen

EPashkin:
Apply https://github.com/gtk-rs/gir/pull/419, https://github.com/gtk-rs/gir/pull/418 and appveyour version fix


EPashkin:
cc @GuillaumeGomez, @sdroege

src/auto/functions.rs
```diff
@@ -170,11 +178,11 @@ pub fn close(fd: i32) -> Result<(), Error> {
 //    unsafe { TODO: call ffi::g_compute_checksum_for_bytes() }
 //}
 
-//pub fn compute_checksum_for_data(checksum_type: /*Ignored*/ChecksumType, data: /*Unimplemented*/&CArray TypeId { ns_id: 0, id: 3 }) -> Option<String> {
+//pub fn compute_checksum_for_data(checksum_type: /*Ignored*/ChecksumType, data: &[u8]) -> Option<String> {
 //    unsafe { TODO: call ffi::g_compute_checksum_for_data() }
 //}
 
-//pub fn compute_checksum_for_string(checksum_type: /*Ignored*/ChecksumType, str: &str, length: isize) -> Option<String> {
+//pub fn compute_checksum_for_string(checksum_type: /*Ignored*/ChecksumType, str: &str) -> Option<String> {
```
sdroege: This can easily be enabled now btw

EPashkin: Thanks, added.

sdroege:
Looks good :) I also thought of GCheckSum but that can also be added later when needed

EPashkin:
I missed `GCheckSum`, but it `g_checksum_get_digest` has wrong gir definition, it can be used without this function?

EPashkin:
As `compute_checksum_for_XX` return string, seems `GCheckSum` still usable without bad function, so I add it too.

sdroege:
Looks good but get_digest would be useful... I guess that needs a manually defined extern "C" function in glib-rs as a workaround for now :(

sdroege:
Good to go in any case, unless you want to add that manual get_digest implementation :)

EPashkin:
I add it at evening with tests for `Checksum`

src/checksum.rs
```diff
@@ -0,0 +1,83 @@
...
+            );
+
+            vec.set_len(digest_len);
+            vec
+        }
+    }
+
+    pub fn get_string(self) -> Option<String> {
+        unsafe {
+            from_glib_none(ffi::g_checksum_get_string(
```
EPashkin: Made this function manual co consume `Checksum` and prevent next `update`s

EPashkin:
@GuillaumeGomez, @sdroege  Not sure that it right way implement `get_digest`.
Also no way to get type from `Checksum` and calculate length, so I used 4096 as initial buffer size.

EPashkin:
Maybe we need add `cargo test` for travis too.

EPashkin:
Added version for `Sha512` https://developer.gnome.org/glib/stable/glib-Data-Checksums.html#GChecksumType

src/checksum.rs
sdroege: Makes sense

src/checksum.rs
```diff
@@ -0,0 +1,83 @@
...
+use libc::size_t;
+use Checksum;
+use ffi;
+use translate::*;
+
+impl Checksum {
+    pub fn get_digest(self) -> Vec<u8> {
+        unsafe {
+            let mut digest_len: size_t = 4096;
+            let mut vec = Vec::with_capacity(digest_len as usize);
```
sdroege: You can use CheckSum::type_get_length() here for knowing the size in bytes

src/checksum.rs
```diff
@@ -0,0 +1,83 @@
...
+
+            ffi::g_checksum_get_digest(
+                mut_override(self.to_glib_none().0),
+                vec.to_glib_none().0,
```
sdroege: vec.as_mut_ptr() seems more correct here

src/checksum.rs
```diff
@@ -0,0 +1,83 @@
...
+        assert_eq!(vec[6], 0x0d);
+        assert_eq!(vec[7], 0x30);
+        assert_eq!(vec[8], 0x87);
+        assert_eq!(vec[9], 0xd5);
+        assert_eq!(vec[10], 0x15);
+        assert_eq!(vec[11], 0xc0);
+        assert_eq!(vec[12], 0x47);
+        assert_eq!(vec[13], 0x3f);
+        assert_eq!(vec[14], 0x86);
+        assert_eq!(vec[15], 0x77);
```
sdroege: For simplicity, you could define an `[u8]` array with all those bytes and then directly compare the two.

src/checksum.rs
sdroege: Ah I see... You could store the type together with the GCheckSum I guess, but this is all becoming ugly. You could use the size of G_CHECKSUM_SHA512 for now... and we need to be careful when new variants are added.

EPashkin:
@sdroege, Fixed, Thanks.

src/checksum.rs
EPashkin: Can't use `type_get_length` for `Sha512` too as it versioned.

EPashkin: Can try get it from from pointer as it type is first element.

sdroege: It might change at any point in time to a different position :) Just use 512/8 for now then?

src/auto/enums.rs
```diff
@@ -8,6 +8,47 @@ use translate::*;
...
+impl ToGlib for ChecksumType {
+    type GlibType = ffi::GChecksumType;
+
+    fn to_glib(&self) -> ffi::GChecksumType {
+        match *self {
+            ChecksumType::Md5 => ffi::G_CHECKSUM_MD5,
+            ChecksumType::Sha1 => ffi::G_CHECKSUM_SHA1,
+            ChecksumType::Sha256 => ffi::G_CHECKSUM_SHA256,
+            #[cfg(feature = "v2_36")]
+            ChecksumType::Sha512 => ffi::G_CHECKSUM_SHA512,
```
sdroege: There's also G_CHECKSUM_SHA384 since 2.51. I think you should update your .gir files at some point in the near future :)

src/checksum.rs
EPashkin: Getting type just not work as `CheckSum` defined in other file.
You think that 4096 too big?

sdroege: It's a whole page, while currently we only need at most 64 bytes :)

src/auto/enums.rs
EPashkin: Yes, we will update it.

EPashkin:
Changed buffer size

sdroege:
:+1: if you also add a comment about that it has to be updated whenever a bigger hash is supported

EPashkin:
That will do?

sdroege:
Sure

GuillaumeGomez:
Well, I didn't have any issue at the start so if it's ok for @sdroege too then let's merge! :D

PR #214, opened 2017-08-17
Be more precise when setting a property with invalid type

fengalin:
When setting an `Object` property with a valid name but an invalid type, `"Invalid property name"` is returned. This modification allows returning `"Invalid property type"` in this case.

Changes return type for `Object::has_property`.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -561,7 +561,7 @@ pub trait ObjectExt: IsA<Object> {
 
     fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &Value) -> Result<(), BoolError>;
     fn get_property<'a, N: Into<&'a str>>(&self, property_name: N) -> Result<Value, BoolError>;
-    fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> bool;
+    fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> Result<(), BoolError>;
     fn get_property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type>;
 
     fn connect<'a, N, F>(&self, signal_name: N, after: bool, callback: F) -> Result<u64, BoolError>
@@ -583,8 +583,8 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
     fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &Value) -> Result<(), BoolError> {
         let property_name = property_name.into();
 
-        if !self.has_property(property_name, Some(value.type_())) {
-            return Err(BoolError("Invalid property name"));
+        if let Err(error) = self.has_property(property_name, Some(value.type_())) {
+            return Err(error);
         }
 
         unsafe {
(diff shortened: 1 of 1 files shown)


src/object.rs
```diff
@@ -628,6 +629,20 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
         }
     }
 
+    fn has_property_with_type<'a, N: Into<&'a str>>(&self, property_name: N, type_: Type) -> Result<(), BoolError> {
+        let property_name = property_name.into();
+        let ptype = self.get_property_type(property_name);
+
+        match ptype {
+            None => Err(BoolError("Invalid property name")),
+            Some(ptype) => if ptype == type_ {
```
GuillaumeGomez: Please put this one in a block, it's not really beautiful like this (even though it's perfectly correct).

GuillaumeGomez:
This is awesome! Just a nit and it's good for me.

cc @EPashkin @sdroege

src/object.rs
fengalin: Yeah, I agree! I'm just not too sure about what the custom is. Would that be better:
``` rust
        if let Some(ptype) = ptype {
            if ptype == type_ {
                Ok(())
            } else {
                Err(BoolError("Invalid property type"))
            }
        } else {
            Err(BoolError("Invalid property name"))
        }
```

GuillaumeGomez: Not worst at least so let's go for this solution! :)

fengalin: Ça marche ! ;)

GuillaumeGomez: Keep it in english so other people can follow as well please (même si je pense que tout le monde aura compris ce que t'as dit / even though I think everyone understood what you wrote). :)

sdroege:
It's fine to change API IMHO, these functions are all new and in no release yet

sdroege:
I.e., please just change the existing function :)

fengalin:
All right, I'll do that!

fengalin:
By the way, any suggestion on the "ugly part"?

sdroege:
What ugly part?

fengalin:
It used to be:
``` rust
        match ptype {
            None => Err(BoolError("Invalid property name")),
            Some(ptype) => if ptype == type_ {
                    Ok(())
                } else {
                    Err(BoolError("Invalid property type"))
                },
        }
```
I must admit, I was not too proud of it. Then upon @GuillaumeGomez request, I proposed this:
``` rust
        if let Some(ptype) = ptype {
            if ptype == type_ {
                Ok(())
            } else {
                Err(BoolError("Invalid property type"))
            }
        } else {
            Err(BoolError("Invalid property name"))
        }
```
But I'm still not convinced.

fengalin:
Never mind... I got it known.

sdroege:
Looks good to me

EPashkin:
LGFM, only commit title better changed.

fengalin:
Do you want me to squash the commits and keep the first commit message only?

GuillaumeGomez:
Yes please!

fengalin:
Not sure if you want me to do something else on this one.

@EPashkin: shall I enhance "Be more precise when setting a property with invalid type"?

EPashkin:
Yes, after squash all good for me. Thanks.

GuillaumeGomez:
Well, he already squashed so let's merge. Thanks!

Issue #224, opened 2017-08-30
source::CallbackGuard should probably print something to stderr before exiting


sdroege:
Otherwise the process just silently exits with error code 101 without anybody knowing what's going on.

EPashkin:
Thanks for noticing this.

sdroege:
Would be nice to put this into a bugfix release of glib, once there is something more important that requires a bugfix release

GuillaumeGomez:
Yep. There is a small fix in `gtk` as well. I'll make a small release in a week or two.

PR #225, opened 2017-08-30
For the CallbackGuard, print something to stderr and use abort() inst…

sdroege:
…ead of exit()

Otherwise the process just silently exits without anybody knowing what's
wrong. With abort() we also make sure that no at_exit handlers are
called and it's easier to attach a debugger and/or get a coredump.

Fixes https://github.com/gtk-rs/glib/issues/224

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -71,8 +71,12 @@ impl Default for CallbackGuard {
 
 impl Drop for CallbackGuard {
     fn drop(&mut self) {
+        use std::io::stderr;
+        use std::io::Write;
+
         if thread::panicking() {
-            process::exit(101);
+            let _ = stderr().write(b"Uncaught panic, exiting");
+            process::abort();
         }
     }
 }


src/source.rs
```diff
@@ -71,8 +71,12 @@ impl Default for CallbackGuard {
 
 impl Drop for CallbackGuard {
     fn drop(&mut self) {
+        use std::io::stderr;
+        use std::io::Write;
+
         if thread::panicking() {
-            process::exit(101);
+            let _ = stderr().write(b"Uncaught panic, exiting");
```
sdroege: Anything more useful to print here, anything to add?

EPashkin: IMHO nothing.

EPashkin:
👍

GuillaumeGomez:
Thanks!

src/source.rs
sdroege: Actually something. It needs a \n at the end. @GuillaumeGomez , can you just push that simple change? Going through a PR seems a bit too much overhead for one character :)

GuillaumeGomez: Ah sure.

PR #226, opened 2017-08-31
Regen

EPashkin:
cc @GuillaumeGomez @sdroege


GuillaumeGomez:
Thanks!

src/auto/enums.rs
```diff
@@ -14,6 +14,7 @@ pub enum ChecksumType {
     Sha256,
     #[cfg(feature = "v2_36")]
     Sha512,
+    Sha384,
```
sdroege: This is missing a version attribute

GuillaumeGomez: I didn't notice this one... Thanks for finding it!

EPashkin: Thanks, @sdroege, I forgot about this

Issue #229, opened 2017-09-01
Use newtype wrapper around u64 for signal handler ids


sdroege:
I.e. what the connect functions return. These are not arbitrary u64 but have a meaning, and we probably want to prevent people from mixing these with other u64 (or doing arithmetic on them, etc).

Might also make sense to *not* implement Copy/Clone on them (and the same for SourceId), as you can only really disconnect a signal handler once.

EPashkin:
👍 for defining new type.
Not sure about disabling copy/clone.

sdroege:
@GuillaumeGomez Do you have an opinion on Copy/Clone? Once that is solved I'll implement this

GuillaumeGomez:
We should remain coherent with the rest: if it's not supposed to be in two places at once, then let's not implement Copy/Clone.

sdroege:
You can only disconnect it once, and if you disconnect it twice you might accidentially disconnect a signal handler that was added after your first disconnect (or get a warning printed on the terminal).

Same thing for SourceId.

PR #231, opened 2017-09-07
Use a newtype wrapper around the signal handler IDs

sdroege:
commit b6ba3c6e23c638f1271efa6971646620bb2b4df9 (HEAD -> signal-handler-newtype)
Author: Sebastian Dröge <sebastian@centricular.com>
Date:   Thu Sep 7 19:55:40 2017 +0300

    Don't implement Copy/Clone for SignalHandlerId and SourceId
    
    These can only really be used once.

commit 3bfe7525982c41adab0ab7ccd584db4f674b4c3d
Author: Sebastian Dröge <sebastian@centricular.com>
Date:   Thu Sep 7 19:53:11 2017 +0300

    Use a newtype wrapper around the signal handler IDs
    
    Fixes https://github.com/gtk-rs/glib/issues/229

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -18,6 +18,7 @@ use value::SetValue;
 use Type;
 use BoolError;
 use Closure;
+use SignalHandlerId;
 
 /// Upcasting and downcasting support.
 ///
@@ -564,7 +565,7 @@ pub trait ObjectExt: IsA<Object> {
     fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> Result<(), BoolError>;
     fn get_property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type>;
 
-    fn connect<'a, N, F>(&self, signal_name: N, after: bool, callback: F) -> Result<u64, BoolError>
+    fn connect<'a, N, F>(&self, signal_name: N, after: bool, callback: F) -> Result<SignalHandlerId, BoolError>
         where N: Into<&'a str>, F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static;
     fn emit<'a, N: Into<&'a str>>(&self, signal_name: N, args: &[&Value]) -> Result<Option<Value>, BoolError>;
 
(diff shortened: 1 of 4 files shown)


sdroege:
@GuillaumeGomez @EPashkin Any comments on this? Once this and the gir change landed, we need to regenerate all bindings that come with signals.

GuillaumeGomez:
Ah sorry, I said it on IRC but I repeat here as well: I like it. All good for me!

src/object.rs
```diff
@@ -702,7 +703,7 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
             if handler == 0 {
                 Err(BoolError("Failed to connect to signal"))
             } else {
-                Ok(handler as u64)
+                Ok(from_glib(handler))
```
EPashkin: `handler as u64`?

EPashkin: I meant: you lost conversion.

sdroege: No, it's internally all based on c_ulong now again, like the FFI functions. No cast is needed here unless you meant something else.

EPashkin: Seems this not fully so:
```
error[E0277]: the trait bound `signal::SignalHandlerId: translate::FromGlib<u64>` is not satisfied
  --> src\signal.rs:57:5
   |
57 |     from_glib(handle)
   |     ^^^^^^^^^ the trait `translate::FromGlib<u64>` is not implemented for `signal::SignalHandlerId`
   |
             <signal::SignalHandlerId as translate::FromGlib<u32>>
   = note: required by `translate::from_glib`
```

sdroege: There was a cast right in the line above, which would have to be removed. Done! Thanks for noticing

sdroege: FWIW, this would've only failed on Windows (32/64 bit) and 32 bit variants of other platforms. Elsewhere c_ulong and u64 are equivalent.

EPashkin:
👍 for merge.

EPashkin:
Or it need regen with https://github.com/gtk-rs/gir/pull/454?

sdroege:
It needs regen with https://github.com/gtk-rs/gir/pull/453

GuillaumeGomez:
Let's wait for the regen then! :)

sdroege:
The regens won't compile without having this GLib change merged first.

sdroege:
@GuillaumeGomez Or do you want all regens first, not compiling, and then merge it all in one go?

GuillaumeGomez:
It's always so annoying... Let's merge this one first!

sdroege:
It definitely is, yes. Once you get https://github.com/gtk-rs/gir/pull/454 and https://github.com/gtk-rs/glib/pull/232 in, I'll send regens for everything. Otherwise there would have to be regens twice

PR #234, opened 2017-09-14
Allow specifying class struct to glib_wrapper!() macro

federicomenaquintero:
This goes along with https://github.com/gtk-rs/gir/pull/459

I'm not 100% sure if this works yet, as my gtk module doesn't build (but glib builds).  Consider it as a proof of concept for now.

### DIFF
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -10,7 +10,7 @@
 ///
 /// ```ignore
 /// glib_wrapper! {
-///     /// Documentation
+///     /// Your documentation goes here
 ///     pub struct $name($kind<$foreign>);
 ///
 ///     match fn {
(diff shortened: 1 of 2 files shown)


src/wrapper.rs
```diff
@@ -112,9 +113,24 @@
 /// }
 /// ```
 ///
+/// #### Non-derivable classes
+///
+/// By convention, GObject implements "final" classes, i.e. those who cannot
+/// be subclassed, by exposing a public Instance struct, but no corresponding
```
sdroege: Note that the instance struct could also be opaquely defined (i.e. only a pointer typedef in the public header, with the actual struct definition hidden). This is not a problem though and handled

EPashkin:
While this PR still not uses `GlibClassType` I vote for merge it as is to fix block for future regens.
@GuillaumeGomez, @sdroege what you think about it?

sdroege:
I would like it if there was a commit that makes glib_wrapper! also emit code like the following

```Rust

#[repr(C)]
pub struct FooClass(ffi::BarFooClass);

impl Foo {
    pub fn get_class(&self) -> &FooClass {
        unsafe {
            let stash = self.to_glib_none();
            let ptr: ffi::Foo = stash.0;
            let class = *(ptr as *const *const ffi::BarFooClass);
            &*(class as *const FooClass)
        }
}
```

This would lead the way to having immutable class methods later (g_object_class_list_properties(), gst_element_class_get_pad_template(), etc). Same for mutable ones (to be called in class_init / base_init) would be something that could be added too then, but would only be useful for myself and the gnome-class macro :)

EPashkin:
Good idea

federicomenaquintero:
Could we get this merged and postpone Sebastian's idea until we decide on the following?

As far as I can tell, generating `pub struct FooClass` requires something like the `concat_idents!()` macro, [which is a nightly-only feature][concat_idents].

It may be possible to also have an internal procedural macro inside glib-rs itself that generates the FooClass identifier.

This is a bit more work than I'd like to put in just right now... maybe we can do this during the hackfest in Berlin?

[concat_idents]: https://doc.rust-lang.org/std/macro.concat_idents.html

EPashkin:
👍 for merge this and related gir's PR as is.
This will need full Regen again, but all CI IMHO will pass.

GuillaumeGomez:
Fine by me. Thanks!

federicomenaquintero:
Thank you!

Issue #223, opened 2017-08-30
Closure::emit() / ObjectExit::emit() could use stack allocated array instead of Vec


sdroege:
https://github.com/gtk-rs/glib/blob/e46aa7f27cc74f3cdcb54d94feb4a8861df21c7f/src/closure.rs#L77
https://github.com/gtk-rs/glib/blob/e46aa7f27cc74f3cdcb54d94feb4a8861df21c7f/src/object.rs#L743

It would make sense to use a stack allocated array here for < N parameters to make things more efficient and not have that many allocations per call. N = 10 for example would be sufficient already for 99% of the cases, and only uses up 10*(8+8+8) == 240 bytes of stack memory.

EPashkin:
Agree

federicomenaquintero:
FWIW, I got pointed the other day to https://crates.io/crates/smallvec which does exactly this :)

sdroege:
As this is all internal for us, we can just manually write the code for that. It's adding about 4 lines to each of these functions :)

Issue #228, opened 2017-09-01
Use ToValue trait for ObjectExt::set_property()


sdroege:
https://github.com/gtk-rs/glib/blob/ba748d15e607daf4a3ef103e7e83333466d7c9eb/src/object.rs#L562-L564

Either as trait objects (which fortunately work for ToValue), or generics. Neither doesn't really make a difference, for you to decide.

Could also use ToValue for ::emit() at least, but as this is more of a plumbing function it is maybe not too useful. And for ::connect() we can't do it anyway, so maybe let's keep the symmetry there.

Comments?

sdroege:
OTOH get_property() must return a Value, so we would lose symmetry there too. I'm fine with keeping things as is.

EPashkin:
It is good idea.

sdroege:
You think the asymmetry is not a problem?

EPashkin:
Yes, it normal for rust, ex. `AsRef<Path>` vs `PathBuf` or even more common `&str` vs `String` :wink:

PR #238, opened 2017-09-29
Value usage improvements

sdroege:
Fixes https://github.com/gtk-rs/glib/issues/223 and https://github.com/gtk-rs/glib/issues/228

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -14,7 +14,7 @@ use std::ptr;
 use std::marker::PhantomData;
 
 use Value;
-use value::SetValue;
+use value::{ToValue, SetValue};
 use Type;
 use BoolError;
 use Closure;
@@ -560,7 +560,7 @@ glib_object_wrapper! {
 pub trait ObjectExt: IsA<Object> {
     fn get_type(&self) -> Type;
 
-    fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &Value) -> Result<(), BoolError>;
+    fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &ToValue) -> Result<(), BoolError>;
     fn get_property<'a, N: Into<&'a str>>(&self, property_name: N) -> Result<Value, BoolError>;
     fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> Result<(), BoolError>;
     fn get_property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type>;
@@ -581,13 +581,14 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
         }
     }
 
-    fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &Value) -> Result<(), BoolError> {
+    fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &ToValue) -> Result<(), BoolError> {
         let property_name = property_name.into();
 
-        if let Err(error) = self.has_property(property_name, Some(value.type_())) {
+        if let Err(error) = self.has_property(property_name, Some(value.to_value_type())) {
             return Err(error);
         }
 
+        let value = value.to_value();
         unsafe {
             gobject_ffi::g_object_set_property(self.to_glib_none().0, property_name.to_glib_none().0, value.to_glib_none().0)
         }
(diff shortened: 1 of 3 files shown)


EPashkin:
and maybe even more generalized by adding function parameter

sdroege:
What would the signature of the function you think of look like?

EPashkin:
I thought about something like this, but this version always allocate vector in stack,
as I don't see way to fill uninitialized vector.
```
fn func<'a, T,U, V: Fn(&U)->T>(f:V, values: &[U], stack: &'a mut[T], heap: &'a mut Vec<T>) -> &'a[T] {
  if values.len() <= stack.len() {
      for (i, arg) in values.iter().enumerate() {
          stack[i] = f(arg);
      }
      &stack[0..values.len()]
  } else {
      heap.reserve(stack.len());
      for (i, arg) in values.iter().enumerate() {
          heap[i] = f(arg);
      }
      heap.as_slice()
  }
}
```

sdroege:
Note that this wouldn't cover both variants. The one in object.rs puts the instance argument as first, additional element into the vector.

src/object.rs
```diff
@@ -736,24 +737,35 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
-            c_args.push(ptr::read(instance.to_glib_none().0));
-            for arg in args {
-                c_args.push(ptr::read(arg.to_glib_none().0));
-            }
+            let mut v_args: Vec<Value>;
+            let mut s_args: [Value; 10] = mem::zeroed();
+            let args = if args.len() < 10 {
+                s_args[0] = Value::from(self);
+                for (i, arg) in args.iter().enumerate() {
+                    s_args[i] = arg.to_value();
```
EPashkin: This not replace s_args[0] ?

EPashkin:
Yes, you right. `values` can be changed to `ExactSizeIterator`.

EPashkin:
.. and second version use `std::iter::once` and `chain`.

EPashkin:
Maybe generalizing not worth memory or speed-wise, so I don't insist.
Also you can try https://crates.io/crates/smallvec as it seems doing almost same.

sdroege:
It seemed like too much to depend on an external crate just for 4 lines of code :) but for refactoring the right solution would be a data type and not a function

Thanks for your other comments, I'll clean it up tomorrow

src/object.rs
```diff
@@ -736,24 +738,33 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+            let mut v_args: Vec<Value>;
+            let mut s_args: [Value; 10] = mem::zeroed();
+            let args = if args.len() < 10 {
+                for (i, arg) in iter::once(&(self as &ToValue)).chain(args).enumerate() {
+                    s_args[i] = arg.to_value();
+                }
+                &s_args[0..args.len()+1]
+            } else {
+                v_args = Vec::with_capacity(args.len() + 1);
+                for arg in iter::once(&(self as &ToValue)).chain(args) {
```
sdroege: I'm not sure how to get rid of the `&(self as &ToValue)` disaster here. Any suggestions?

EPashkin: Why it disaster?
By way, I proposed use `iter::once` for function, if used directly then old variant looks normal (after index fix and `with_capacity` on second branch)

sdroege: It looks ugly but works :) do you prefer it like this or the previous version?

EPashkin: IMHO almost all same as long as it works

src/value.rs
```diff
@@ -93,6 +93,7 @@ use gobject_ffi;
 /// (e.g. numeric types) don't.
 ///
 /// See the [module documentation](index.html) for more details.
+#[repr(C)]
 pub struct Value(gobject_ffi::GValue);
```
sdroege: Let's get this in too then, this is not a new requirement for this PR but already before Closure and other places casted between glib::Value and gobject_ffi::GValue for preventing copies.

src/object.rs
sdroege: 👍 Let's keep it as now then

src/value.rs
EPashkin: IMHO better be `repr(internal)` but now I can't find it at all.

EPashkin:
@sdroege Thanks
👍 for merge.

GuillaumeGomez:
Let's merge then. Thanks to both of you!

PR #240, opened 2017-10-12
Some fixes in the Signals handling

thiblahute:
Namely:
  - Make all the Signal related APIs usable as Object methods
  - Fix signal_handler_(un)block arguments

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -566,9 +566,14 @@ pub trait ObjectExt: IsA<Object> {
     fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> Result<(), BoolError>;
     fn get_property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type>;
 
+    fn block_signal(&self, handler_id: &SignalHandlerId);
+    fn unblock_signal(&self, handler_id: &SignalHandlerId);
+    fn stop_signal_emission(&self, signal_name: &str);
+
     fn connect<'a, N, F>(&self, signal_name: N, after: bool, callback: F) -> Result<SignalHandlerId, BoolError>
         where N: Into<&'a str>, F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static;
     fn emit<'a, N: Into<&'a str>>(&self, signal_name: N, args: &[&ToValue]) -> Result<Option<Value>, BoolError>;
+    fn disconnect(&self, handler_id: SignalHandlerId);
 
     fn downgrade(&self) -> WeakRef<Self>;
 }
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Wo nice!

sdroege:
:+1: as discussed on irc earlier :) thanks!

src/object.rs
```diff
@@ -566,6 +566,11 @@ pub trait ObjectExt: IsA<Object> {
     fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> Result<(), BoolError>;
     fn get_property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type>;
 
+    fn block_signal(&self, handler_id: &SignalHandlerId);
+    fn unblock_signal(&self, handler_id: &SignalHandlerId);
+    fn stop_signal_emission(&self, signal_id: u32, detail: glib_ffi::GQuark);
```
sdroege: No ffi types in the API here. We need GQuark bindings, or pass &str here and convert internally

thiblahute: Should I just expose `stop_signal_emission_by_name` then? Renamed to `stop_signal_emission` maybe?

sdroege: Yeah, let's do the quark stuff once we have bindings for that. That's also how I did it for the signal emitter and connector function

sdroege: I mean, also rename for now. We can break API later once quarks are there

src/object.rs
```diff
@@ -566,6 +566,11 @@ pub trait ObjectExt: IsA<Object> {
...
+    fn stop_signal_emission_by_name(&self, signal_name: &str);
+
     fn connect<'a, N, F>(&self, signal_name: N, after: bool, callback: F) -> Result<SignalHandlerId, BoolError>
```
sdroege: Should probably also add a `disconnect` here for completeness

src/signal.rs
```diff
@@ -57,13 +57,13 @@ pub unsafe fn connect(receiver: *mut gobject_ffi::GObject, signal_name: &str, tr
     from_glib(handle)
 }
 
-pub fn signal_handler_block<T: IsA<Object>>(instance: &T, handler_id: SignalHandlerId) {
+pub fn signal_handler_block<T: IsA<Object>>(instance: &T, handler_id: &SignalHandlerId) {
```
EPashkin: Why `&` added?

EPashkin: Sorry, now understand, it because it non-copy.

src/object.rs
thiblahute: Done.

src/object.rs
thiblahute: Done.

sdroege:
Good to be merged IMHO

GuillaumeGomez:
For me too. What about you @EPashkin?

EPashkin:
👍 too

sdroege:
@GuillaumeGomez let's get it in then :)

PR #244, opened 2017-11-10
Allow generating a class struct wrapper

sdroege:
…mutable reference to it from the object

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -6,7 +6,7 @@
 
 use translate::*;
 use types::{self, StaticType};
-use wrapper::{UnsafeFrom, Wrapper};
+use wrapper::{UnsafeFrom, ClassWrapper, Wrapper};
 use ffi as glib_ffi;
 use gobject_ffi;
 use std::mem;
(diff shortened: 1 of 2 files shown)


src/object.rs
```diff
@@ -584,11 +607,8 @@ pub trait ObjectExt: IsA<Object> {
 
 impl<T: IsA<Object> + SetValue> ObjectExt for T {
     fn get_type(&self) -> Type {
-        unsafe {
-            let obj = self.to_glib_none().0;
-            let klass = (*obj).g_type_instance.g_class as *mut gobject_ffi::GTypeClass;
-            from_glib((*klass).g_type)
-        }
+        let klass = self.get_class();
+        from_glib(klass.0.g_type_class.g_type)
```
sdroege: Here is even some usage of it already, removing some unsafe code

src/object.rs
```diff
@@ -859,3 +878,17 @@ impl<T: IsA<Object> + ?Sized> Clone for WeakRef<T> {
 
 unsafe impl<T: IsA<Object> + Send + Sync + ?Sized> Send for WeakRef<T> {}
 unsafe impl<T: IsA<Object> + Send + Sync + ?Sized> Sync for WeakRef<T> {}
+
+pub trait Classed<C>
+  where C: Wrapper {
+    fn get_class(&self) -> &C::ClassType;
+}
```
sdroege: New trait that only exists so we can implement it generically below for all GObjects

sdroege:
More work needed here also, just comment on it already if you have any comments

sdroege:
Not needed anymore, this is implemented now and all good :)

PR #248, opened 2017-11-12
Add Char wrapper; implement KeyFile::set_list_separator() with it

federicomenaquintero:
I'm in the process of adding "impl ToGlib" for all the fundamental types, so that gnome-class will have an easier time passing them around: then it can just call .to_glib() on every method argument with a fundamental type.

Along the way I ran into the distinction between Rust's `char` and Glib's C `c_char`.  Here we introduce a `Char` wrapper, which can be created fallibly from a Rust `char`.  When it succeeds, the result can be passed to C functions that take a `c_char`.

Finally, we implement `KeyFile::set_list_separator` in terms of this.


src/char.rs
```diff
@@ -0,0 +1,79 @@
...
+/// This function plainly expects a byte as the `separator` argument.  However,
+/// having this function exposed to Rust as the following would be inconvenient:
+///
+/// ```ignore
+/// impl KeyFile {
+///     pub fn set_list_separator(&self, separator: i8) { }
+/// }
+/// ```
+///
+/// This would be inconvenient because users would have to do the conversion from a Rust `char` to an `i8` by hand.  It would be the same case for `libc::c_char`, which is just a type alias
```
sdroege: They could pass `b'a'` for example

Gir.toml
```diff
@@ -252,6 +252,10 @@ status = "generate"
        # file parameter needs to be a PathBuf
        ignore = true
        [[object.function]]
+       name = "set_list_separator"
+       # separator parameter neds to be a glib::Char
```
EPashkin: `need`

EPashkin:
Better do regen with same `gir` version if possible.
IMHO `Char` need `ToGlib/FromGlib` for `gir` can generate functions (and this too).
I don't found comments about `Fundamental: UChar` in gtk-rs, but this type maybe also needed.

EPashkin:
@federicomenaquintero I'm very sorry for long delay.
You agree to add `UChar` type with same functionality?
I almost "fixed" generator, so manual implementation for `KeyFile::set_list_separator` will not needed.

EPashkin:
@GuillaumeGomez, seems @federicomenaquintero don't have time, maybe we just merge PR as is and then I make PR that add `UChar` and `KeyFile::set_list_separator` generation?

GuillaumeGomez:
Sure, let's do it like that then. Thanks everyone!

PR #249, opened 2017-11-12
Also require ToGlibPtr<_, *const _> for the IsA<_> trait

sdroege:
The macro is implementing both anyway, and we might need to be able to
generate const pointers for various FFI API.

@zeenix

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -121,11 +121,13 @@ impl<T: IsA<Object>> Cast for T { }
 /// `T` always implements `IsA<T>`.
 pub trait IsA<T: StaticType + UnsafeFrom<ObjectRef> + Wrapper>: StaticType + Wrapper +
     Into<ObjectRef> + UnsafeFrom<ObjectRef> +
-    for<'a> ToGlibPtr<'a, *mut <T as Wrapper>::GlibType> { }
+    for<'a> ToGlibPtr<'a, *mut <T as Wrapper>::GlibType> +
+    for<'a> ToGlibPtr<'a, *const <T as Wrapper>::GlibType> { }
 
 impl<T> IsA<T> for T
 where T: StaticType + Wrapper + Into<ObjectRef> + UnsafeFrom<ObjectRef> +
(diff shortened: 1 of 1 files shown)


sdroege:
This now breaks type inference in various places. Do we want that?

zeenix:
I sent a patch to fix this upstream in libosinfo so let's see if that get's merged instead: 

https://www.redhat.com/archives/libosinfo/2017-November/msg00008.html

EPashkin:
I don't found any usage of `*const` in generated code.
In manual code it seems usable (if works) only in `sourceview/src/style.rs` (not checked `gstreamer-rs`).
So seems mostly usage lies in glib.

EPashkin:
@zeenix Can you show example of function with `const` that generated bad way?

zeenix:
@EPashkin see the link to my patch email above for such functions in upstream. Do you want me to paste the generated bad code?

EPashkin:
@zeenix Yes, I on windows, so I can't generate it easily, so please, show me bad code.

EPashkin:
.. with part of  .gir-file for this function.

EPashkin:
Found working version for one function (with restore constness in sys):
```
    pub fn osinfo_install_config_param_get_name(config_param: *const OsinfoInstallConfigParam) -> *const c_char;

    fn get_name(&self) -> Option<String> {
        unsafe {
            let stash:Stash<*mut _, _> = self.to_glib_none();
            from_glib_none(ffi::osinfo_install_config_param_get_name(stash.0 as *const _))
        }
    }
```
Theoretically we can change `gir` to produce this for `&self` but better find better way.

EPashkin:
@zeenix Gir updated to fix your problem, please check if it really helps.

zeenix:
@EPashkin thanks so much. The patch to libosinfo was in the end accepted actually so it's not that much an issue anymore but good to have the fix for existing libosinfo releases.

zeenix:
I tried to quickly test but there are currently other issues I'm facing with libosinfo-rs crate. I'll try to check later but feel free to close this issue.

sdroege:
Outdated and not really needed anymore

PR #250, opened 2017-11-12
Change `CallbackGuard` to require explicit destruction

alexcrichton:
This commit alters the API of the `CallbackGuard` in a semver-breaking
way. Previously the guard would abort the process if the current thread
were panicking when it was dropped. After this change, however, the
`thread::panicking` function isn't used and instead an explicit call to
destruction is required. If a `CallbackGuard` is dropped "normally" (aka
the compiler auto-drops it) then that likely indicates a panic (or a
bug) and the process will be aborted.

This tweak brings with it a few benefits:

* LLVM can optimize this much better as it can infer that a chunk of
  code between the creation of a guard and its drop doesn't panic.
  Previously it would not optimize away creation/dropping the guard but
  after this change LLVM should be able to optimize this away. Put
  another way, after this commit LLVM can optimize away usage of
  `CallbackGuard` if it can prove the code won't panic.

* Runtime-wise this should be a bit cheaper on the "fast path" as
  checking `thread::panicking` requires... well... at least a function
  call! By calling `defuse` though it's a noop at runtime which means
  that no extra code happens on the fast path when a panic doens't
  happen.

* Finally, a slightly more esoteric benefit, but if a thread is
  panicking *already* and during a destructor some glib calls were made,
  this strategy will no longer abort the process. Ideally the callback
  guard only aborts the process if the state of the thread panicking
  *changes* rather than only is true on exit. This is likely to only
  come up pretty rarely, though!

### DIFF
--- a/src/main_context.rs
+++ b/src/main_context.rs
@@ -54,16 +54,18 @@ impl MainContext {
 
 #[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
 unsafe extern "C" fn trampoline(func: gpointer) -> gboolean {
-    let _guard = CallbackGuard::new();
+    let guard = CallbackGuard::new();
     let func: &RefCell<Box<FnMut() + 'static>> = transmute(func);
     (&mut *func.borrow_mut())();
+    guard.defuse();
 
     glib_ffi::G_SOURCE_REMOVE
 }
 
 unsafe extern "C" fn destroy_closure(ptr: gpointer) {
-    let _guard = CallbackGuard::new();
+    let guard = CallbackGuard::new();
     Box::<RefCell<Box<FnMut() + 'static>>>::from_raw(ptr as *mut _);
+    guard.defuse();
 }
 
 fn into_raw<F: FnMut() + Send + 'static>(func: F) -> gpointer {
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -88,7 +88,8 @@ pub fn signal_stop_emission_by_name<T: IsA<Object>>(instance: &T, signal_name: &
 }
 
 unsafe extern "C" fn destroy_closure(ptr: *mut c_void, _: *mut gobject_ffi::GClosure) {
-    let _guard = CallbackGuard::new();
+    let guard = CallbackGuard::new();
     // destroy
     Box::<Box<Fn()>>::from_raw(ptr as *mut _);
+    guard.defuse();
 }
--- a/src/source.rs
+++ b/src/source.rs
@@ -3,9 +3,7 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use std::cell::RefCell;
-use std::mem::transmute;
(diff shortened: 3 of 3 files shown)


EPashkin:
@alexcrichton Thanks.
But IMHO this need be postponed until we do current planned release.

sdroege:
Yes, but let's get it in right after the next release. This requires changes in the code generator and manual code using this currently, but improves compiler output a lot.

GuillaumeGomez:
For the next release, I'd just like to have [this PR](https://github.com/gtk-rs/cairo/pull/165) getting merged first then I'll do it. It already has a lot changes and waiting more would be problematic I think.

sdroege:
From what I understand, all this catching of panics at the FFI boundary is not necessary anymore now that https://github.com/rust-lang/rust/pull/46833 is merged. We could get rid of all the `callback_guard!` everywhere, we only need to depend on a new-enough Rust (i.e. 1.23 or 1.24 in this case?).

@alexcrichton can you confirm?

alexcrichton:
@sdroege that is indeed correct!

sdroege:
Great, thanks for confirming :) then let's just go with that instead of making things more complex

GuillaumeGomez:
So what should we do in here? Moving forward?

sdroege:
With 1.24 we can get rid of the callback_guard stuff completely as the compiler takes care of that already.

sdroege:
1.24 is out, we can remove all the callback guards \o/

EPashkin:
@sdroege I will do it tomorrow, if you don't started already.

sdroege:
Please go ahead :) I would otherwise do it some time next week

EPashkin:
@GuillaumeGomez This seems can be closed now

GuillaumeGomez:
Indeed.

sdroege:
Unfortunately Rust 1.24.1 reverts this change because Windows proves to be an enemy of progress again: https://github.com/rust-lang/rust/issues/48251

So for 1.24 and 1.25 these changes should all be fine, for 1.24.1 not.

EPashkin:
Sad story

sdroege:
Or might not happen. Let's wait

sdroege:
I did, so we need to revert these until 1.25

EPashkin:
Or postpone crate release?

sdroege:
That would mean another 6 weeks or 12 in the worst case if they don't manage to put it into 1.25.

EPashkin:
@GuillaumeGomez so we revert changes in gir and all your crates?

GuillaumeGomez:
I suppose... I'm really depressed by this issue...

sdroege:
Me too, that's why I didn't send any PR yet. It's also IMHO the wrong decision by the Rust team. They break backwards compatible to make something work again that only worked by pure luck before due to how undefined behaviour was at that point.

GuillaumeGomez:
Yes, however we don't have much to say about this... I understand their position, I just don't agree with. :p

Issue #251, opened 2017-11-12
Provided a boxed type for a Box<Any>


sdroege:
This would be useful for passing around arbitrary Rust objects via GValue and similar things. It's very easy to implement.

@GuillaumeGomez @EPashkin Would you like to have this in glib?

EPashkin:
@sdroege can you give usage example?

GuillaumeGomez:
It'd be nice. :D

sdroege:
Like storing in a GTK list model :) Or in my case, I'd like to transport some Rust value in a GStreamer message

EPashkin:
Then it really useful. Include it to glib or not IMHO depends only on dependencies this needed.

sdroege:
No dependencies other than std or glib itself needed

PR #252, opened 2017-11-13
Implement AnyValue, a wrapper around std::any::Any

sdroege:
This allows to store any Rust type (that is 'static, Clone, Send, Sync,
Any) to be stored in GValues and passed around as GValues through C
code, while being able to take out the original values in Rust again.

Fixes https://github.com/gtk-rs/glib/issues/251

Let me know if this looks good to you, then I add documentation and especially a code example to the documentation. Test is included to prove that it actually works as expected.


GuillaumeGomez:
:+1:

src/value.rs
```diff
@@ -697,3 +699,139 @@ numeric!(i64, g_value_get_int64, g_value_set_int64);
...
+pub struct AnyValue {
+    val: Box<Any>,
+    copy_fn: Arc<Fn(&Any) -> Box<Any> + Send + Sync + 'static>,
+}
+
+impl AnyValue {
+    pub fn new<T: Any + Clone + Send + Sync + 'static>(val: T) -> Self {
+        let val: Box<Any> = Box::new(val);
+        let copy_fn = Arc::new(|val: &Any| {
+            let copy = val.downcast_ref::<T>().unwrap().clone();
```
EPashkin: May be better use `.expect` ?

EPashkin:
@sdroege Thanks

src/value.rs
GuillaumeGomez: Oh indeed!

sdroege: It can't possibly fail, but sure :)

sdroege:
I'll first fix https://github.com/gtk-rs/glib/issues/253, and then this here will get a SendAnyValue and AnyValue

sdroege:
Done now, this should be merged after https://github.com/gtk-rs/glib/pull/255 and https://github.com/gtk-rs/glib/pull/256

EPashkin:
@sdroege Thanks again
👍

GuillaumeGomez:
Conflict.

sdroege:
Fixed :)

EPashkin:
@sdroege can you add TODO  here? https://github.com/gtk-rs/glib/pull/256#discussion_r151182123

sdroege:
Done, and another little commit that helps usage of the new types

sdroege:
All good to go?

GuillaumeGomez:
It was for me before all these adds so waiting for @EPashkin's confirmation.

EPashkin:
@sdroege Third thanks ;)
Still 👍

GuillaumeGomez:
Thanks! :)

Issue #254, opened 2017-11-14
DerefMut on TypedValue is unsafe


sdroege:
```rust
let mut meh: TypedValue<i32> = (&123i32).into();
let meh2: Value = (&1u8).to_value();
*meh = meh2;
```

And suddenly we have a `TypedValue<i32>` that actually contains an `u8`.

sdroege:
Is `DerefMut` ever actually needed or can we just remove it? @GuillaumeGomez @EPashkin ? :)

GuillaumeGomez:
Oh damn. That's quite terrible! 😱 

And no idea about `DerefMut`...

sdroege:
I don't think `DerefMut` is of any use here, other than overwriting the value with a different one of potentially a different type. I wouldn't know how to use it otherwise at least.

EPashkin:
Agree that current version is unsafe.
Just removing it will broke `impl<'a, T: FromValueOptional<'a> + SetValue> TypedValue<T> {`,
so maybe it can be replaced with internal alternative.

sdroege:
Sure, that can easily be rewritten. I meant I see no usage *outside* value.rs :)

PR #255, opened 2017-11-15
Don't implement DerefMut on TypedValue

sdroege:
It allows replacing the underlying Value inside the TypedValue with one
    of a different type and breaks type-safety:
    
      let mut meh: TypedValue<i32> = (&123i32).into();
      let meh2: Value = (&1u8).to_value();
      *meh = meh2;
    
    Fixes https://github.com/gtk-rs/glib/issues/254

Also includes some trivial cleanup.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -416,6 +416,7 @@ impl Drop for ValueArray {
 /// accepted.
 ///
 /// See the [module documentation](index.html) for more details.
+#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
:+1:

EPashkin:
@sdroege  Thanks,
👍 too

GuillaumeGomez:
Thanks!

Issue #253, opened 2017-11-13
GValue can be used to violate Send/Sync traits


sdroege:
E.g. you can store a gtk::Widget in a GValue and then send it to another thread. GValue is (and should be) Send, but it's content might not be.

A solution might be to have two kinds of Value, one Send and one not. And being able to transform from the Send one to the other, similar to TypedValue.

sdroege:
SendValue seems most reasonable, but not sure how to do this most painless. Code generator needs support for this too, so that Send type functions only allow SendValue.

EPashkin:
Agree.

sdroege:
Let's also get this one fixed before the release then :) known safety issues are not fun

sdroege:
Working on it btw, not very focussed right now but adding a trait impl at a time to make things work without problems

PR #256, opened 2017-11-15
Implement SendValue and make Value not implement Send

sdroege:
This is a wrapper around Value that ensures that only Send types are
    stored inside it, and as such also implements Send itself. Value does
    not implement Send anymore.
    
    Fixes https://github.com/gtk-rs/glib/issues/253

What is missing now is to let `gir` emit `SendValue` instead of value if the context it appears in is `Send`

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -94,7 +94,7 @@ use gobject_ffi;
 ///
 /// See the [module documentation](index.html) for more details.
 #[repr(C)]
-pub struct Value(gobject_ffi::GValue);
+pub struct Value(gobject_ffi::GValue, PhantomData<*const c_void>);
 
 impl Value {
     /// Tries to downcast to a `TypedValue`.
@@ -212,7 +212,7 @@ impl<T> From<TypedValue<T>> for Value {
 
 impl Uninitialized for Value {
     unsafe fn uninitialized() -> Value {
-        Value(mem::zeroed())
+        Value(mem::zeroed(), PhantomData)
     }
 }
 
--- a/src/value.rs
+++ b/src/value.rs
@@ -12,6 +12,10 @@
 //! dereferences to `Value` so it can be used everywhere `Value` references are
 //! accepted.
 //!
+//! [`SendValue`](struct.SendValue.html) is a version of [`Value`](struct.Value.html)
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
:+1:

sdroege:
I'm not sure how to best do this in gir, @EPashkin? Basically the `Value` type should be mapped to `SendValue` wherever the signal/object is `Send`. Is there a single place where I could do that? Only option I see right now is to do changes everywhere where signals, trampolines, functions are generated (and then also handle &, &mut, Option<_>, Option<&_>, Option<&mut_> separately).

sdroege:
This shouldn't block merging it btw, it is useful as is already (by making Value non-Send).

sdroege:
I guess one could insert a fake type for `SendValue` into the type library, and in a post-processing step after analysis went through replace every `glib::Value` with that based on the above conditions. Makes sense?

src/value.rs
```diff
@@ -92,9 +96,12 @@ use gobject_ffi;
 /// Some types (e.g. `String` and objects) support `None` values while others
 /// (e.g. numeric types) don't.
 ///
+/// `Value` does not implement the `Send` trait, but [`SendValue`](struct.SendValue.html) can be
+/// used instead.
+///
 /// See the [module documentation](index.html) for more details.
 #[repr(C)]
-pub struct Value(gobject_ffi::GValue);
+pub struct Value(gobject_ffi::GValue, PhantomData<*const c_void>);
```
EPashkin: @sdroege What role does PhantomData?

sdroege: See commit message :) It makes the struct non-Send. Without having anything inside the struct that is not Send, the compiler will implicitely derive Send.

EPashkin: Missed this, Thanks.

EPashkin:
👍

GuillaumeGomez:
Thanks!

src/value.rs
EPashkin: `impl !Sync for T {}` still nightly-only, maybe this need TODO?

sdroege: Ok

Issue #246, opened 2017-11-12
IsA<> trait has to be unsafe


sdroege:
As @alexcrichton noticed yesterday, the `IsA<>` trait must be unsafe. Otherwise you could safely implement it e.g. `impl IsA<gtk::Button> for gtk::Label` or similarly silly things.

Making the `ToGlibPtr` trait unsafe would also solve that, but that one generally seems fine. You can safely convert a `gtk::Button` to a `*mut gtk_ffi::GtkButton`, you just can't dereference that pointer safely (but you never could do that anyway).

sdroege:
@GuillaumeGomez @EPashkin Any opinions?

EPashkin:
Unsafe `IsA<>` seems reasonable.

GuillaumeGomez:
It seems a bit weird for me since the `IsA` trait is implemented only from the `gtk-rs` crates. But apart from this, I don't have any objection.

sdroege:
ToGlibPtr and ToGlibPtrMut have the same problem, they could be used for changing types in incompatible ways by doing a bad, safe impl.

So should also be unsafe. FromGlibPtr should be fine as its functions are all unsafe already

sdroege:
This should be handled before the next release

sdroege:
Actual ToGlibPtr* is fine, you can already safely cast between arbitrary pointer types.

PR #257, opened 2017-11-15
Mark the IsA<_> trait unsafe

sdroege:
Otherwise it is possible, in safe code, to implement IsA<_> between
incompatible types and cause all kinds of bugs.

Fixes https://github.com/gtk-rs/glib/issues/246


EPashkin:
Thanks, @sdroege 
👍

GuillaumeGomez:
Thanks!

PR #264, opened 2017-11-27
Implement Value traits for Vec<String> more consistently

sdroege:
In all other places we treat NULL as an empty array instead of
considering the type to be nullable, do the same here.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -770,12 +770,14 @@ impl SetValueOptional for str {
 
 impl<'a> FromValueOptional<'a> for Vec<String> {
     unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
+        Some(<Vec<String> as FromValue>::from_value(value))
+    }
+}
+
+impl<'a> FromValue<'a> for Vec<String> {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Ah? I thought that we considered NULL as NULL. It can be problematic in certain if someone would want to unset something, no?

sdroege:
See https://github.com/gtk-rs/glib/blob/f597da677bcc58152fa4862b9b0c19dbb4cfb22c/src/translate.rs#L1090

Also `Vec<String>` is generated as that by GIR and never as `Option<Vec<String>>`

sdroege:
Also the translation into a pointer always creates an empty array instead of NULL

GuillaumeGomez:
Hum strange. But yes, in this context, this PR makes sense.

cc @EPashkin

sdroege:
I noticed this because of https://github.com/gtk-rs/gir/pull/496 btw.

EPashkin:
@sdroege Thanks.
@GuillaumeGomez 👍 for this

GuillaumeGomez:
Thanks!

PR #266, opened 2017-11-28
Add Value::from_type() to initialize an empty value from a Type

sdroege:
This is only allowed for types that can be stored in GValues and
otherwise panics. The returned value contains the empty/default value of
the corresponding type.

As suggested by @EPashkin here: https://github.com/gtk-rs/gir/pull/496#issuecomment-347242579
This could be used by GIR for all the property getters to initialize the value

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -627,9 +627,7 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
         };
 
         unsafe {
-            let mut value = Value::uninitialized();
-
-            gobject_ffi::g_value_init(value.to_glib_none_mut().0, property_type.to_glib());
+            let mut value = Value::from_type(property_type);
             gobject_ffi::g_object_get_property(self.to_glib_none().0, property_name.to_glib_none().0, value.to_glib_none_mut().0);
 
             // This can't really happen unless something goes wrong inside GObject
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
:+1:

sdroege:
We might also want to make `Value::uninitialized()` safe, I don't think there's anything unsafe about it or is there? The returned value is just empty and useless, but it can be filled with actual content by other code.

sdroege:
Also we might want to change `Value::is()` to consider subtyping (`Type::is_a()`) instead of just equality, like the `G_VALUE_HOLDS(val, type)` macro in C does.

sdroege:
Done that, but for making `Value::uninitialized` safe we have to do something about the `Uninitialized` trait. Is it used for anything at all other than `Value`?

Which reminded me that the `InstanceType` trait is completely unused currently and can be removed, done

src/object.rs
```diff
@@ -627,9 +627,8 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
         };
 
         unsafe {
-            let mut value = Value::uninitialized();
+            let mut value = Value::from_type(property_type);
```
EPashkin: Empty line

EPashkin:
IMHO `Value::uninitialized()` better be unsafe, as value must not used until it initialized

EPashkin:
@sdroege Thanks

sdroege:
Removed the empty line, mergeable now I guess? Then gir can be updated

GuillaumeGomez:
Yep, just waiting for CIs.

GuillaumeGomez:
Thanks!

PR #269, opened 2017-12-05
Various Path FFI improvements

sdroege:
See commit messages for details.

The first is effectively not changing anything other than adding comments in the code, using expect() instead of unwrap() with more details, and not lossy converting non-UTF8 paths from GLib to Rust on Windows but fail instead on invalid UTF-8 (otherwise we might access the wrong file! potential data loss, etc, and it's a bug in the C library if that ever happens)

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -418,11 +418,20 @@ fn path_to_c(path: &Path) -> CString {
     // representable in UTF-8 and thus can't possibly be passed to GLib.
     // Passing NULL or the empty string to GLib can lead to undefined behaviour, so
     // the only safe option seems to be to simply panic here.
-    CString::new(path.to_str()
+    let path_str = path.to_str()
         .expect("Path can't be represented as UTF-8")
-        .to_owned()
-        .into_bytes()
-    ).expect("Invalid path with NUL bytes")
+        .to_owned();
+
+    // On Windows, paths can have \\?\ prepended for long-path support. See
+    // MSDN documentation about CreateFile
+    //
+    // We have to get rid of this and let GLib take care of all these
+    // weirdnesses later
+    if path_str.starts_with("\\\\?\\") {
+        CString::new(path_str[4..].as_bytes())
+    } else {
+        CString::new(path_str.as_bytes())
+    }.expect("Invalid path with NUL bytes")
 }
 
 impl<'a> ToGlibPtr<'a, *const c_char> for Path {
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -397,16 +397,32 @@ impl GlibPtrDefault for String {
 }
 
 #[cfg(unix)]
-fn path_to_c(path: &Path) -> Option<CString> {
+fn path_to_c(path: &Path) -> CString {
+    // GLib paths on UNIX are always in the local encoding, just like in Rust
+    //
+    // Paths on UNIX must not contain NUL bytes, in which case the conversion
+    // to a CString would fail. The only thing we can do then is to panic, as passing
+    // NULL or the empty string to GLib would cause undefined behaviour.
     use std::os::unix::ffi::OsStrExt;
-    CString::new(path.as_os_str().as_bytes()).ok()
+    CString::new(path.as_os_str().as_bytes())
+        .expect("Invalid path with NUL bytes")
 }
 
 #[cfg(not(unix))]
-fn path_to_c(path: &Path) -> Option<CString> {
-    path.to_str().and_then(|s| {
-        CString::new(s.as_bytes()).ok()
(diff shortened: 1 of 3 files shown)


sdroege:
Please carefully review, @EPashkin and @GuillaumeGomez :)

The second commit is also untested on Windows, if someone can give it some testing... you can get the annoying paths by using Path::canonicalize().

src/translate.rs
```diff
@@ -397,24 +397,49 @@ impl GlibPtrDefault for String {
...
+    let path_str = path.to_str()
+        .expect("Path can't be represented as UTF-8")
+        .to_owned();
+
+    // On Windows, paths can have \\?\ prepended for long-path support. See
+    // MSDN documentation about CreateFile
+    //
+    // We have to get rid of this and let GLib take care of all these
+    // weirdnesses later
+    if path_str.starts_with("\\\\?\\") {
```
GuillaumeGomez: Still makes me laugh so hard.

GuillaumeGomez:
Seems good to me, just a bit sad for all the new panics... Otherwise, huge :+1: for the very complete comments, tt helps a lot for understanding!

src/translate.rs
```diff
@@ -397,24 +397,49 @@ impl GlibPtrDefault for String {
...
+    // Paths on UNIX must not contain NUL bytes, in which case the conversion
+    // to a CString would fail. The only thing we can do then is to panic, as passing
+    // NULL or the empty string to GLib would cause undefined behaviour.
     use std::os::unix::ffi::OsStrExt;
-    CString::new(path.as_os_str().as_bytes()).ok()
+    CString::new(path.as_os_str().as_bytes())
+        .expect("Invalid path with NUL bytes")
 }
 
 #[cfg(not(unix))]
```
EPashkin: Maybe this need changed to `сfg(windows)` and previous to `cfg(not(windows))` ?

GuillaumeGomez: Oh true! And the `#[cfg(unix)]` should be changed to `#[cfg(not(windows))]`.

sdroege: Yeah I was wondering about that. It's more like windows and not-windows instead of unix and not-unix.

GuillaumeGomez: Then we all agree haha.

sdroege: Done

sdroege:
> The second commit is also untested on Windows, if someone can give it some testing... you can get the annoying paths by using Path::canonicalize().

Works fine on Windows

GuillaumeGomez:
Question: should we add tests for these functions? I think it might be a good idea (even more since it's testable!).

sdroege:
> Question: should we add tests for these functions? I think it might be a good idea (even more since it's testable!).

I was wondering about that, but is there a proper way of doing things with files during tests? How/where could we create random files and use them during the test?

GuillaumeGomez:
There are great crates for that like [tempdir](https://crates.io/crates/tempdir) (then you create your files in the temporary directory).

sdroege:
How would I add that as a dependency only for tests? dev-dependencies?

GuillaumeGomez:
`dev-dependencies` indeed. :)

sdroege:
Ok, writing something

sdroege:
There

GuillaumeGomez:
Thanks!

@EPashkin: We don't have a `cargo test` ran in travis nor appveyor?

EPashkin:
Travis don't have :(
https://github.com/gtk-rs/glib/blob/master/appveyor.yml#L19

GuillaumeGomez:
@sdroege Can you add a `cargo test` in `.travis.yml` please?

sdroege:
Sure, done

GuillaumeGomez:
Thanks! Just waiting for CIs now and then I merge (unless you have something @EPashkin?).

EPashkin:
@GuillaumeGomez All looks good and reasonable.
@sdroege Thanks

EPashkin:
There many `GLib-GObject-CRITICAL **: g_value_unset: assertion 'G_IS_VALUE (value)' failed` in travis log

sdroege:
From another test though, but I also can't reproduce it here. All clean here

Can you reproduce?

EPashkin:
I'm not: it on linux.
IMHO better just merge as warning seems not related to this PR and can be fixed later.

sdroege:
I don't get these on Linux either, that's where I'm usually testing.

GuillaumeGomez:
Then I merge. Thanks!

EPashkin:
Errors don't repeated on Ubuntu 16.04 LTS too.

PR #270, opened 2017-12-05
Fix ToGlibContainerFromSlice impls for strings, boxed/object/shared s…

sdroege:
…tructs

It was missing parenthesis, so instead of a NULL pointer at the end
there was only a single NULL byte

This is likely the cause for the critical warnings on travis during the tests @EPashkin @GuillaumeGomez

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -124,7 +124,7 @@ macro_rules! glib_boxed_wrapper {
                 let v: Vec<_> = t.iter().map(|s| s.to_glib_none()).collect();
 
                 let v_ptr = unsafe {
-                    let v_ptr = glib_ffi::g_malloc0(mem::size_of::<*const $ffi_name>() * t.len() + 1) as *mut *const $ffi_name;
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Outch. Not good, thanks a lot for the fix! (so common in C `malloc` calls...)

EPashkin:
Bad that it don't fix errors, but big thanks for finding this.

sdroege:
> so common in C malloc calls...

That's why there's `g_new` and related in GLib :)

sdroege:
I have a similar case in GStreamer btw: https://travis-ci.org/sdroege/gstreamer-rs/jobs/312082352#L800

This one I also can't reproduce locally, it only seems to happen on travis.

sdroege:
Ah, this is it: https://bugzilla.gnome.org/show_bug.cgi?id=755766

Problem is that travis uses an ancient GLib version. But we can work-around that, let me send a PR.

PR #273, opened 2017-12-07
Add bindings for g_file_test() and make use of it in the Path transla…

sdroege:
…tion tests

This ensures that we're not just doing the string transformations
correctly in one way or another, but actually refer to the same/correct
files.


GuillaumeGomez:
:+1:

sdroege:
CI still happy, good

GuillaumeGomez:
Thanks!

PR #274, opened 2017-12-07
Char wrapper 2

EPashkin:
Followup https://github.com/gtk-rs/glib/pull/248
Use https://github.com/gtk-rs/gir/pull/489

cc @GuillaumeGomez, @sdroege


GuillaumeGomez:
:+1:

EPashkin:
@federicomenaquintero Please check this too. I removed `From<Char> for i8`, may be it inappropriate for your usage.

EPashkin:
Added forgoted FromGlib for UChar and test for it

federicomenaquintero:
Thanks for the heads-up.  I haven't worked on gnome-class for a little while, so I need to remind myself what I was doing :)  I'll do that and see what's up with the changes in gir and glib.

src/char.rs
```diff
@@ -23,7 +27,7 @@
 /// done in the `new` function; see its documentation for details.
 ///
 /// The inner `i8` (which is equivalent to `libc::c_char` can be extracted with `.0`, or
-/// by calling `i8::from(my_char)`.
+/// by calling `my_char.to_glib()`.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct Char(pub i8);
```
sdroege: Sorry for bringing up some C bullshit here, but `char` can be signed or unsigned depending on the platform. On PowerPC, `char` is unsigned for example (and gcc has a `-fsigned-chars` parameter for that very reason). And `gchar` in GLib is equivalent to `char`, so it might be signed or unsigned.

sdroege: Of course nothing in the C standard says that chars must be 8 bits, but let's ignore that for now. All platforms where GLib runs have 8 bit chars.

src/char.rs
```diff
@@ -50,30 +54,124 @@ impl Char {
...
+            Some(UChar(c as u8))
+        }
+    }
+}
+
+impl From<UChar> for char {
+    fn from(c: UChar) -> char {
+        c.0 as char
+    }
+}
```
sdroege: Should there be `TryFrom` impls for the other direction? Or rather, with `TryFrom` still being unstable, the same function implemented in `impl Char` / `impl UChar`?

src/char.rs
EPashkin: Then maybe completely replace `i8` with `c_char` ?
Then in PowerPC case we have `Char(u8)` and `UChar(u8)` (if it has this definition in `libc`)

EPashkin: But then maybe problem with testing.

src/char.rs
EPashkin: IMHO `new` will be replaced with `TryFrom` when it stabilized.
Or you meant something other?

src/char.rs
sdroege: Problem with testing what? But yes, that would make sense. It would at least mirror what happens in GLib then

src/char.rs
sdroege: Indeed, nevermind then :)

src/char.rs
EPashkin: Currently in test just used `65i8` and `65u8`, so "PowerPC" and other will need other tests.

EPashkin: @GuillaumeGomez As I remember you prefer `i32` over `c_int` or `gint`.
Can I make reverse here?

EPashkin: Added replacement as `fixup!` commit, I do rebase after accepting this way.

EPashkin:
@GuillaumeGomez and at this too, please

GuillaumeGomez:
An easy one this time, great! Thanks!

PR #276, opened 2017-12-10
Allow creating Bytes from any AsRef<[u8]> by copying or taking ownership of it

sdroege:
See individual commit messages

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -42,7 +42,8 @@ glib_wrapper! {
 
 impl Bytes {
     /// Copies `data` into a new shared slice.
-    fn new(data: &[u8]) -> Bytes {
+    fn new<T: AsRef<[u8]>>(data: T) -> Bytes {
+        let data = data.as_ref();
         unsafe { from_glib_full(glib_ffi::g_bytes_new(data.as_ptr() as *const _, data.len())) }
     }
 
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -53,6 +53,23 @@ impl Bytes {
             from_glib_full(glib_ffi::g_bytes_new_static(data.as_ptr() as *const _, data.len()))
         }
     }
+
+    /// Takes ownership of `data` and creates a new `Bytes` without copying.
+    pub fn from_owned<T: AsRef<[u8]> + Send + 'static>(data: T) -> Bytes {
+        let data: Box<Box<AsRef<[u8]>>> = Box::new(Box::new(data));
+        let (size, data_ptr) = {
+            let data = (**data).as_ref();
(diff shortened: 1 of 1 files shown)


src/bytes.rs
```diff
@@ -42,7 +42,8 @@ glib_wrapper! {
 
 impl Bytes {
     /// Copies `data` into a new shared slice.
-    fn new(data: &[u8]) -> Bytes {
+    fn new<T: AsRef<[u8]>>(data: T) -> Bytes {
```
sdroege: This is also something we should teach GIR about

EPashkin: Not sure that we need transfer ownership in this case

src/bytes.rs
```diff
@@ -52,6 +53,23 @@ impl Bytes {
             from_glib_full(glib_ffi::g_bytes_new_static(data.as_ptr() as *const _, data.len()))
         }
     }
+
+    /// Takes ownership of `data` and creates a new `Bytes` without copying.
+    pub fn from_owned<T: AsRef<[u8]> + Send + 'static>(data: T) -> Bytes {
+        let data: Box<Box<AsRef<[u8]>>> = Box::new(Box::new(data));
```
EPashkin: Maybe we should use `data: Box<Box<T>>` here?

sdroege: Doesn't really make a difference. We only have the box to keep around the data and free it later when needed. If we would make it a specific box instead of a trait object, a new drop function would have to be created by the compiler for each possibly type passed in here.

src/bytes.rs
sdroege: We don't, you can still pass any reference, see the unit test. Only if it was `'static` it would always pass ownership (or require that it's a static reference).

src/bytes.rs
EPashkin: In current case each `from_owned` will not have own copy of `drop_box`?

sdroege: No, there's only a single function that drops a box that contains a trait object. It's not a generic function

EPashkin: Thanks, I thought that it created for any instance of outer generic function

sdroege: Defining a plain function inside another function only makes it invisible from outside the function, it's a way of hiding. Try using any of the outer functions generic types in the inner function, the compiler will complain that it can't do that :)

Same thing with static variables that you define inside a function btw

EPashkin:
Ping and 👍

sdroege:
Ping whom? @GuillaumeGomez ?

GuillaumeGomez:
Yes, me. :)

Thanks!

PR #281, opened 2017-12-28
signal: Make SignalHandlerId derive from Clone

philn:
I think this would be useful :)

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -13,7 +13,7 @@ use source::CallbackGuard;
 use translate::{from_glib, FromGlib, ToGlib, ToGlibPtr};
 
 /// The id of a signal that is returned by `connect`.
-#[derive(Debug, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


EPashkin:
As I remember @sdroege specially don't derive `Clone` to prevent double disconnecting

philn:
Huh... ok... so how would I block a signal handler from 2 different closures?

EPashkin:
Use `Rc<Cell<>>`?

philn:
I already tried with a `RefCell` but it's captured by the first closure.

EPashkin:
Strange, normally, you `clone` RC before moving into closures, so only "half" captured.

philn:
Ah well, I got it working finally with some `Arc<Mutex<>>` trickery :)

EPashkin:
Do not forget to close this PR if problem solved 😉

sdroege:
Yes, the id should be unique so you only can ever remove it once

PR #282, opened 2018-01-19
Regen

GuillaumeGomez:
cc @EPashkin


src/auto/functions.rs
```diff
@@ -91,6 +92,7 @@ pub fn base64_encode(data: &[u8]) -> Option<String> {
 //    unsafe { TODO: call ffi::g_base64_encode_step() }
 //}
 
+#[deprecated]
```
sdroege: Version-less deprecations are less than ideal. The function might've been deprecated in the very latest release, but you can't use the replacement yet because you can't depend on the very latest release (because e.g. Ubuntu LTS is still 10 releases in the past).

EPashkin: This appeared only if it deprecated before `min_cfg_version = "2.32"`.

EPashkin:
👍

src/auto/functions.rs
sdroege: Ok, good then :)

GuillaumeGomez:
I merge then. Thanks for the reviews!

PR #283, opened 2018-01-20
Crate

EPashkin:
cc @GuillaumeGomez


EPashkin:
@GuillaumeGomez tests passed, can you publish this version, so sourceview PR will pass CI.

GuillaumeGomez:
Sure, thanks!

GuillaumeGomez:
Published.

EPashkin:
@GuillaumeGomez Tag it please.

GuillaumeGomez:
Damn, not used to do it by hand. I'll need to update the website as well.

GuillaumeGomez:
Tagged as well.

EPashkin:
Thanks

src/types.rs
```diff
@@ -187,10 +187,6 @@ impl StaticType for Vec<String> {
     }
 }
 
-pub trait InstanceType {
```
sdroege: This breaks API in theory, depends on if you want to ignore that. Nobody probably uses this

GuillaumeGomez: We'll keep this under silence if you don't mind. I want to wait a bit more for a bigger release. ;)

sdroege: Fine with me, but it should be a conscious decision and not an accident

GuillaumeGomez: I have to admit it completely slipped out of my review. Should have been more careful...

Issue #284, opened 2018-01-20
0.4.1 is broken on all arches except x86/x86_64


ignatenkobrain:
```rust
error[E0600]: cannot apply unary operator `-` to type `u8`
   --> src/char.rs:143:54
    |
143 |         assert_eq!(Char::new(255 as char), Some(Char(-1)));
    |                                                      ^^
error[E0308]: mismatched types
   --> src/char.rs:144:46
    |
144 |         assert_eq!(Char::new('ñ'), Some(Char(241 as u8 as i8)));
    |                                              ^^^^^^^^^^^^^^^ expected u8, found i8
error[E0308]: mismatched types
   --> src/char.rs:158:9
    |
158 |         assert_eq!(Char::new('A').unwrap().to_glib(), 65i8);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u8, found i8
    |
    = note: this error originates in a macro outside of the current crate
error[E0308]: mismatched types
   --> src/char.rs:168:36
    |
168 |         assert_eq!(char::from(Char(65i8)), 'A');
    |                                    ^^^^ expected u8, found i8
error[E0308]: mismatched types
   --> src/char.rs:174:25
    |
174 |         assert_eq!(Char(65i8), from_glib(65i8));
    |                         ^^^^ expected u8, found i8
error[E0277]: the trait bound `char::Char: translate::FromGlib<i8>` is not satisfied
   --> src/char.rs:174:32
    |
174 |         assert_eq!(Char(65i8), from_glib(65i8));
    |                                ^^^^^^^^^ the trait `translate::FromGlib<i8>` is not implemented for `char::Char`
    |
    = help: the following implementations were found:
              <char::Char as translate::FromGlib<u8>>
    = note: required by `translate::from_glib`
```

ignatenkobrain:
0.4.0 was totally fine.

GuillaumeGomez:
Erf, we'll need to add tests for other architectures as well then...

cc @EPashkin @sdroege

sdroege:
That's the problem of c_char being u8 on some archs and i8 elsewhere? I think I mentioned that in the original PR for this. You need to cast everything to i8 if that's the behavior you want

sdroege:
Preparing a PR for fixing this now

GuillaumeGomez:
Thanks @sdroege!

PR #285, opened 2018-01-21
Fix Char tests on platforms where c_char is u8 instead of i8

sdroege:
Also add correct casts to the UChar tests while we're at it.

Fixes https://github.com/gtk-rs/glib/issues/284

### DIFF
--- a/src/char.rs
+++ b/src/char.rs
@@ -139,12 +139,12 @@ mod tests {
 
     #[test]
     fn converts_single_byte_chars() {
-        assert_eq!(Char::new(0 as char), Some(Char(0)));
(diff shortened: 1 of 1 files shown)


EPashkin:
@sdroege Thanks

ignatenkobrain:
Let me try this on our altarches

ignatenkobrain:
This works just fine ;)

sdroege:
What architecture is that btw? PPC?

ignatenkobrain:
ppc64, ppc64le, s390x, aarch64, armv7hl

sdroege:
I didn't know armv7hl and aarch64 also had unsigned chars by default

GuillaumeGomez:
Thanks!

PR #286, opened 2018-01-23
Unignore markup_escape_text. Fixes #280.

### LINKED ISSUE #280: markup_escape_text
Hi! I can't find a `markup_escape_text` function. In your Gir.toml it says "ignored = true", so I guess it's disabled. Is there any alternative, or any plans to enable it?

tmiasko:


GuillaumeGomez:
*Seems* good to me. Strange we didn't generate it earlier.

cc @EPashkin @sdroege

src/auto/functions.rs
```diff
@@ -786,6 +786,13 @@ pub fn main_depth() -> i32 {
 //    unsafe { TODO: call ffi::g_markup_collect_attributes() }
 //}
 
+pub fn markup_escape_text(text: &str) -> Option<String> {
```
sdroege: This can't return NULL AFAIU

Maybe some of the other markup related functions can also be enabled now?

tmiasko: I presume that following should get rid of Option:
```
[object.function.return]
nullable = false
```
In this case this doesn't work correctly right now.

EPashkin: It works correctly for me
```toml
    [[object.function]]
    name = "markup_escape_text"
    [object.function.return]
    nullable = false
```

GuillaumeGomez: @EPashkin: Please don't break indent! :scream:

EPashkin:
@tmiasko can you also remove "#gsize" comment for "convert" and other 3 next? Its just not usable in rust?

src/auto/functions.rs
EPashkin: Oh yes, I will fix comment

tmiasko:
I made return type non nullable (it seems I had put two brackets around
object.function.return so it didn't work earlier). 

@EPashkin my impression is that default codegen doesn't do the right thing
for those above. I don't want to spend time on them as I don't need them so far.

EPashkin:
I want remove comment as it meant: "This function ignored due gir works bad with gsize type" and that wrong for this 4 function and its ignored for different reason.

But if you totally don't want do this, no problem, you have 👍 for merge as is.

GuillaumeGomez:
Then once CI is done, I merge.

EPashkin:
CI passed

PR #287, opened 2018-01-23
 Simplify KeyFile::to_data 

EPashkin:
It always return string https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html#g-key-file-to-data. Also it return normall null-terminated string so `length` unneeded.

cc @GuillaumeGomez, @sdroege 

Based on #286

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -176,19 +176,19 @@ status = "generate"
     ignore = true
     [[object.function]]
     name = "convert"
-    #gsize
+    #unusable
     ignore = true
     [[object.function]]
     name = "filename_to_utf8"
-    #gsize
+    #unusable
     ignore = true
     [[object.function]]
     name = "locale_from_utf8"
-    #gsize
+    #unusable
     ignore = true
     [[object.function]]
     name = "locale_to_utf8"
-    #gsize
+    #unusable
     ignore = true
     [[object.function]]
     name = "markup_escape_text"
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
I suppose you can remove the first commit now.

EPashkin:
It will not present in merge.

GuillaumeGomez:
Ok, now all good!

GuillaumeGomez:
Thanks!

Gir.toml
```diff
@@ -176,19 +176,19 @@ status = "generate"
...
     ignore = true
     [[object.function]]
     name = "filename_to_utf8"
-    #gsize
+    #unusable
     ignore = true
     [[object.function]]
     name = "locale_from_utf8"
-    #gsize
+    #unusable
```
sdroege: Unusable seems a bit harsh. These can all be manually bound if we want

src/key_file.rs
```diff
@@ -63,4 +63,11 @@ impl KeyFile {
             }
         }
     }
+
+    pub fn to_data(&self) -> String {
+        unsafe {
+            let ret = ffi::g_key_file_to_data(self.to_glib_none().0, ptr::null_mut(), ptr::null_mut());
```
sdroege: This can never fail?! Why is there a GError then :)

sdroege: Yes it can't fail, the GError is never ever used in the code. How weird!

PR #288, opened 2018-01-25
Remove Default impl for SourceId

sdroege:
While 0 is never a valid SourceId, it is also not valid to use it with
any functions that work on SourceIds. It causes assertions at runtime.
The pattern of having a "non-existing" SourceId is better suited by
using an Option, and once NonZero is stabilized this can even be done
without introducing any memory overhead.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -28,6 +28,7 @@ impl ToGlib for SignalHandlerId {
 impl FromGlib<c_ulong> for SignalHandlerId {
     #[inline]
     fn from_glib(val: c_ulong) -> SignalHandlerId {
+        assert_ne!(val, 0);
         SignalHandlerId(val)
     }
 }
--- a/src/source.rs
+++ b/src/source.rs
@@ -29,6 +29,7 @@ impl ToGlib for SourceId {
 impl FromGlib<u32> for SourceId {
     #[inline]
     fn from_glib(val: u32) -> SourceId {
+        assert_ne!(val, 0);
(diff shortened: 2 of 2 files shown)


EPashkin:
Not fully understand your reason and solution, but I not against this change.
Maybe also make inner `u32` private, the add assert !=0 to `impl FromGlib<u32> for SourceId` ?

GuillaumeGomez:
The reasoning makes sense for me so 👍

sdroege:
>  Maybe also make inner u32 private, the add assert !=0 to impl FromGlib<u32> for SourceId ?

It already is private fortunately :) We could add an assertion in the `FromGlib` impl, but don't do so for `SignalHandlerId` and similar things either. Should we? If we ever get 0 from C code, things are probably very broken

EPashkin:
There will problem with `SignalHandlerId` and https://developer.gnome.org/gobject/stable/gobject-Signals.html#g-signal-connect-data: as I understand it can return 0 on wrong input

sdroege:
`g_signal_connect` returns 0 if you provide an invalid object, or a signal name that does not exist on the object. In the bindings this can't really happen, in Object::connect() this can happen and would return `Err` already (I hope, I'll check).

Ok, so I check the above and add assertions for these cases to the `FromGlib` impls.

sdroege:
Looks all fine

EPashkin:
👍

GuillaumeGomez:
Thanks!

PR #290, opened 2018-02-07
Add GIOCondition

sdroege:


sdroege:
@EPashkin @tmiasko GIR broke :) It includes the wrong imports when regenerating for glib. Must be `use value::XXX`.

EPashkin:
@sdroege I merged fix. Bug was added in https://github.com/gtk-rs/gir/commit/6472453538fb15cfa67a6c39d88061068833cf88 ;)

sdroege:
And updated, should be fine now :) Thanks!

src/auto/functions.rs
```diff
@@ -643,15 +643,15 @@ pub fn intern_string<'a, P: Into<Option<&'a str>>>(string: P) -> Option<String>
...
 //    unsafe { TODO: call ffi::g_io_add_watch() }
 //}
 
-//pub fn io_add_watch_full<P: Into<Option</*Unimplemented*/Fundamental: Pointer>>>(channel: /*Ignored*/&IOChannel, priority: i32, condition: /*Ignored*/IOCondition, func: /*Unknown conversion*//*Unimplemented*/IOFunc, user_data: P, notify: /*Unknown conversion*//*Unimplemented*/DestroyNotify) -> u32 {
+//pub fn io_add_watch_full<P: Into<Option</*Unimplemented*/Fundamental: Pointer>>>(channel: /*Ignored*/&IOChannel, priority: i32, condition: IOCondition, func: /*Unknown conversion*//*Unimplemented*/IOFunc, user_data: P, notify: /*Unknown conversion*//*Unimplemented*/DestroyNotify) -> u32 {
 //    unsafe { TODO: call ffi::g_io_add_watch_full() }
 //}
 
-//pub fn io_create_watch(channel: /*Ignored*/&IOChannel, condition: /*Ignored*/IOCondition) -> Option<Source> {
+//pub fn io_create_watch(channel: /*Ignored*/&IOChannel, condition: IOCondition) -> Option<Source> {
```
GuillaumeGomez: Still not ready for IOChannel?

GuillaumeGomez:
Seems all good for me!

EPashkin:
@sdroege Please, update gir submodule too

src/auto/functions.rs
sdroege: Maybe, but even if we were this would have to be manual so that the callback, etc can be set on the Source. See glib/src/source.rs for some other examples.

sdroege:
> @sdroege Please, update gir submodule too

Done :)

sdroege:
@GuillaumeGomez Should be good to go then :)

GuillaumeGomez:
We generally update git submodules in a commit on its own. Please try to do it like this next time. 😉 

Otherwise thanks!

PR #294, opened 2018-02-20
Bump version to match `crate` branch

talklittle:
To avoid problems with multiple `glib` versions (0.4.0 and 0.4.1) being resolved by Cargo, when working with git versions of gtk-rs crates.


EPashkin:
@talklittle Thanks
👍

GuillaumeGomez:
Sorry but versions are updated when we make a new release. Changing the version number now would make the new release problematic.

talklittle:
@GuillaumeGomez I think it is an oversight that the `master` branch and `crate` branch have different versions though. `crate` is currently 0.4.1 while `master` is 0.4.0, and that seems wrong to me. Shouldn't they be in sync?

sdroege:
Actually master should be versioned 0.5.0 probably, it's API incompatible with the version on crates.io and it's what is going to become 0.5.0.

talklittle:
@sdroege Why is the `glib` crate different than all the other gtk-rs crates in this regard?

If I drop a `[patch.crates-io]` into my application's Cargo.toml, I can override other gtk-rs crates (gtk, gdk, gtk-sys, etc.) since the git versions share the same version numbers as the respective crates, and can be pinned to a revision that is equivalent to the crate on crates.io.

However, currently `glib` is 0.4.0 on master branch even though a [comparison with the crate branch on version 0.4.1](https://github.com/gtk-rs/glib/compare/crate) reveals they are essentially equivalent. Since they have different versions, however, Cargo 1.24 does not properly override the `glib` dependency; it tries to import 2 different `glib`s due to the 2 different version numbers provided.

Again, `glib` is the only gtk-rs crate that currently behaves this way, and I think it's because the version number was accidentally bumped only in the `crate` branch and not in the `master` branch at the same time, in the merge commit 31bdffb91f61b954b4b043f8016dba5817635ee5

sdroege:
> @sdroege Why is the glib crate different than all the other gtk-rs crates in this regard?

It's not, they're all the same, I believe. But for why the branch setup is like it is you'll have to ask @GuillaumeGomez . I don't think it's correct or easy how it is, I'm doing it different for my crates (master is always using GIT versions of all the gtk-rs crates and is the latest development version, 0.X branch is a bugfix branch of the 0.X releases and that's where the releases from crates.io are made from every now and then).

sdroege:
But master for the gtk-rs repositories is always the latest development version. If it is mostly equivalent to the last release in case of glib that's just coincidence.

talklittle:
@sdroege Thanks very much for the replies. I'll concede the point after having come to a better understanding of the situation now. Your point that the `crate` branch is not really a "snapshot" of `master`, but rather treated like a separate branch, e.g. `0.4` branch including bugfixes, helped me to understand the problem better. So this PR is a band-aid solution that would fall apart again in the future.

sdroege:
I think without changing anything else, master should be updated to 0.5.0 (and correspondingly for the other repositories).

GuillaumeGomez:
Oh indeed, my bad! Sorry for the misunderstanding. Weird that `master` wasn't updated...

sdroege:
@GuillaumeGomez 0.5.0 would be more correct though, this is only ~ the same as the 0.4.1 release by coincidence. Similarly all the other master branches should be updated to a higher minor version.

GuillaumeGomez:
@sdroege: The crate version update only occurs when I make a new release. But yes, next release will be `0.5.0` for glib. In this case, it was a mistake that the `master` branch didn't have the same version number that the `©rates.io` one.

sdroege:
The master branch should not have the same version as what is on crates.io. For all the other crates at least the API is different, so the minor version should be higher than what is on crates.io.

GuillaumeGomez:
We can't guess at the time we make a release if the next one will be a minor or not and if there will have minor fixes so it seems logical to me that this version update is taking place during the release.

sdroege:
If there are API changes in master (like for all the other crates), we know that there will be a major release for the next time. Also that's the reason why a bugfix branch for doing bugfix releases is useful to have

GuillaumeGomez:
I think you missed my point. At the moment of the release, we don't know if there will be API changes for the next one.

sdroege:
Sure, of course not. But the Cargo.toml of e.g. gtk should now say 0.4.0 (instead of 0.3.0), as current master is API-different from 0.3.0 (which is on crates.io).

GuillaumeGomez:
I think it's debatable. I never paid attention to crate's version when I pull them from a git repository and I'm surprised some do since it doesn't have much meaning. I don't like the idea to have a bugfix branch in order to match the `crates.io` version either. :-/

PR #302, opened 2018-03-27
Implement Object::new() around g_object_newv() for instantiating obje…

sdroege:
…cts by type id

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -583,6 +583,42 @@ glib_object_wrapper! {
     Object, GObject, GObjectClass, @get_type gobject_ffi::g_object_get_type()
 }
 
+impl Object {
+    pub fn new(type_: Type, properties: &[(&str, &ToValue)]) -> Result<Object, BoolError> {
+        use std::ffi::CString;
+
+        if !type_.is_a(&Object::static_type()) {
+            return Err(BoolError("Can't instantiate non-GObject objects"));
+        }
+
+        let params = properties.iter()
+                               .map(|&(name, value)|
+                                    (CString::new(name).unwrap(), value.to_value()))
+                               .collect::<Vec<_>>();
+
+        let params_c = params.iter()
+                             .map(|&(ref name, ref value)|
+                                  gobject_ffi::GParameter {
+                                      name: name.as_ptr(),
+                                      value: unsafe { *value.to_glib_none().0 }
+                                  })
+                             .collect::<Vec<_>>();
+
+        unsafe {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Seems good to me!

src/object.rs
```diff
@@ -583,6 +583,42 @@ glib_object_wrapper! {
     Object, GObject, GObjectClass, @get_type gobject_ffi::g_object_get_type()
 }
 
+impl Object {
+    pub fn new(type_: Type, properties: &[(&str, &ToValue)]) -> Result<Object, BoolError> {
+        use std::ffi::CString;
+
+        if !type_.is_a(&Object::static_type()) {
+            return Err(BoolError("Can't instantiate non-GObject objects"));
```
sdroege: In theory this could be an assertion, it's a definite programming error

src/object.rs
```diff
@@ -583,6 +583,42 @@ glib_object_wrapper! {
...
+        }
+
+        let params = properties.iter()
+                               .map(|&(name, value)|
+                                    (CString::new(name).unwrap(), value.to_value()))
+                               .collect::<Vec<_>>();
```
sdroege: We could also check for all properties if they actually exist and are of the given value, and assert otherwise or return an error.

src/object.rs
```diff
@@ -583,6 +583,42 @@ glib_object_wrapper! {
...
+                                      value: unsafe { *value.to_glib_none().0 }
+                                  })
+                             .collect::<Vec<_>>();
+
+        unsafe {
+            let ptr = gobject_ffi::g_object_newv(type_.to_glib(), params_c.len() as u32, mut_override(params_c.as_ptr()));
+            if ptr.is_null() {
+                Err(BoolError("Can't instantiate object"))
+            } else {
+                if type_.is_a(&from_glib(gobject_ffi::g_initially_unowned_get_type())) {
```
EPashkin: Maybe better `} else if ..`

sdroege: Why? That whole branch is the non-NULL case :)

EPashkin:
@sdroege Thanks, +1
Only small nit, that can be ignored.

src/object.rs
EPashkin: For me it looks as tree-way choice, and all 3 cases is equivalent

sdroege:
Ok, let's get it in then?

GuillaumeGomez:
Let's go then!

PR #303, opened 2018-03-29
Use ptr::NonNull for Boxed/Shared wrappers

sdroege:
This allows the compiler to optimize better, and in our case these
pointers are guaranteed to be not NULL.

Requires Rust 1.25. We might want to do the same for Stash maybe, or even the translation traits. But that's far more intrusive.

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -8,6 +8,7 @@ use std::ops::{Deref, DerefMut};
 use std::fmt;
 use std::marker::PhantomData;
 use std::mem;
+use std::ptr;
 use translate::*;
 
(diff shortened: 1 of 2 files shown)


EPashkin:
Very good, but IMHO better wait for stable 1.26 or even 1.27

sdroege:
You mean to prevent another annoyance like with the callback guards?

EPashkin:
Yes

sdroege:
Ok, not that important in any case :) I updated GStreamer to use it already though

GuillaumeGomez:
Then I'll merge it as well (you should have remind me this one! ><).

Issue #307, opened 2018-04-07
Object set_property doesn't work with interfaces / derived types.


tmiasko:
Object set_property currently validates that value has exactly the same type as
property, which means you can't use it with derived types / types that
implement given interface.

sdroege:
Indeed, want to send a PR? Probably the same issue with the GValue getter/setter for objects

sdroege:
You could use `Type::is_a` here. Transform is not ideal as there are transform functions from/to strings, for example. IMHO all this should only consider subtyping

The checks could be removed but I'd prefer a proper error on the Rust side over an "unobservable" critical in GLib

tmiasko:
But my point is, that `g_object_set_property()` already uses those transforms, and we are just making this invalid.

sdroege:
So you can pass `1` to a boolean property and it becomes `TRUE`? Basically any fundamental type could be set on a string property. And even from integers to enums/flags. I did not know that but according to `gobject.c` you're right. How ugly :)

So a correct check would be `g_value_type_transformable()` and  `g_param_value_validate()` (which unfortunately already needs the transformed value... plus the flags check). But we might want to be more restrictive here, e.g. passing arbitrary integers to enum/flags properties kind of defeats the purpose of our nice enum/flags types.

My feeling would be, like I said above, that we should allow subtyping but no other kinds of type mismatches. For the same reason we use proper enums/flags, and Rust does not automatically cast between different integer types, etc.. We might want to move `g_param_value_validate()` into the Rust wrapper though to also do range checking.

And as I mentioned above, I assume the same problem with regards to subtyping exists in the GValue getter/setter impls.

sdroege:
The GValue setters/getters already take subtyping into account, and only that (no transforms)

tmiasko:
This was fixed in #311.

PR #308, opened 2018-04-08
Replicate g_object_set_property conversion and validation in Rust.

tmiasko:
Extend the set of values accepted by set_property, which previously
required that value type and property type must match exactly, by
requiring only that it must be possible to transform value into the
property type.

Extend the set of errors that can be detected on the Rust side by using
g_param_value_validate to ensure that provided value is valid.

Avoid duplicating conversion work by passing already converted value to
g_object_set_property.

Fixes issue #307.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -650,14 +650,40 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
 
     fn set_property<'a, N: Into<&'a str>>(&self, property_name: N, value: &ToValue) -> Result<(), BoolError> {
         let property_name = property_name.into();
+        let property_value = value.to_value();
 
-        if let Err(error) = self.has_property(property_name, Some(value.to_value_type())) {
-            return Err(error);
-        }
-
-        let value = value.to_value();
         unsafe {
-            gobject_ffi::g_object_set_property(self.to_glib_none().0, property_name.to_glib_none().0, value.to_glib_none().0)
+            let object = self.to_glib_none().0;
+            let klass = (*object).g_type_instance.g_class as *mut gobject_ffi::GObjectClass;
+            let pspec = gobject_ffi::g_object_class_find_property(klass, property_name.to_glib_none().0);
+            if pspec.is_null() {
+                return Err(BoolError("property not found"));
+            }
+
+            let writeable = (*pspec).flags & gobject_ffi::G_PARAM_WRITABLE != 0;
+            if !writeable {
+                return Err(BoolError("property is not writeable"));
+            }
+
+            let mut transformed_value = Value::uninitialized();
+            gobject_ffi::g_value_init(transformed_value.to_glib_none_mut().0, (*pspec).value_type);
+            let transformed: bool = from_glib(gobject_ffi::g_value_transform(
+                    property_value.to_glib_none().0,
+                    transformed_value.to_glib_none_mut().0));
+            if !transformed {
+                return Err(BoolError("property can't be set from given type"));
+            }
+
+            let changed: bool = from_glib(gobject_ffi::g_param_value_validate(
+                    pspec, transformed_value.to_glib_none_mut().0));
+            let change_allowed = (*pspec).flags & gobject_ffi::G_PARAM_LAX_VALIDATION != 0;
+            if changed && !change_allowed {
+                return Err(BoolError("property can't be set from given value, it is invalid or out of range"));
+            }
+
+            gobject_ffi::g_object_set_property(self.to_glib_none().0,
+                                               property_name.to_glib_none().0,
+                                               transformed_value.to_glib_none().0);
         }
 
         Ok(())


sdroege:
I disagree with this one as it allows setting integers on bool properties, for example. We shouldn't throw type safety completely away :)

IMHO this should allow exact type and subtypes only

tmiasko:
> We shouldn't throw type safety completely away :)

I don't think we are. GLib intentionally supports some of them and not others:

```c
  /* some transformations are a bit questionable,
   * we currently skip those
   */
#define SKIP____register_transform_func(type1,type2,transform_func)     /* skip questionable transforms */ \
...
  SKIP____register_transform_func (G_TYPE_BOOLEAN,      G_TYPE_FLOAT,           value_transform_int_float);
  SKIP____register_transform_func (G_TYPE_BOOLEAN,      G_TYPE_DOUBLE,          value_transform_int_double);
  SKIP____register_transform_func (G_TYPE_ENUM,         G_TYPE_BOOLEAN,         value_transform_long_bool);
  SKIP____register_transform_func (G_TYPE_ENUM,         G_TYPE_FLOAT,           value_transform_long_float);
  SKIP____register_transform_func (G_TYPE_ENUM,         G_TYPE_DOUBLE,          value_transform_long_double);
  SKIP____register_transform_func (G_TYPE_FLAGS,        G_TYPE_BOOLEAN,         value_transform_ulong_bool);
  SKIP____register_transform_func (G_TYPE_FLAGS,        G_TYPE_ENUM,            value_transform_ulong_long);
  // ... those above are subset of those skipped, see source code details ..
```

If you are interested in providing additional type safety, I think that should
be done by generation of custom setters where they aren't already provided in C
API.

Imposing some arbitrary limitation on `g_object_set_property`, could hinder its
use in more generic context. Consider how much trouble property with a long
type could cause (though, it seems more like black swan kind of thing).

sdroege:
> I don't think we are. GLib intentionally supports some of them and not others:

We allow converting any integer into an enum/flags or bool for example.

What's your use-case for this change btw?

> If you are interested in providing additional type safety, I think that should
be done by generation of custom setters where they aren't already provided in C
API.

That won't help in case of GStreamer as properties are provided by plugins often, without any bindings and no API other than the normal GStreamer and GObject API

> Imposing some arbitrary limitation on g_object_set_property, could hinder its
use in more generic context. Consider how much trouble a property with the long
type could cause (though, it seems more like black swan kind of thing).

Using `long` in APIs is a bad idea, yes :) We don't even have GValue trait impls for that one, but if we had they should probably be around `i64` on the Rust side. And you'd need to provide the type information explicitly in any case as `long` either overlaps with `i32` or `i64`.

tmiasko:
> We allow converting any integer into an enum/flags or bool for example.

My first impression after reading GLib source was that conversion from integer
to bool would be allowed but only for integers with values 0 and 1. Anything
that changes the value in the process would be caught after the check following
`g_param_value_validate`, unless `G_PARAM_LAX_VALIDATION` flag is set (which
seems to be very uncommon).

Alas, now I see this is incorrect, because transforms already put the value
into correct range, so the check performed based on results from
`g_param_value_validate` would be never triggered (this looks like a bug in
GLib).

I think that completely strict checking would cause more trouble than it is
worth, especially for integer constants. I would prefer check that values are
not changed in the process, but I have very little interest in actually writing
it. So current version is what I am proposing, given that it solves the real
problem with settings interfaces.

sdroege:
> I think that completely strict checking would cause more trouble than it is
worth, especially for integer constants.

Can you give a concrete example?

src/object.rs
```diff
@@ -650,14 +650,40 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+            let object = self.to_glib_none().0;
+            let klass = (*object).g_type_instance.g_class as *mut gobject_ffi::GObjectClass;
+            let pspec = gobject_ffi::g_object_class_find_property(klass, property_name.to_glib_none().0);
+            if pspec.is_null() {
+                return Err(BoolError("property not found"));
+            }
+
+            let writeable = (*pspec).flags & gobject_ffi::G_PARAM_WRITABLE != 0;
+            if !writeable {
+                return Err(BoolError("property is not writeable"));
```
sdroege: Could also check for readability in `get_property` I guess

src/object.rs
```diff
@@ -650,14 +650,40 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+            }
+
+            let mut transformed_value = Value::uninitialized();
+            gobject_ffi::g_value_init(transformed_value.to_glib_none_mut().0, (*pspec).value_type);
+            let transformed: bool = from_glib(gobject_ffi::g_value_transform(
```
sdroege: My proposal would be to simply replace these lines here with
```rust
if !gobject_ffi::g_type_check_value_holds(mut_override(property_value.to_glib_none().0),  (*pspec).value_type) { return Err(...); }
```

and then passing `property_value` onwards. That should cover your use-case of interface/parent classes and not mix in the dubious integer transforms.

tmiasko:
> > I think that completely strict checking would cause more trouble than it is worth, especially for integer constants.
> 
> Can you give a concrete example?

```rust
// Which one is it:
o.set_property("name", &10);
o.set_property("name", &10u64);
o.set_property("name", &10u32);
o.set_property("name", &10i64);
o.set_property("name", &10i32);
```

sdroege:
> // Which one is it:

The one that is it according to the documentation / GObject property of the C GObject. If it's a UINT, it is a u32, for example.

sdroege:
@GuillaumeGomez @EPashkin Any opinions? Open question is basically if we should only allow silently converting between any integer/bool/flag/enum types and subclassing to the property type (like GLib does itself unfortunately), or only allow the exact type and subclassing.

Argument for the first is that this is what GLib does, argument for the second is that Rust also does not convert automatically between arbitrary integer types and bools, or even worse integers and enums/flags (the latter allowing to set values outside the enum flag/range).

GuillaumeGomez:
I prefer explicit over implicit in Rust since it's the language preference. So if possible, I'd prefer not convert silently.

sdroege:
OK, then I'll send a PR on top of this one later  that does exactly that as @tmiasko said he doesn't want to do that. Allowing only subclassing and otherwise exactly the same type, plus additional tests for the property better

I'll also add a test or two for that, if I can find API in gio that can be used for that.

tmiasko:
Let me make it clear, if it wasn't in https://github.com/gtk-rs/glib/pull/308#issuecomment-379757902, I don't like what GLib does either.
What @sdroege is proposing is fine with me, just not my favorite solution.

sdroege:
@tmiasko Understood :) Do you want to update your PR (plus the additional checks in the getter), or do you want me to do that? I'm fine either way

sdroege:
See https://github.com/gtk-rs/glib/pull/311, this extends on top of this PR

sdroege:
This one can be closed then

GuillaumeGomez:
Yes sir!

PR #309, opened 2018-04-08
Implement pushing/popping a thread default main context in a panic-sa…

sdroege:
…fe way

Not sure about the name of the struct. And maybe we need to somehow prevent it from implementing `Send`.

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -329,6 +329,16 @@ concurrency = "send+sync"
        # source id is a newtype
        ignore = true
 
+       [[object.function]]
+       name = "push_thread_default"
+       # panic-safe popping
+       ignore = true
+
+       [[object.function]]
+       name = "pop_thread_default"
+       # panic-safe popping
+       ignore = true
+
 [[object]]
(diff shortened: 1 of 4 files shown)


tmiasko:
What exactly could panic here? 

This approach still allows one to cross the lifetimes of different contexts.
Using a method that accepts a closure would eliminate the issue (as long one
does not call push / pop, manually). Maybe something alone the lines of:

```rust
ctx.with_thread_default(|| {
  ...
});

with_thread_default_context(ctx, || {
  ...  
});

thread_default(ctx, || {
  ...  
});
```

sdroege:
> What exactly could panic here?

If something between push and pop panics, pop would never happen

You're right though, it's possible to mess up the order here still and it would ideally have to be done with a closure like you suggest. Thanks!

sdroege:
The question then is whether we want that or if it's too annoying to use. Opinions?

sdroege:
Closing in favour of https://github.com/gtk-rs/glib/pull/319

PR #310, opened 2018-04-09
Implement Sync and Send traits for all weak references.

tmiasko:
Provide a wrapper around GWeakRef that can be safely transferred and
shared with other threads. Weak references containing types that are
Sync can be safely upgraded across different threads. In all other cases
this could potentially lead to thread safety issues. Introduce a dynamic
check that validates that upgrade operation is performed on the correct
thread.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -9,10 +9,15 @@ use types::{self, StaticType};
 use wrapper::{UnsafeFrom, Wrapper};
 use ffi as glib_ffi;
 use gobject_ffi;
-use std::mem;
-use std::ptr;
+use std::error;
+use std::fmt;
 use std::iter;
 use std::marker::PhantomData;
+use std::mem;
+use std::ptr;
+use std::result;
+use std::sync::Arc;
+use std::thread;
 
 use Value;
 use value::{ToValue, SetValue};
@@ -636,6 +641,15 @@ pub trait ObjectExt: IsA<Object> {
     fn emit<'a, N: Into<&'a str>>(&self, signal_name: N, args: &[&ToValue]) -> Result<Option<Value>, BoolError>;
     fn disconnect(&self, handler_id: SignalHandlerId);
 
+    /// Creates a new weak reference to this object.
+    ///
+    /// Returned reference can be upgraded back into strong reference only from the same thread.
+    /// Attempts to do so from different threads will be unsuccessful.
+    ///
+    /// See also [`downgrade_sync`][downgrade_sync] which allows upgrading references from
+    /// different threads.
+    ///
+    /// [downgrade_sync]: trait.SyncObjectExt.html#method.downgrade_sync
     fn downgrade(&self) -> WeakRef<Self>;
 }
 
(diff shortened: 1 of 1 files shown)


sdroege:
Like SendCell but as weak reference? How would you ensure in your example that the object is not destroyed in the meantime, who would hold the reference? And why not use SendCell instead?

This seems to basically duplicate it and diverges from the GObject API. It should be equivalent to use a `SendCell<WeakRef<T>>`

Maybe WeakRef should only be Send/Sync if the contained type is?

tmiasko:
> How would you ensure in your example that the object is not destroyed in the meantime, who would hold the reference?

The point of using weak references is that the object can be destroyed in the
meantime. In this case upgrade method just returns None, in the same way other
weak pointers in Rust do. 

> And why not use SendCell instead?

Because SendCell has to be either unsafe or leak memory?

> This seems to basically duplicate it and diverges from the GObject API. 

It has two additional features: upgrades are thread-safe, and weak pointer can
be safely cloned. It also retains compatibility with GObject API in principle.
I didn't implement usual conversion routines because weak pointers aren't
generally used in public API, so I don't have any need for them.

> Maybe WeakRef should only be Send/Sync if the contained type is?

IMHO the primary uses case are things that aren't Sync, like GtkWidgets where
you need to prepare data before updating widget. If widget is destroyed in the
meantime that is perfectly fine, you just don't update it. Not to mention that
there are only a handful of things that are Sync and are Objects, and for them
you may not need weak references in the first place.

sdroege:
Sorry if my comment seemed harsh, I wrote it in a hurry on the phone.

Your changes look useful, and the thing I missed why `SendCell` can't be used is that `SendCell<WeakRef<T>>` would leak/be unsafe because it would potentially be destroyed on another thread, but it actually is fine to destroy the `WeakRef` from another thread. Correct?

src/object.rs
```diff
@@ -866,53 +880,112 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
     }
 }
 
-impl<T: IsA<Object> + ?Sized> Clone for WeakRef<T> {
-    fn clone(&self) -> Self {
-        unsafe {
-            let c = WeakRef(Box::new(mem::uninitialized()), PhantomData);
+/// An error returned by [`WeakRef::try_upgrade`](struct.WeakRef.html#method.try_upgrade).
+#[derive(Debug)]
+pub struct UpgradeError {
```
sdroege: We usually just use BoolError for such things here

tmiasko:
Yes, WeakRef can be safely destroyed in any thread, because it doesn't own the object referenced.

sdroege:
> Yes, WeakRef can be safely destroyed in any thread, because it doesn't own the object referenced.

Yeah, agreed. And because `WeakRef` handling in GObject actually is thread-safe.

src/object.rs
tmiasko: I was thinking about adding additional debug information like thread IDs, that
is why I used separate error type and left a room to change implementation
details without breaking API. Though, neither GLib thread IDs, nor Rust IDs
correspond to native threads IDs you could easily see in debugger, which I
think would be most useful. Thus for now I left error message without
additional details.

sdroege:
@GuillaumeGomez @EPashkin Any opinions? IMHO good to go

GuillaumeGomez:
My expertise in this area is very close to nothing so if it seems good to you, then it's good for me. :)

sdroege:
Go for it then, if you like the new type for the error

GuillaumeGomez:
Then let's go!

sdroege:
This makes usage of weak references quite annoying.

Whenever having a weak reference you now need to think whether you call `downgrade()` or `downgrade_sync()`. Both will always work, but one of them will panic if `upgrade()` is called from a different thread.

Previously sending a weak ref from `downgrade()` to another thread would simply be a compiler error if the type was not sync.


I think we should find another solution here. IMHO we should revert this PR and instead provide a `SendWeakRef` wrapper that solves this by behaving like `SendCell` but not panic in `drop`?

@tmiasko @GuillaumeGomez opinions?

GuillaumeGomez:
If it requires more thoughts on user's side, then it's not good and it means the binding should be improved instead. I'm totally in favor of another solution.

sdroege:
It does things implicitly that lead to panics now and need users to think about it all the time when using the API.

My proposed solution would give a compiler error and users need to explicitly opt-in for the panic-behaviour if they want it.

GuillaumeGomez:
I really prefer this.

PR #311, opened 2018-04-10
Improve set/get_property validity checks and allow to set properties from values that are a subtype of the property type

sdroege:
See https://github.com/gtk-rs/glib/pull/308, extends that PR.

@tmiasko @GuillaumeGomez


sdroege:
I can't find any property where this could easily be tested. Anybody knows one?

GuillaumeGomez:
Nothing's coming to my mind...

sdroege:
Ah in GStreamer, giostreamsrc for examples takes `GInputStream` on the `stream` property.

sdroege:
Works in GStreamer btw, should be good to go then

GuillaumeGomez:
Thanks!

PR #313, opened 2018-04-16
Remove some functions that operate on raw i32 fds

sdroege:
These should all use RawFd, if useful at all from Rust. It's unsafe to
call any of these functions on arbitrary integers.


GuillaumeGomez:
Let's go for it for now!

PR #314, opened 2018-04-16
Add a Futures Executor around MainContext and various Future/Stream implementations

sdroege:
This Executor will also be used in GIO later for Futures around the async operations (and GSources)

### DIFF
--- a/src/main_context_futures.rs
+++ b/src/main_context_futures.rs
@@ -71,7 +71,7 @@ unsafe extern "C" fn prepare(
         // XXX: This is not actually correct, we should not dispatch the
         // GSource here already but we need to know its current status so
         // that if it is not ready yet something can register to the waker
-        if let Async::Ready(_) = source.poll() {
+        if let Async::Ready(()) = source.poll() {
             source.state.store(DONE, Ordering::SeqCst);
             cur = DONE;
         } else {
@@ -114,7 +114,7 @@ unsafe extern "C" fn dispatch(
         .state
         .compare_and_swap(READY, NOT_READY, Ordering::SeqCst);
     if cur == READY {
-        if let Async::Ready(_) = source.poll() {
+        if let Async::Ready(()) = source.poll() {
             source.state.store(DONE, Ordering::SeqCst);
             cur = DONE;
         } else {
--- a/src/main_context_futures.rs
+++ b/src/main_context_futures.rs
@@ -278,3 +278,77 @@ impl Executor for MainContext {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::thread;
+    use std::sync::mpsc;
+    use futures_channel::oneshot;
+    use futures_util::future;
+
+    #[test]
+    fn test_spawn() {
+        let mut c = MainContext::new();
+        let l = ::MainLoop::new(Some(&c), false);
+
+        let (sender, receiver) = mpsc::channel();
+        let (o_sender, o_receiver) = oneshot::channel();
+
+        let l_clone = l.clone();
+        c.spawn(o_receiver
+                .map_err(|_| unimplemented!())
+                .and_then(move |()| {
+                    sender.send(()).unwrap();
+                    l_clone.quit();
+
+                    Ok(())
+                })
+        );
+
+        thread::spawn(move || {
+            l.run();
+        });
+
+        o_sender.send(()).unwrap();
+
+        let _ = receiver.recv().unwrap();
+    }
+
+    #[test]
+    fn test_spawn_local() {
+        let mut c = MainContext::new();
+        let l = ::MainLoop::new(Some(&c), false);
+
+        c.push_thread_default();
+        let l_clone = l.clone();
+        c.spawn_local(future::lazy(move |_ctx| {
+            l_clone.quit();
+
+            Ok(())
+        }));
+
+        l.run();
+
+        c.pop_thread_default();
+    }
+
+    #[test]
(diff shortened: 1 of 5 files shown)


EPashkin:
`cargo build --features futures` on Windows
```
error[E0277]: the trait bound `*mut libc::c_void: std::marker::Send` is not satisfied in `[closure@src\source_futures.rs:157:23: 162:6 pid:source::Pid, priority:source::Priority]`
   --> src\source_futures.rs:157:5
    |
157 |     SourceFuture::new(move |send| {
    |     ^^^^^^^^^^^^^^^^^ `*mut libc::c_void` cannot be sent between threads safely
    |
    = help: within `[closure@src\source_futures.rs:157:23: 162:6 pid:source::Pid, priority:source::Priority]`, the trait `std::marker::Send` is not implemented for `*mut libc::c_void`
    = note: required because it appears within the type `source::Pid`
    = note: required because it appears within the type `[closure@src\source_futures.rs:157:23: 162:6 pid:source::Pid, priority:source::Priority]`
note: required by `<source_futures::SourceFuture<F, T>>::new`
   --> src\source_futures.rs:31:5
    |
31  |     pub fn new(create_source: F) -> Box<Future<Item = T, Error = Never>> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0277]: the trait bound `*mut libc::c_void: std::marker::Send` is not satisfied in `std::option::Option<(source::Pid, i32)>`
   --> src\source_futures.rs:159:9
    |
159 |         ::child_watch_source_new(pid, None, priority, move |pid, code| {
    |         ^^^^^^^^^^^^^^^^^^^^^^^^ `*mut libc::c_void` cannot be sent between threads safely
    |
    = help: within `std::option::Option<(source::Pid, i32)>`, the trait `std::marker::Send` is not implemented for `*mut libc::c_void`
    = note: required because it appears within the type `source::Pid`
    = note: required because it appears within the type `(source::Pid, i32)`
    = note: required because it appears within the type `std::option::Option<(source::Pid, i32)>`
    = note: required because of the requirements on the impl of `std::marker::Send` for `futures_channel::lock::Lock<std::option::Option<(source::Pid, i32)>>`
    = note: required because it appears within the type `futures_channel::oneshot::Inner<(source::Pid, i32)>`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<futures_channel::oneshot::Inner<(source::Pid, i32)>>`
    = note: required because it appears within the type `futures_channel::oneshot::Sender<(source::Pid, i32)>`
    = note: required because it appears within the type `std::option::Option<futures_channel::oneshot::Sender<(source::Pid, i32)>>`
    = note: required because it appears within the type `[closure@src\source_futures.rs:159:55: 161:10 send:std::option::Option<futures_channel::oneshot::Sender<(source::Pid, i32)>>]`
note: required by `source::child_watch_source_new`
   --> src\source.rs:352:1
    |
352 | / pub fn child_watch_source_new<'a, N: Into<Option<&'a str>>, F>(pid: Pid, name: N, priority: Priority, func: F) -> Source
353 | | where F: FnMut(Pid, i32) + Send + 'static {
354 | |     unsafe {
355 | |         let source = glib_ffi::g_child_watch_source_new(pid.0);
...   |
366 | |     }
367 | | }
    | |_^

error: aborting due to 2 previous errors
```

sdroege:
Ah good. @EPashkin are pids `Send` on Windows? I assume they are, just that we don't express them as such currently?

sdroege:
I've added a commit to that effect

EPashkin:
@sdroege https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#GPid
says that process id on windows is Handle, and IMHO it Send.

EPashkin:
Build and test with `--features futures` passed for me locally. Thanks

sdroege:
Ok, then we only need someone to review all this and then can get it merged :)

src/main_context_futures.rs
```diff
@@ -0,0 +1,392 @@
...
+    *timeout = -1;
+
+    let mut cur = source
+        .state
+        .compare_and_swap(INIT, NOT_READY, Ordering::SeqCst);
+    if cur == INIT {
+        // XXX: This is not actually correct, we should not dispatch the
+        // GSource here already but we need to know its current status so
+        // that if it is not ready yet something can register to the waker
+        if let Async::Ready(_) = source.poll() {
```
EPashkin: `_` can be only `()` ?

sdroege: Yes, all tasks/futures that are spawned on the executor must result either `()` or `Never` (i.e. they also must be infallible)

EPashkin: Then maybe use `Async::Ready(())` in if lets ?

EPashkin:
@sdroege IMHO PR looks very good, Thanks.

src/main_context_futures.rs
sdroege: Done

sdroege:
Hrm, due to moving from impl trait to boxed futures this lost the ability to have `Send` implemented on the resulting future automatically. Will try to find a solution for that.

EPashkin:
@GuillaumeGomez I confirming that examples builds fine with this PR,
on test on CI is same path problem,
so seems it time to merge this.

GuillaumeGomez:
Thanks, let's go!

PR #316, opened 2018-04-17
Mark g_main_context_default as non-nullable.

tmiasko:
The default main context is always available and non-null. Also run
test in a new separate context to avoid interference between tests.


sdroege:
Ah that's the reason for the test failures! I also ran into that with the Futures test and fixed it like that.

Go for it :)

sdroege:
cc @GuillaumeGomez

GuillaumeGomez:
Yeay!

PR #319, opened 2018-04-18
Provide panic-safe way to change thread default context.

tmiasko:
Alternative proposal to #309.


sdroege:
Makes sense and is nicer than my solution

EPashkin:
@GuillaumeGomez this seems ready too, only same path error on appveyour,

GuillaumeGomez:
Thanks!

Issue #279, opened 2017-12-23
glib::KeyFile::get_boolean() returns () for both true and false


0x3UH4224D:
glib::KeyFile::get_boolean() returns () for both true and false values.

conf file:
```
[foo]
boo=false;
zoo=true;
```

usage
```
use glib;
let conf = glib::KeyFile::new();
conf.load_from_file("./conf", glib::KeyFileFlags::NONE)?;

let boo = conf.get_boolean("foo", "boo"); // the returned value suppose to be Ok(false) but I got Ok(())
let zoo = conf.get_boolean("foo", "zoo"); // the same here, I got Ok(()) while I should get Ok(true)
```

sdroege:
Needs to be manually implemented so that the return value ends up in the Result instead of being discarded

EPashkin:
This can be closed

PR #315, opened 2018-04-16
Fix dox

GuillaumeGomez:

PR #320, opened 2018-04-19
Make KeyFile get_boolean and has_key return bool

bvinc:


bvinc:
Should close #279

src/key_file.rs
```diff
@@ -70,4 +70,21 @@ impl KeyFile {
             from_glib_full(ret)
         }
     }
+
+    pub fn get_boolean(&self, group_name: &str, key: &str) -> Result<bool, Error> {
+        unsafe {
+            let mut error = ptr::null_mut();
+            let ret = ffi::g_key_file_get_boolean(self.to_glib_none().0, group_name.to_glib_none().0, key.to_glib_none().0, &mut error);
+            if error.is_null() { Ok(from_glib(ret)) } else { Err(from_glib_full(error)) }
```
EPashkin: As I understand from https://github.com/gtk-rs/gir/issues/580 this function better use `g_key_file_get_value` internally?

EPashkin: Or this if need 3th case if g_key_file_get_boolean return false and error = null for `false` value

bvinc: I'm not sure what you mean.  The problem that I'm fixing is that GIR assumed that the `gboolean` return value from g_key_file_get_boolean was redundant and useless, and so it generated a function that dropped the `gboolean` return value and returned `Result<(), Error>` instead.  This is an OK assumption because it's often the right thing to do for functions that throw and also return `gboolean`.

But in this particular case, the `gboolean` return value is exactly what we want to return back to the caller.

The `&mut error` parameter is still important and still needs to be checked.

It looks good to me.

EPashkin: Oops, I missed that if condition changed.
No problem for me now.

EPashkin:
Also now you think that `g_spawn_check_exit_status` don't need fix?

bvinc:
PR #321 
I decided to do it in a different PR since it's a weird one off that might need more review.

EPashkin:
@GuillaumeGomez only problem is `canonicalize()` but IMHO it better fixed in #315.
By adding to assert it only for macs.

GuillaumeGomez:
It doesn't work anyway so I think we'll just ignore macos...

PR #321, opened 2018-04-19
Fix spawn_check_exit status to make sense and return bool

bvinc:
I'm not sure who is going to be calling this function, but it's a little silly to have it returning a `Result<(), Error>` which can't provide the caller with any useful information.  This fixes the return value.

I ended up adding a new file, `src/functions.rs` with only this one function in it.

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -271,6 +271,10 @@ status = "generate"
     pattern = "POLLFD_FORMAT"
     #for C printf
     ignore = true
+    [[object.function]]
+    name = "spawn_check_exit_status"
+    #boolean return value needs to be returned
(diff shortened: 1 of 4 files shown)


EPashkin:
From upstream source code `g_spawn_check_exit_status` has no cases of returning false and error==null,
so IMHO generated return type is good for rust

https://github.com/GNOME/glib/blob/master/glib/gspawn.c#L886-L926
https://github.com/GNOME/glib/blob/master/glib/gspawn-win32.c#L1222-L1239

EPashkin:
@GuillaumeGomez, @sdroege please look at this PR

PR #323, opened 2018-04-20
Fix cfg attribute for RawFd.

tmiasko:
Avoid defining RawFd twice when feature v2_36 is enabled.

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -4,9 +4,9 @@
 
 use std::cell::RefCell;
 use std::mem::transmute;
-#[cfg(any(feature = "v2_36", unix))]
+#[cfg(all(unix, any(feature = "v2_36", feature = "dox")))]
 use std::os::unix::io::RawFd;
-#[cfg(any(feature = "v2_36", windows))]
+#[cfg(all(not(unix), feature = "dox"))]
 use libc::c_int as RawFd;
 use std::process;
 use std::thread;


src/source.rs
```diff
@@ -4,9 +4,9 @@
 
 use std::cell::RefCell;
 use std::mem::transmute;
-#[cfg(any(feature = "v2_36", unix))]
+#[cfg(any(all(feature = "v2_36", unix), feature = "dox"))]
 use std::os::unix::io::RawFd;
-#[cfg(any(feature = "v2_36", windows))]
+#[cfg(all(feature = "v2_36", windows))]
 use libc::c_int as RawFd;
```
EPashkin: Seems this will be defined twice if `features `dox v2_36` on windows.

tmiasko: Should be fixed now, thanks.

src/source.rs
```diff
@@ -4,9 +4,9 @@
 
 use std::cell::RefCell;
 use std::mem::transmute;
-#[cfg(any(feature = "v2_36", unix))]
+#[cfg(all(unix, any(feature = "v2_36", feature = "dox")))]
 use std::os::unix::io::RawFd;
-#[cfg(any(feature = "v2_36", windows))]
+#[cfg(all(windows, any(feature = "v2_36", feature = "dox")))]
 use libc::c_int as RawFd;
```
EPashkin: Now it don't build docs for other that unix/windows.
Maybe use `cfg(all(not(unix), feature = "dox"))`, it works for me.

tmiasko: Ok, defaulting to libc::c_int should be fine, but does glib actually support anything else?

EPashkin: IMHO if no `libc::c_int` then platform don't have support at all.

EPashkin:
@tmiasko Thanks for insisting fix.
@GuillaumeGomez  CI passed without warnings, this IMHO ready for merge.

GuillaumeGomez:
Indeed, thanks a lot!

PR #325, opened 2018-04-22
Add a new futures-nightly feature that uses futures-stable

sdroege:
This allows using async-await when enabled, but requires nightly for now.


EPashkin:
@sdroege Thanks.
👍

PR #330, opened 2018-04-30
OsString support and regen

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/592


GuillaumeGomez:
If it's just a wrapper above the C utils, there is no point to bind it since Rust has its own methods to interact with environment. I think you took the good decision by removing this (useless) step.

sdroege:
Yes, remove

EPashkin:
Updated

tmiasko:
Is there any reason for having functions designed to work with environment,
g_get_environ, g_listenv, g_unsetenv, g_getenv, etc., exposed in Rust?
Especially that Rust ones are a little bit more thread-safe, i.e.,
they are synchronized internally.

GuillaumeGomez:
No, thqt's why we said that he'd better remove them.

tmiasko:
Those in Gio.AppLaunchContext and ApplicationCommandLine are useful and without direct Rust replacement.

EPashkin:
@GuillaumeGomez Travis single failure related to futures error, @sdroege already send PR.

GuillaumeGomez:
Let's merge then. Thanks!

PR #331, opened 2018-05-04
Add WeakRef::new() for creating a new, empty WeakRef

sdroege:


GuillaumeGomez:
:+1:

src/object.rs
```diff
@@ -963,6 +963,20 @@ pub struct WeakRef<T: ?Sized> {
     marker: PhantomData<*const T>,
 }
 
+impl<T> WeakRef<T> {
+    pub fn new() -> Self {
+        unsafe {
+            let weak = WeakRef {
+                inner: Arc::new(WeakRefInner(mem::zeroed())),
+                thread: Some(thread::current().id()),
```
tmiasko: None? It should be safe regardless of thread.

src/object.rs
```diff
@@ -963,6 +963,20 @@ pub struct WeakRef<T: ?Sized> {
     marker: PhantomData<*const T>,
 }
 
+impl<T> WeakRef<T> {
+    pub fn new() -> Self {
```
tmiasko: Maybe Default::default instead / in addition?

src/object.rs
sdroege: True

src/object.rs
sdroege: Either would be fine. What would people prefer?

sdroege:
Updated

sdroege:
@GuillaumeGomez should be good to go :)

GuillaumeGomez:
Indeed. Thanks!

Issue #332, opened 2018-05-06
Rename MainContext::ref_thread_default() to ::get_thread_default() and remove the other one


sdroege:
The behaviour of `get_thread_default` to return `None` for the default main context is not very intuitive or useful, and that's the only difference in behaviour on the Rust bindings side.

I would replace it with `ref_thread_default`.

CC @GuillaumeGomez @tmiasko @EPashkin @russel

EPashkin:
Currently `ref_thread_default` return Option too, but agree that it return None only if constructor fail.
So 👍

sdroege:
`ref_thread_default` never returns `None` iirc (unless all is broken), will check the C code if people agree that this is the way forward though

EPashkin:
Problems:
Documentation will be wrong after rename,
Not sure if this real case, but it possible for someone need to know has current thread specific MainContext or not.

tmiasko:
After looking at C code, making g_main_context_ref_thread_default non-nullable looks fine.
But I would be wary of renaming it. I find it beneficial that one can switch to another 
language and get the same semantics from bindings. Renaming will only add to the confusion
for people who already know GLib or look at documentation for C.

sdroege:
Then let's go with that :)

sdroege:
Thanks for your input

PR #333, opened 2018-05-06
Mark MainContext::ref_thread_default() as non-nullable with regards t…

sdroege:
…o the return value

Fixes https://github.com/gtk-rs/glib/issues/332


GuillaumeGomez:
Yup!

PR #335, opened 2018-05-09
Only use from_glib_none() in Object::new() if the object is actually …

sdroege:
…floating

Some GObjects are based on GInitiallyUnowned but their constructor
functions return an already sunken floating reference.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -613,7 +613,7 @@ impl Object {
             let ptr = gobject_ffi::g_object_newv(type_.to_glib(), params_c.len() as u32, mut_override(params_c.as_ptr()));
             if ptr.is_null() {
                 Err(BoolError("Can't instantiate object"))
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
:+1:

tmiasko:
I would presume that those kind of problems don't happen other way around, 
so that looks reasonable. 

It is somewhat unfortunate that we have to put workarounds like that, 
could you give some examples where this happens?

sdroege:
https://bugzilla.gnome.org/show_bug.cgi?id=743062

sdroege:
Also floating references are just a workaround and historical mistake because people were too lazy 15 years ago to change a lot of GTK-based code. It's a disaster :)

EPashkin:
@sdroege Thanks, all LGFM

sdroege:
@GuillaumeGomez let's get this in then :)

GuillaumeGomez:
Yup, thanks!

PR #336, opened 2018-05-09
Add Array type

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gir/issues/596

cc @EPashkin @sdroege

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -786,6 +786,140 @@ impl Drop for HashTable {
     }
 }
 
+impl<'a, T> ToGlibContainerFromSlice<'a, *const glib_ffi::GArray> for &'a T
+where T: GlibPtrDefault + ToGlibPtr<'a, <T as GlibPtrDefault>::GlibType> {
+    type Storage = (Option<Array>, Vec<Stash<'a, <T as GlibPtrDefault>::GlibType, &'a T>>);
+
+    #[inline]
+    fn to_glib_none_from_slice(t: &'a [&'a T]) -> (*const glib_ffi::GArray, Self::Storage) {
+        let (list, stash) = ToGlibContainerFromSlice::<*mut glib_ffi::GArray>::to_glib_none_from_slice(t);
+        (list as *const glib_ffi::GArray, stash)
+    }
+
+    #[inline]
+    fn to_glib_container_from_slice(_t: &'a [&'a T]) -> (*const glib_ffi::GArray, Self::Storage) {
+        unimplemented!()
+    }
+
+    #[inline]
+    fn to_glib_full_from_slice(_t: &[&'a T]) -> *const glib_ffi::GArray {
+        unimplemented!()
+    }
+}
+
+pub struct Array(*mut glib_ffi::GArray);
+
+impl Drop for Array {
+    fn drop(&mut self) {
+        unsafe { glib_ffi::g_array_free(self.0, true.to_glib()); }
+    }
+}
+
+impl<'a, T> ToGlibContainerFromSlice<'a, *mut glib_ffi::GArray> for T
+where T: GlibPtrDefault + ToGlibPtr<'a, <T as GlibPtrDefault>::GlibType> {
+    type Storage = (Option<Array>, Vec<Stash<'a, <T as GlibPtrDefault>::GlibType, T>>);
+
+    #[inline]
+    fn to_glib_none_from_slice(t: &'a [T]) -> (*mut glib_ffi::GArray, Self::Storage) {
+        let stash_vec: Vec<_> =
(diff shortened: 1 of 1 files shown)


src/translate.rs
```diff
@@ -786,6 +786,77 @@ impl Drop for HashTable {
...
+    fn to_glib_full_from_slice(_t: &[&'a T]) -> *const glib_ffi::GArray {
+        unimplemented!()
+    }
+}
+
+pub struct Array(*mut glib_ffi::GArray);
+
+impl Drop for Array {
+    fn drop(&mut self) {
+        unsafe { glib_ffi::g_array_free(self.0, false.to_glib()); }
```
sdroege: Needs to be `true`, not `false`

sdroege:
The `From*` impls are still missing

src/translate.rs
```diff
@@ -786,6 +786,77 @@ impl Drop for HashTable {
...
+    }
+}
+
+impl<'a, T> ToGlibContainerFromSlice<'a, *mut glib_ffi::GArray> for T
```
sdroege: This will only work for storing pointers inside the `GArray`. Other use cases are plain fundamental values and storing `struct`s by value. The latter is what @russel actually needs AFAIU.

sdroege: Basically the difference between `Vec<T>` and `Vec<Box<T>>`

GuillaumeGomez: Oh I see. I'll update then.

src/translate.rs
```diff
@@ -786,6 +786,77 @@ impl Drop for HashTable {
...
+where T: GlibPtrDefault + ToGlibPtr<'a, <T as GlibPtrDefault>::GlibType> {
+    type Storage = (Option<Array>, Vec<Stash<'a, <T as GlibPtrDefault>::GlibType, T>>);
+
+    #[inline]
+    fn to_glib_none_from_slice(t: &'a [T]) -> (*mut glib_ffi::GArray, Self::Storage) {
+        let stash_vec: Vec<_> =
+            t.iter().map(|v| v.to_glib_none()).collect();
+        let mut arr: *mut glib_ffi::GArray = ptr::null_mut();
```
tmiasko: Unlike GSList for example, NULL is not a valid way to initialize array. The same below.

GuillaumeGomez:
Updated. Is it what you had in mind @sdroege?

src/translate.rs
```diff
@@ -786,6 +786,140 @@ impl Drop for HashTable {
...
+                                                                             t.len() as u32);
+            for ptr in t.iter().map(|v| v.to_glib_full()) {
+                arr = glib_ffi::g_array_append_vals(arr, Ptr::to(ptr), 1);
+            }
+            arr
+        }
+    }
+}
+
+impl<T> FromGlibContainerAsVec<<T as GlibPtrDefault>::GlibType, *mut glib_ffi::GArray> for T
```
sdroege: We probably also want one for `*const` here

sdroege:
This looks ok, but someone also needs to actually test it. And I don't think it will work for the case I mentioned in https://github.com/gtk-rs/glib/pull/336#discussion_r187118119 . I.e. an array that contains a flat struct instead of pointers to the struct. I guess we'll get to that once it's needed :)

Also you can trivially add `GPtrArray` bindings now, it's mostly copy&paste of what you did now. The only part that is not the same between the two is the case I mentioned above.

sdroege:
Also I believe your changes only work for pointer types anyway right now? E.g. a `GArray<u16>` would not work, which is what @russel would need.

You should be able to take the same pattern as for plain arrays for those fundamental types.

GuillaumeGomez:
Ok, I added `PtrArray` and I starting to get completely lost in my own code... I think we should *really* split this file into smaller ones because it's getting really too huge. Also, I might need some help to implement what @russel needs exactly.

sdroege:
We should maybe add some tests for these things also. What he's going to need is `GPtrArray` and `GArray<u16>` from what I can see. I think the stuff you implemented so far only allows for `GArray` containing pointers. You need the other traits that we implement for e.g. `Vec<u8>` / `&[u8]`

GuillaumeGomez:
Ok, I'll give it a try.

GuillaumeGomez:
Discussion continued with as follow: writing a new interface between traits, we need the traits in order to have something working like this:

```
type.trait() -> trait.trait() -> real_type.trait()
```

It provides the big advantage to not have to provide an implementation for every type (which is a big plus).

PS: @sdroege will improve the explanations later on. 😆

elmarco:
@GuillaumeGomez  #482 should deprecate that PR

GuillaumeGomez:
Indeed, it's completely useless. Let's close it then!

PR #340, opened 2018-06-18
add To/FromValue trait to VariantType

vhdirk:
This PR enables wrapping a ```VariantTy(pe)``` in ```Value```

### DIFF
--- a/src/variant_type.rs
+++ b/src/variant_type.rs
@@ -226,15 +226,23 @@ impl SetValue for VariantTy {
 
 impl SetValueOptional for VariantTy {
     unsafe fn set_value_optional(value: &mut Value, this: Option<&Self>) {
-        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
+        use std::ptr;
+        let p = match this{
+            Some(ref t) => t.to_glib_none().0 as glib_ffi::gpointer,
+            None => ptr::null()
+        };
+        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, p)
     }
 }
 
 impl<'a> FromValueOptional<'a> for &'a VariantTy {
     unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
         let cvty = gobject_ffi::g_value_get_boxed(value.to_glib_none().0) as *mut glib_ffi::GVariantType;
-        assert!(!cvty.is_null());
-        Some(VariantTy::from_ptr(cvty))
+        if cvty.is_null() {
+            None
+        } else {
+            Some(VariantTy::from_ptr(cvty))
+        }
     }
 }
 
@@ -250,11 +258,16 @@ impl SetValue for VariantType {
     }
 }
 
-// impl SetValueOptional for VariantType {
-//     unsafe fn set_value_optional(value: &mut Value, this: Option<&Self>) {
-//         gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
-//     }
-// }
+impl SetValueOptional for VariantType {
+    unsafe fn set_value_optional(value: &mut Value, this: Option<&Self>) {
+        use std::ptr;
+        let p = match this{
+            Some(ref t) => t.to_glib_none().0 as glib_ffi::gpointer,
+            None => ptr::null()
+        };
+        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, p)
+    }
+}
 
 impl<'a> FromValueOptional<'a> for VariantType {
     unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
@@ -412,6 +425,9 @@ mod tests {
         let tyv2 = ty1.to_value();
         let ty4 = tyv2.get::<VariantType>().unwrap();
         assert_eq!(ty3, ty4);
+
+        assert_eq!(VariantTy::static_type(), VariantTy::static_type());
+
     }
 
 }
--- a/src/variant_type.rs
+++ b/src/variant_type.rs
@@ -233,7 +233,7 @@ impl SetValueOptional for VariantTy {
 impl<'a> FromValueOptional<'a> for &'a VariantTy {
     unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
         let cvty = gobject_ffi::g_value_get_boxed(value.to_glib_none().0) as *mut glib_ffi::GVariantType;
-        assert!(cvty.is_null());
+        assert!(!cvty.is_null());
         Some(VariantTy::from_ptr(cvty))
     }
 }
@@ -407,6 +407,11 @@ mod tests {
         let tyv = ty1.to_value();
         let ty2 = tyv.get::<VariantType>().unwrap();
         assert_eq!(ty1, ty2);
+
+        let ty3 = VariantTy::new("*").unwrap();
+        let tyv2 = ty1.to_value();
+        let ty4 = tyv2.get::<VariantType>().unwrap();
+        assert_eq!(ty3, ty4);
     }
 
 }
--- a/src/variant_type.rs
+++ b/src/variant_type.rs
@@ -12,7 +12,7 @@ use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::ops::Deref;
 use std::slice;
-use value::{Value, ToValue, SetValue, FromValueOptional};
+use value::{Value, ToValue, SetValue, FromValueOptional, SetValueOptional};
 use gobject_ffi;
 
 /// Describes `Variant` types.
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
cc @sdroege @EPashkin

src/variant_type.rs
```diff
@@ -208,6 +213,36 @@ impl ToOwned for VariantTy {
...
+
+impl StaticType for VariantType {
+    fn static_type() -> Type {
+        unsafe { from_glib(glib_ffi::g_variant_type_get_gtype()) }
+    }
+}
+
+impl SetValue for VariantTy {
+    unsafe fn set_value(value: &mut Value, this: &Self) {
+        gobject_ffi::g_value_take_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
```
sdroege: Shouldn't this be `set_boxed`? The rust part takes a reference, not ownership of the variant ty

src/variant_type.rs
```diff
@@ -2,6 +2,9 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+use libc::{c_void};
```
sdroege: Unneeded {}

src/variant_type.rs
```diff
@@ -208,6 +213,36 @@ impl ToOwned for VariantTy {
...
+
+impl SetValue for VariantTy {
+    unsafe fn set_value(value: &mut Value, this: &Self) {
+        gobject_ffi::g_value_take_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
+    }
+}
+
+impl SetValueOptional for VariantTy {
+    unsafe fn set_value_optional(value: &mut Value, this: Option<&Self>) {
+        gobject_ffi::g_value_take_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
```
sdroege: Is NULL a valid VariantTy?

vhdirk: Probably not. I guess NULL would be ```"*"```?

sdroege: Needs to be checked in the GLib code, in the implementation of the boxed<->variantty code. Will check tomorrow if nobody else is faster :)

sdroege: It goes into `g_variant_type_copy()`, which says that it must not be `NULL`. So no `SetValueOptional` for both and you can implement `FromValue` for both

src/variant_type.rs
```diff
@@ -208,6 +212,47 @@ impl ToOwned for VariantTy {
...
+impl SetValue for VariantTy {
+    unsafe fn set_value(value: &mut Value, this: &Self) {
+        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
+    }
+}
+
+impl<'a> FromValueOptional<'a> for &'a VariantTy {
+    unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
+        let cvty = gobject_ffi::g_value_dup_boxed(value.to_glib_none().0) as *mut glib_ffi::GVariantType;
+        if cvty.is_null() {
```
sdroege: assert for NULL here as it's not allowed

src/variant_type.rs
```diff
@@ -208,6 +212,47 @@ impl ToOwned for VariantTy {
...
+
+impl SetValue for VariantTy {
+    unsafe fn set_value(value: &mut Value, this: &Self) {
+        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
+    }
+}
+
+impl<'a> FromValueOptional<'a> for &'a VariantTy {
+    unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
+        let cvty = gobject_ffi::g_value_dup_boxed(value.to_glib_none().0) as *mut glib_ffi::GVariantType;
```
sdroege: You're leaking the memory here, this has to be `get_boxed()`

src/variant_type.rs
```diff
@@ -208,6 +212,47 @@ impl ToOwned for VariantTy {
...
+
+impl SetValue for VariantType {
+    unsafe fn set_value(value: &mut Value, this: &Self) {
+        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
+    }
+}
+
+impl<'a> FromValueOptional<'a> for VariantType {
+    unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
+        Option::<VariantType>::from_glib_full(gobject_ffi::g_value_dup_boxed(value.to_glib_none().0) as *mut glib_ffi::GVariantType)
```
sdroege: This should be `from_glib_none()` and `get_boxed()`.

See also this comment in the code from `from_glib_full`:

```rust
impl FromGlibPtrFull<*const glib_ffi::GVariantType> for VariantType {
    unsafe fn from_glib_full(ptr: *const glib_ffi::GVariantType) -> VariantType {
        // Don't assume ownership of a const pointer.
        // A transfer: full annotation on a `const GVariantType*` is likely a bug.
        VariantTy::from_ptr(ptr).to_owned()
    }
}
```

src/variant_type.rs
sdroege: So needed: `StaticType`, `FromValueOptional`, `FromValue` and `SetValue` for both types

vhdirk:
How come a VariantType can't be null? This test seems to tell otherwise? https://gitlab.gnome.org/GNOME/glib/blob/master/gio/tests/actions.c

sdroege:
> How come a VariantType can't be null? This test seems to tell otherwise? https://gitlab.gnome.org/GNOME/glib/blob/master/gio/tests/actions.c

Inside `GValue` it is using `g_variant_type_copy()`, which will assert if `NULL` is passed:
https://gitlab.gnome.org/GNOME/glib/blob/b9aed426d198ccf27078263d982bb247c82ccaee/glib/gvarianttype.c#L330 and https://gitlab.gnome.org/GNOME/glib/blob/b9aed426d198ccf27078263d982bb247c82ccaee/glib/gvarianttype.c#L184-188

That might be a bug on the GLib side though. Do you need it to be `NULL` for anything?

src/variant_type.rs
```diff
@@ -224,14 +224,17 @@ impl SetValue for VariantTy {
...
 impl<'a> FromValueOptional<'a> for &'a VariantTy {
     unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
-        let cvty = gobject_ffi::g_value_dup_boxed(value.to_glib_none().0) as *mut glib_ffi::GVariantType;
-        if cvty.is_null() {
-            None
-        } else {
-            Some(VariantTy::from_ptr(cvty))
-        }
+        let cvty = gobject_ffi::g_value_get_boxed(value.to_glib_none().0) as *mut glib_ffi::GVariantType;
+        assert!(cvty.is_null());
```
sdroege: This is missing a negation. You want to assert that it is not `NULL`

vhdirk:
> That might be a bug on the GLib side though. Do you need it to be NULL for anything?

Not necessarily. Only for aping SimpleAction

sdroege:
> > That might be a bug on the GLib side though. Do you need it to be NULL for anything?
> 
> Not necessarily. Only for aping SimpleAction

How does it work in `SimpleAction` (why does it not run into these assertions?), what are the `NULL` types used for exactly?

vhdirk:
The docs of SimpleAction state:
> The type of the parameter that must be given when activating the action.

I've never used GAction before, so I'm not sure what that means. And my rust version of SimpleAction is still panicking somewhere without a clear trace...

vhdirk:
More from the docs

> Since GLib 2.40, if no handler is connected to this signal then the default behaviour for boolean-stated actions with a NULL parameter type is to toggle them via the “change-state” signal.

sdroege:
That hints at a bug in GLib then, that clearly says that `NULL` types are expected behaviour... but running `g_boxed_copy()` on such `GValue`s (as from the signals) would assert. Can you file a bug against GLib about that?

vhdirk:
From what I can tell, it is not so much GValueType that is not nullable, rather the type string itself?

sdroege:
> From what I can tell, it is not so much GValueType that is not nullable, rather the type string itself?

I'm not sure what you mean with this. `GVariantType` seems to be possible to be `NULL` in various contexts, just not when stored in a `GValue` (but `GSimpleAction` seems to use it as such nonetheless).

vhdirk:
Looking at the implementation of ```g_variant_type_copy```, I think the documentation is wrong.

```
GVariantType *
g_variant_type_copy (const GVariantType *type)
{
  gsize length;
  gchar *new;

  g_return_val_if_fail (g_variant_type_check (type), NULL);

  <snip>
}
```

There's no assertion here. It will just return another null pointer when you provide it a null pointer.

sdroege:
> g_return_val_if_fail (g_variant_type_check (type), NULL);

This is an assertion (it will return `FALSE` for `NULL`)

vhdirk:
Are you sure?

According to the docs:
```
#define             g_return_val_if_fail(expr,val)
```
It will return ```val``` if ```expr``` evaluates to ```FALSE```.

So in this case, it will return ```NULL``` when ```g_variant_type_check(type)``` returns ```FALSE```. Which is the case when ```type``` is ```NULL``` to begin with.

sdroege:
It's only to be used as a guard against programming mistakes. It's an assertion, but not taking down your application unless yet set `G_DEBUG=fatal_criticals`. It causes a `g_critical()` warning, which is also printed on stderr.

vhdirk:
Ah, I though that was just a shorthand.  So yeah, this is a bug in glib then.

sdroege:
> Ah, I though that was just a shorthand. So yeah, this is a bug in glib then.

Lovely :) Not sure what we can do about that, need a workaround until a new GLib version arrives. But your code making it nullable is correct then, just GLib isn't correct.

sdroege:
I would suggest to file a bug there and go ahead with your implementation. It will not cause unsafety, it only causes ugly warnings on the terminal and those are GLib's problem

src/variant_type.rs
```diff
@@ -208,6 +212,56 @@ impl ToOwned for VariantTy {
...
+    }
+}
+
+impl SetValue for VariantType {
+    unsafe fn set_value(value: &mut Value, this: &Self) {
+        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as glib_ffi::gpointer)
+    }
+}
+
+// impl SetValueOptional for VariantType {
```
sdroege: You can include this one now :)

vhdirk:
I'll create the glib bug report soon.

vhdirk:
https://gitlab.gnome.org/GNOME/glib/issues/1427

sdroege:
@GuillaumeGomez Let's get this in then?

GuillaumeGomez:
Yes! Thanks @vhdirk!

PR #342, opened 2018-06-23
Expose signal::signal_handler_disconnect

philn:


EPashkin:
@philn Thanks
👍

philn:
BTW why is this function not taking a reference like the others?

EPashkin:
Because disconnecting can be happened only one time

GuillaumeGomez:
So we all agree. Thanks @philn!

PR #343, opened 2018-06-24
Fix clippy warnings

EPashkin:
cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/main_context.rs
+++ b/src/main_context.rs
@@ -26,7 +26,7 @@ impl MainContext {
         }
     }
 
-    pub fn find_source_by_id(&self, source_id: SourceId) -> Option<Source> {
+    pub fn find_source_by_id(&self, source_id: &SourceId) -> Option<Source> {
         unsafe {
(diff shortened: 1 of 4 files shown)


src/main_context.rs
```diff
@@ -26,7 +26,7 @@ impl MainContext {
         }
     }
 
-    pub fn find_source_by_id(&self, source_id: SourceId) -> Option<Source> {
+    pub fn find_source_by_id(&self, source_id: &SourceId) -> Option<Source> {
```
EPashkin: @sdroege Strange why find want consume non-copy type

GuillaumeGomez:
👍

src/main_context.rs
GuillaumeGomez: That's very weird because I assume that we can get the `SourceId` from the `Source`. However, if no `Source` matches the `SourceId`, you just don't have the `SourceId` anymore... Maybe it is the purpose?

sdroege: Should be by reference, yes

sdroege:
Looks good to me

GuillaumeGomez:
Thanks!

PR #349, opened 2018-07-03
Add GBinding bindings

sdroege:
Kind of ugly because of the autogenerated code that can't be autogenerated due to how gir works :)

Also untested so far, please review carefully. Will report back once I know that it works.


GuillaumeGomez:
Should we split apart GObject and GLib?

sdroege:
> Should we split apart GObject and GLib?

GLib by itself is rather useless in Rust IMHO, and the code is all entangled (just because of the GType related things already, which are from GObject but used in GLib).

GuillaumeGomez:
Indeed but it's a requirement that we need for the moment. And right now we can't generate two bindings at once if I understand things correctly.

sdroege:
> Indeed but it's a requirement that we need for the moment. And right now we can't generate two bindings at once if I understand things correctly.

I would prefer to fix gir so it can output to e.g. `src/auto_gobject` via some commandline parameter

src/object.rs
```diff
@@ -931,6 +934,57 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+                )
+            )
+        }
+    }
+
+    fn bind_property_full<'a, 'b, O: IsA<Object>, N: Into<&'a str>, M: Into<&'a str>, F: Fn(&::Binding, &Value) -> Option<Value> + Send + Sync + 'static, G: Fn(&::Binding, &Value) -> Option<Value> + Send + Sync + 'static>(&self, source_property: N, target: &O, target_property: M, flags: ::BindingFlags, transform_to: F, transform_from: G) -> Option<::Binding> {
+        let source_property = source_property.into();
+        let target_property = target_property.into();
+
+        let to_closure = ::Closure::new(move |values| {
```
sdroege: This is wrong, fixing

EPashkin:
@sdroege I thought that path be `src/gobject/auto`, this way we can separate manual code too.
But your suggestion also acceptable.

sdroege:
> @sdroege I thought that path be src/gobject/auto, this way we can separate manual code too.
> But your suggestion also acceptable.

It just has to be configurable, and more important gir has to create the use statements for GObject like for GLib

src/object.rs
```diff
@@ -655,6 +655,9 @@ pub trait ObjectExt: IsA<Object> {
     ///
     /// [downgrade_sync]: trait.SyncObjectExt.html#method.downgrade_sync
     fn downgrade(&self) -> WeakRef<Self>;
+
+    fn bind_property<'a, 'b, O: IsA<Object>, N: Into<&'a str>, M: Into<&'a str>>(&self, source_property: N, target: &O, target_property: M, flags: ::BindingFlags) -> Option<::Binding>;
+    fn bind_property_full<'a, 'b, O: IsA<Object>, N: Into<&'a str>, M: Into<&'a str>, F: Fn(&::Binding, &Value) -> Option<Value> + Send + Sync + 'static, G: Fn(&::Binding, &Value) -> Option<Value> + Send + Sync + 'static>(&self, source_property: N, target: &O, target_property: M, flags: ::BindingFlags, transform_to: Option<F>, transform_from: Option<G>) -> Option<::Binding>;
```
sdroege: This is suboptimal as you can't just pass `None`, it needs type annotations everywhere.

Opinions: This either needs 3 functions (first, second, both closures) or a builder (then all functions can be combined). IMHO a builder is nicer, then the flags could also become optional for example.

sdroege: Builder it is if nobody complains :)

sdroege:
Ok, please review the API if that makes sense to you :) One usage I have here right now is

```rust
        item.bind_property("age", &label, "label")
            .flags(glib::BindingFlags::DEFAULT | glib::BindingFlags::SYNC_CREATE)
            .transform_from(|_binding: &glib::Binding, value: &glib::Value| -> Option<glib::Value> {
                let age = value.get::<u32>().unwrap();
                Some(format!("{}", age).to_value())
            })
            .build();
```

src/object.rs
```diff
@@ -1051,3 +1060,78 @@ impl<T: IsA<Object> + StaticType + UnsafeFrom<ObjectRef> + Wrapper + ?Sized> Wea
...
+    }
+
+    pub fn transform_to<F: Fn(&::Binding, &Value) -> Option<Value> + Send + Sync + 'static>(self, func: F) -> Self {
+        Self {
+            transform_to: Some(Self::transform_closure(func)),
+            ..self
+        }
+    }
+
+    pub fn flags(self, flags: ::BindingFlags) -> Self {
```
sdroege: Instead of this, one could have a `fn bidirectional(self) -> Self` etc for each of the flags

sdroege:
And autogenerated now

Gir_GObject.toml
```diff
@@ -0,0 +1,25 @@
...
+ignore = [
+]
+
+manual = [
+    "GObject.Object",
+]
+
+[[object]]
+name = "GObject.*"
+status = "ignore"
```
EPashkin: In final version this need be status = "generate"

sdroege: Why? I don't care about any of those functions. They can be added when someone needs them :)

And more important, all that can be autogenerated need manual bindings (and many of them have already, either here or in my `gobject-subclass` crate)

src/object.rs
```diff
@@ -1051,3 +1060,78 @@ impl<T: IsA<Object> + StaticType + UnsafeFrom<ObjectRef> + Wrapper + ?Sized> Wea
     }
 }
 
+pub struct BindingBuilder<'a, S: IsA<Object> + 'a, T: IsA<Object> + 'a> {
```
EPashkin: Maybe this better moved to `src/gobject/binding_builder.rs `

Gir_GObject.toml
EPashkin: Ok

src/object.rs
sdroege: Ok, will move it later if @GuillaumeGomez has no other comments

src/gobject/auto/enums.rs
```diff
@@ -0,0 +1,7 @@
+// This file was generated by gir (https://github.com/gtk-rs/gir)
```
GuillaumeGomez: Is it normal to have this empty file?

sdroege: Yes

GuillaumeGomez: Can we not have it if it's useless? :)

sdroege: Sure, same for flags I guess :) But can we do that in a separate PR, it requires some further gir changes but I can put that on my list then

GuillaumeGomez: Fine by me!

GuillaumeGomez:
Thanks!

PR #352, opened 2018-07-06
Add ParamSpec bindings

sdroege:


src/gobject/auto/param_spec.rs
```diff
@@ -0,0 +1,99 @@
...
+use object::IsA;
+use std::mem;
+use std::ptr;
+use translate::*;
+
+glib_wrapper! {
+    pub struct ParamSpec(Object<ffi::GParamSpec, ffi::GParamSpecClass>);
+
+    match fn {
+        get_type => || ffi::intern(),
```
sdroege: We need a way to override the get_type functions. This one here has a constant, `G_TYPE_PARAMSPEC`

sdroege:
For some reason things like `g_param_spec_bool`, `g_param_spec_int`, etc (all the constructors) are not generated by GIR. Not sure yet why, @EPashkin do you have an idea?

Also how would we best override the `get_type` function? Configuration that takes a string, and that is just put there instead of whatever the GIR says?

Gir_GObject.toml
```diff
@@ -11,15 +11,42 @@ single_version_file = true
...
 [[object]]
 name = "GObject.*"
 status = "ignore"
+
+[[object]]
+name = "GObject.ParamSpec"
+status = "generate"
+    [[object.function]]
+    name = "sink"
+    ignore = true # memory management
```
EPashkin: Please move comment to previous line

src/gobject/auto/param_spec.rs
EPashkin: Agree

EPashkin:
IMHO config string  `get_type: Option<String>` is enough

sdroege:
While at it, I would add the same for ref/copy/unref/free functions the same way. That allows me to get rid of some more manual bindings

EPashkin:
Thanks in advance

src/gobject/auto/param_spec.rs
```diff
@@ -0,0 +1,99 @@
...
+use gobject_ffi as ffi;
+use ffi as glib_ffi;
+use gobject_ffi;
+use object::IsA;
+use std::mem;
+use std::ptr;
+use translate::*;
+
+glib_wrapper! {
+    pub struct ParamSpec(Object<ffi::GParamSpec, ffi::GParamSpecClass>);
```
sdroege: This is actually not a `GObject` but another fundamental type. So I guess we don't really get around manually binding this one after all...

Will update the PR accordingly

sdroege:
Updated accordingly and made use of it in `object.rs` a bit

src/object.rs
```diff
@@ -673,33 +675,32 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
-            let object = self.to_glib_none().0;
-            let klass = (*object).g_type_instance.g_class as *mut gobject_ffi::GObjectClass;
-            let pspec = gobject_ffi::g_object_class_find_property(klass, property_name.to_glib_none().0);
-            if pspec.is_null() {
+        let pspec = match self.find_property(property_name) {
+            Some(pspec) => pspec,
+            None => {
                 return Err(BoolError("property not found"));
-            }
+            },
```
EPashkin: comma unneeded

src/object.rs
```diff
@@ -673,33 +675,32 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+        };
 
-            let writeable = (*pspec).flags & gobject_ffi::G_PARAM_WRITABLE != 0;
-            if !writeable {
-                return Err(BoolError("property is not writeable"));
-            }
+        if !pspec.get_flags().contains(::ParamFlags::WRITABLE) {
```
EPashkin: Maybe flags need check for `CONSTRUCT_xxx`, some properties have WRITABLE and CONSTRUCT_ONLY, so actually not writable

sdroege: Thanks, good idea :) Will change in a few minutes

src/object.rs
sdroege: done

src/object.rs
sdroege: done

src/object.rs
EPashkin: Strange, I still see it, and in other `match self.find_property` down too.
Seems you forgot to push 😉

sdroege: I indeed forgot to push it, yes :) Thanks

sdroege:
All green :)

EPashkin:
👍 then

src/param_spec.rs
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    pub fn value_array(name: &str, nick: &str, blurb: &str, element_spec: &ParamSpec, flags: ParamFlags) -> Option<ParamSpec> {
+        unsafe {
+            from_glib_none(ffi::g_param_spec_value_array(name.to_glib_none().0, nick.to_glib_none().0, blurb.to_glib_none().0, element_spec.to_glib_none().0, flags.to_glib()))
+        }
+    }
+
+
```
GuillaumeGomez: Extra empty line.

src/param_spec.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub fn variant<'a, P: Into<Option<&'a ::Variant>>>(name: &str, nick: &str, blurb: &str, type_: &::VariantTy, default_value: P, flags: ParamFlags) -> Option<ParamSpec> {
+        unsafe {
+            let default_value = default_value.into();
+            from_glib_none(ffi::g_param_spec_variant(name.to_glib_none().0, nick.to_glib_none().0, blurb.to_glib_none().0, type_.to_glib_none().0, default_value.to_glib_none().0, flags.to_glib()))
+        }
+    }
+
```
GuillaumeGomez: Extra empty line.

src/param_spec.rs
sdroege: fixed

src/param_spec.rs
sdroege: fixed

sdroege:
And green again :)

GuillaumeGomez:
Thanks!

PR #354, opened 2018-07-06
Add ValueArray bindings

sdroege:
Can't be autogenerated because all the functions return the array itself, and because it's `from_glib_none` each function would create a copy of the whole array.

Also manually added the `Deref` impls, very useful


EPashkin:
Looks good, thanks

GuillaumeGomez:
Thanks!

PR #356, opened 2018-07-11
Add downcast_ref() and similar functions for all the casting machinery

sdroege:
See individual commits.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -50,6 +50,28 @@ pub trait Cast: IsA<Object> {
         unsafe { T::from(self.into()) }
     }
 
+    /// Upcasts an object to a reference of its superclass or interface `T`.
+    ///
+    /// *NOTE*: This statically checks at compile-time if casting is possible. It is not always
+    /// known at compile-time, whether a specific object implements an interface or not, in which case
+    /// `upcast` would fail to compile. `dynamic_cast` can be used in these circumstances, which
+    /// is checking the types at runtime.
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// let button = gtk::Button::new();
+    /// let widget = button.upcast_ref::<gtk::Widget>();
+    /// ```
+    #[inline]
+    fn upcast_ref<T>(&self) -> &T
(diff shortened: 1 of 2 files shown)


sdroege:
The object ones are preventing the `obj.clone().downcast::<Foo>()` calls I have in various places

src/object.rs
```diff
@@ -50,6 +50,28 @@ pub trait Cast: IsA<Object> {
...
+    /// ```ignore
+    /// let button = gtk::Button::new();
+    /// let widget = button.upcast_ref::<gtk::Widget>();
+    /// ```
+    #[inline]
+    fn upcast_ref<T>(&self) -> &T
+    where T: StaticType + UnsafeFrom<ObjectRef> + Wrapper,
+          Self: IsA<T> {
+        unsafe {
+            mem::transmute(self)
```
sdroege: The object casts are safe because all the types have the same memory representation apart from their name and the phantom data type, and `IsA` is an unsafe trait so must only be implemented if it's actually a proper GObject wrapper type

src/value.rs
```diff
@@ -139,6 +139,24 @@ impl Value {
...
+    ///
+    /// Returns `Some(&TypedValue<T>)` if the value carries a type corresponding
+    /// to `T` and `None` otherwise.
+    pub fn downcast_ref<'a, T: FromValueOptional<'a> + SetValue>(&self) -> Option<&TypedValue<T>> {
+        unsafe {
+            let ok = from_glib(
+                gobject_ffi::g_type_check_value_holds(mut_override(self.to_glib_none().0),
+                    T::static_type().to_glib()));
+            if ok {
+                Some(mem::transmute(self))
```
sdroege: The value transmutes are safe because the only difference between `Value` and `TypedValue` is the zero-sized phantom data.

EPashkin:
@sdroege Looks good

src/value.rs
EPashkin: This comment not better be in code? ;)

GuillaumeGomez:
I clearly don't know enough this part of GLib and the code looks good to me so I'll trust you and @EPashkin's opinion on this one. Thanks!

PR #361, opened 2018-07-27
Implement work-around for thread IDs potentially being re-usable by t…

sdroege:
…he OS


GuillaumeGomez:
:+1: But I might have miss something so waiting confirmation from @EPashkin.

EPashkin:
@sdroege Thanks, good idea. 👍

PR #362, opened 2018-07-28
Add Quark type

GuillaumeGomez:


src/quark.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    pub fn intern_string(s: &str) -> Option<String> {
+        unsafe {
+            from_glib_none(ffi::g_intern_string(s.to_glib_none().0))
+        }
+    }
+
+    pub fn intern_static_string(s: &str) -> Option<String> {
```
GuillaumeGomez: I'm just worry about this implementation: do we have ownership on the returned string or not? If so, we'll have issues because we're not supposed to free it...

EPashkin: `from_glib_none` IMHO is right, but this function don't make sense in rust and better not implemented.
Also it already in https://github.com/gtk-rs/glib/blob/master/src/auto/functions.rs#L614 and better ignored with `intern_string`

GuillaumeGomez: Sounds good to me!

GuillaumeGomez:
Removed the two functions.

src/quark.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    pub fn from_static_string(s: &'static str) -> Quark {
+        unsafe {
+            from_glib(ffi::g_quark_from_static_string(s.to_glib_none().0))
+        }
+    }
+
+    pub fn to_string(&self) -> Option<String> {
+        unsafe {
+            from_glib_none(ffi::g_quark_to_string(self.to_glib()))
```
sdroege: The return value has static lifetime and is never NULL. You can directly use CStr here and return a static string slice

src/quark.rs
```diff
@@ -0,0 +1,54 @@
...
+impl Quark {
+    pub fn from_string(s: &str) -> Quark {
+        unsafe {
+            from_glib(ffi::g_quark_from_string(s.to_glib_none().0))
+        }
+    }
+
+    pub fn from_static_string(s: &'static str) -> Quark {
+        unsafe {
+            from_glib(ffi::g_quark_from_static_string(s.to_glib_none().0))
```
sdroege: This is wrong. The string passed to C must have static lifetime but due to conversion it doesn't here, is only in this scope.

We can't bind this function in a meaningful way due to the difference between C and Rust strings. The trailing 0

GuillaumeGomez: So I just remove it as well?

src/quark.rs
sdroege: The returned string here is the same as below in to_string(). Static string slice would be correct

But you can't bind the static variant, only the other. And both intern functions are not very useful in Rust

src/quark.rs
sdroege: Yes

GuillaumeGomez: Ok!

src/quark.rs
```diff
@@ -0,0 +1,54 @@
...
+use ffi;
+use translate::*;
+
+#[repr(C)]
+pub struct Quark(ffi::GQuark);
+
+impl Quark {
+    pub fn from_string(s: &str) -> Quark {
+        unsafe {
+            from_glib(ffi::g_quark_from_string(s.to_glib_none().0))
```
sdroege: If you want to prevent a copy, you could make this one here use to_glib_full() and have it call the static variant of the function. Same overall behaviour but one less copy

EPashkin:
@GuillaumeGomez Please remove generated `intern_(static_)?string` too

GuillaumeGomez:
@EPashkin They're already removed no? Or are you talking about something outside of `Quark`?

src/quark.rs
```diff
@@ -0,0 +1,50 @@
...
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use ffi;
+use translate::*;
+
+use std::ffi::CStr;
+
+#[repr(C)]
+pub struct Quark(ffi::GQuark);
```
sdroege: Now also add Eq and PartialEq and Clone and Copy impls :) You can derive them, those are the correct ones

Otherwise all good and perfect

EPashkin:
@GuillaumeGomez I talking about https://github.com/gtk-rs/glib/blob/master/src/auto/functions.rs#L614

src/quark.rs
sdroege: I guess Hash and the Ord ones too

GuillaumeGomez:
@EPashkin If I remove them, we won't have access to the intern string anywhere. If it's what you want then I'll just remove them.

src/quark.rs
GuillaumeGomez: Ok! ;)

sdroege:
That API makes no sense from Rust, you can remove :)

EPashkin:
@GuillaumeGomez Yes, I think that both function unusable in rust

GuillaumeGomez:
I did. I think it's a first for me to remove things instead of adding ones haha.

GuillaumeGomez:
Updated.

sdroege:
Perfect, I'm offline again :)

EPashkin:
👍

GuillaumeGomez:
Thanks to both of you for your reviews! :)

Issue #363, opened 2018-07-29
Move `ErrorDomain` trait to safe `GQuark` bindings


sdroege:
See summary. Now that we have `GQuark` bindings we should also make use of them :)

@GuillaumeGomez Something for you?

GuillaumeGomez:
You can do it too if you want. lot of work on rustdoc. ;)

EPashkin:
@sdroege You meant something like this?
```rust
pub trait ErrorDomain: Copy {
    /// Returns the quark identifying the error domain.
    ///
    /// As returned from `g_some_error_quark`.
    fn domain() -> Quark;
```
```rust
impl ErrorDomain for FileError {
    fn domain() -> glib_ffi::GQuark {
        unsafe { from_glib(glib_ffi::g_file_error_quark()) }
    }
```
Then it IMHO work for me as git need change too.

sdroege:
Yes, exactly that.

EPashkin:
@GuillaumeGomez All 4 PR linked to issue IMHO ready for merge

GuillaumeGomez:
I assume it's normal they're all failing?

EPashkin:
Yes, only gtk that I will add latter must pass CI before merge

GuillaumeGomez:
Done!

sdroege:
Let's close this one then, all but gtk done :)

PR #364, opened 2018-07-29
Error domain quark

EPashkin:
Part of #363 

cc @GuillaumeGomez, @sdroege


sdroege:
Looks good to go to me but needs regen everywhere

GuillaumeGomez:
Then let's regen everywhere! :)

PR #368, opened 2018-08-08
Fix build with two-phase-borrows (aka NLL)

sdroege:


GuillaumeGomez:
Why this change btw?

sdroege:
Because otherwise we would mutably borrow from something behind a const pointer, and that's not allowed and mistakingly works with the old borrow checker :)

GuillaumeGomez:
Ok! Thanks!

PR #370, opened 2018-08-17
Print error quark as string in the Debug impl

sdroege:


EPashkin:
@sdroege Thanks, 👍 too as it small and don't need recheck

GuillaumeGomez:
Thanks!

PR #371, opened 2018-08-28
Implement SendUniqueCell and corresponding SendUnique trait

sdroege:
This allows sending reference counted objects to different threads as
long as only a single reference to the object exists.

----

See here for an example usage: https://github.com/sdroege/gstreamer-rs/blob/e8db89464ed51148a710d54c9969e076f3aba702/examples/src/bin/pango-cairo.rs

I plan to add this as another "concurrency" mode to gir, and then most of GIO can be Send-able via this. For Pango this is unfortunately not the case as the layout has a reference to the context (and the context to the fontmap) and uses those, so the refcount of the layout itself is not sufficient for sendability.

That is, this needs very careful investigation for enabling on reference counted objects.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -203,6 +203,12 @@ pub use param_spec::ParamSpec;
 mod quark;
 pub use quark::Quark;
 
+pub mod send_unique;
+pub use send_unique::{
+    SendUniqueCell,
+    SendUnique,
+};
+
 #[cfg(feature="futures")]
 mod main_context_futures;
 #[cfg(feature="futures")]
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Seems good to me! Thanks!

sdroege:
@EPashkin what do you think?

EPashkin:
@sdroege Thanks, 👍

sdroege:
Ok, now I only need to add things to gir and send a couple of PRs :) Thanks!

PR #372, opened 2018-08-29
Add various Send/Sync impls where appropriate

sdroege:


GuillaumeGomez:
New problem with futures (again) apparently. Otherwise seems good to me, thanks!

EPashkin:
👍 too. Thanks, @sdroege

EPashkin:
Note:
Examples IMHO compiles on (I was build only one binary so not sure)
rustc 1.30.0-nightly (d41f21f11 2018-08-24)
binary: rustc
commit-hash: d41f21f11a249ad78990eb9a1ec899ec123c7136

But produce same error on
rustc 1.30.0-nightly (7061b2775 2018-08-28)
binary: rustc
commit-hash: 7061b2775782bb48c0a70d3c79ec711134396687
commit-date: 2018-08-28
host: x86_64-pc-windows-gnu

sdroege:
I would propose to remove the futures-nightly test from travis. futures 0.2 is not maintained any longer and futures 0.3 needs a nightly compiler so I wouldn't be happy to generally port thing to that.

Maybe we can have a futures 0.2 feature for stable in glib, and a futures 0.3 feature for nightly. Probably means duplicating the two `_futures.rs` files though.

Opinions?

EPashkin:
If futures-preview 0.2 don't get fixes anymore,
then IMHO we really can just ignore `future-nightly` as it currently don't add any new functional.
So I vote for removing it from testing until crate be more stable.

GuillaumeGomez:
I agree, let's comment tests for now.

PR #373, opened 2018-08-30
Remove futures-nightly feature

sdroege:
futures 0.2 is not maintained anymore and the nightly parts don't work
anymore with latest nightly toolchain

----

If I understand how to optionally depend (based on a feature flag) on futures 0.3, I'll re-add this in a different way.

Does someone know how that would look like in `Cargo.toml`?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ futures-core-preview = { version = "0.2", optional = true }
 futures-executor-preview = { version = "0.2", optional = true }
 futures-channel-preview = { version = "0.2", optional = true }
 futures-util-preview = { version = "0.2", optional = true }
-futures-stable-preview = { version = "0.2", optional = true }
 
 glib-sys = { git = "https://github.com/gtk-rs/sys" }
 gobject-sys = { git = "https://github.com/gtk-rs/sys" }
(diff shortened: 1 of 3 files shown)


EPashkin:
Maybe separating is possible with https://github.com/Kimundi/rustc-version-rs and https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies
Seems https://internals.rust-lang.org/t/setting-cfg-nightly-on-nightly-by-default/1893/10 don't reached conclusion.

GuillaumeGomez:
Instead of removing code, wouldn't it be better to comment it?

sdroege:
I removed it because for the future 0.3 version I'll simply go with a `main_context_futures_nightly.rs` or similar. There are too many differences to make it work in a reasonable way inside the same file.

EPashkin:
About loading different crates version by feature:
Checked on other crate that `[target.'cfg(feature="futures-nightly")'.dependencies]` not works
and `[target.'cfg(not(feature="futures-nightly"))'.dependencies]` always included same as `[dependencies]`
😭

GuillaumeGomez:
Ok, let's go with removal then.

GuillaumeGomez:
Restarted travis.

sdroege:
See https://github.com/rust-lang/cargo/issues/5954 for the missing cargo feature

Can we get this merged now, or ...? :)

GuillaumeGomez:
Sure, thanks!

EPashkin:
@sdroege https://github.com/rust-lang/cargo/issues/1197 active since 2015 😭 
Also somewhere I see that it better not implemented since features can be set externally (by parents) and can produce conflicting set.

PR #375, opened 2018-09-05
Move to the futures crate instead of individual crates and re-export …

sdroege:
…from here

Re-exporting makes it easier for people to mix futures 0.2 and futures
0.1/0.3 in the same application as inside Cargo.toml only one can be
specified. This allows to use e.g. tokio and GLib futures easily in the
same application.

CC @alatiera

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,13 +82,7 @@ extern crate glib_sys as ffi;
 extern crate gobject_sys as gobject_ffi;
 
 #[cfg(feature="futures")]
-extern crate futures_core;
-#[cfg(feature="futures")]
-extern crate futures_executor;
-#[cfg(feature="futures")]
-extern crate futures_channel;
-#[cfg(feature="futures")]
-extern crate futures_util;
+pub extern crate futures;
 
 use std::ffi::CStr;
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Well, at least it's shorter. :)

src/lib.rs
```diff
@@ -82,13 +82,7 @@ extern crate glib_sys as ffi;
...
 
 #[cfg(feature="futures")]
-extern crate futures_core;
-#[cfg(feature="futures")]
-extern crate futures_executor;
-#[cfg(feature="futures")]
-extern crate futures_channel;
-#[cfg(feature="futures")]
-extern crate futures_util;
+pub extern crate futures;
```
alatiera: Didn't knew you could re-export whole crates with just `pub`. This is neat!

alatiera:
This looks nice, thanks a lot for taking the time and doing it!

sdroege:
I have a WIP branch for futures 0.3 FWIW, it requires quite some changes. But that's for some future release.

sdroege:
Good to go @GuillaumeGomez

GuillaumeGomez:
Ok!

PR #384, opened 2018-10-11
Use ptr::add(i) instead of ptr::offset(i as isize)

sdroege:


sdroege:
@GuillaumeGomez all green :)

GuillaumeGomez:
Nice! I just discovered this method thanks to you. Thanks!

EPashkin:
@sdroege Thanks

PR #388, opened 2018-11-15
Add OptionGroup

GuillaumeGomez:
cc @zeenix


sdroege:
I don't think it's a good idea to generate this. It's really meant for C and in Rust we have better options for commandline parsing. Any bindings around this will never be very ergonomic :)

GuillaumeGomez:
I agree on that. :)

PR #389, opened 2018-11-16
lib: Introduce GString

philn:
This new module keeps track of CStrings to help avoid allocating new Strings in
gir's generated code.


philn:
@sdroege I think this is ready for review finally :)

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use gobject_ffi;
+use ffi as glib_ffi;
+use value::{FromValueOptional, Value};
+
+#[derive(Debug, Hash, PartialOrd, Ord)]
+pub struct CStringHolder {
+    slice: Box<CStr>,
+    ptr: *const c_char,
+    owned: bool
```
GuillaumeGomez: Nit: Missing comma.

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+        unsafe {
+            let slice = CStr::from_ptr(ptr);
+            Self { ptr, slice: Box::from(slice), owned: true }
+        }
+    }
+
+    pub fn from(s: &[u8]) -> Self {
+        let slice = match CStr::from_bytes_with_nul(s) {
+            Ok(s) => s,
+            Err(e) => panic!(e)
```
GuillaumeGomez: Nit: Missing comma.

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+use types::{StaticType, Type};
+
+use gobject_ffi;
+use ffi as glib_ffi;
+use value::{FromValueOptional, Value};
+
+#[derive(Debug, Hash, PartialOrd, Ord)]
+pub struct CStringHolder {
+    slice: Box<CStr>,
+    ptr: *const c_char,
```
sdroege: Only store the pointer here, not the CStr. The CStr can be created when needed in the functions that return the string slice

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+    owned: bool,
```
sdroege: I would only do this for owned strings, otherwise it doesn't not really make sense or does it? Do you have an example usage for a non-owned string?

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+            Err(e) => panic!(e),
+        };
+        let ptr = slice.as_ptr();
+        assert!(!ptr.is_null());
+        Self { ptr, slice: Box::from(slice), owned: false }
+    }
+
+    pub fn as_str(&self) -> &str {
+        let bytes = self.slice.to_bytes();
+        match std::str::from_utf8(bytes) {
```
sdroege: Here you could do
```
  let bytes = slice::from_raw_parts(ptr, len); // calculate the length via strlen in the constructor
  let cstr = CString::from_bytes_with_nul_unchecked(bytes);
  cstr.to_str().unwrap()
```

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+impl From<CStringHolder> for String {
+    fn from(holder: CStringHolder) -> Self {
+        String::from(holder.as_str())
+    }
+}
+
+impl From<String> for CStringHolder {
```
sdroege: This one would involve copying, not sure where it would be useful to use?

GuillaumeGomez:
Awesome! :+1:

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+    unsafe fn from_glib_none(ptr: *const c_char) -> Self {
+        assert!(!ptr.is_null());
+        CStringHolder::new(ptr)
+    }
+}
+
+impl FromGlibPtrNone<*mut i8> for CStringHolder {
+    unsafe fn from_glib_none(ptr: *mut i8) -> Self {
+        assert!(!ptr.is_null());
+        CStringHolder::new(ptr)
```
sdroege: This would then free the pointer we don't own. We can't have the `from_glib_none()` impls for this type as it owns the pointer

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+impl StaticType for CStringHolder {
+    fn static_type() -> Type {
+        unimplemented!();
```
sdroege: `String::static_type()` could be returned here

src/cstringholder.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+impl<'a> FromValueOptional<'a> for CStringHolder {
+    unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
+        from_glib_none(gobject_ffi::g_value_get_string(value.to_glib_none().0))
```
sdroege: This could use `dup_string()` and then you can take ownership of the value

src/translate.rs
```diff
@@ -1088,26 +1089,25 @@ impl FromGlibPtrNone<*const c_char> for String {
...
     }
 }
 
 impl FromGlibPtrNone<*mut c_char> for String {
     #[inline]
     unsafe fn from_glib_none(ptr: *mut c_char) -> Self {
         assert!(!ptr.is_null());
-        String::from_utf8_lossy(CStr::from_ptr(ptr).to_bytes()).into_owned()
+        let holder = cstringholder::CStringHolder::new(ptr as *const c_char);
+        String::from(holder)
```
sdroege: This is creating a copy. What is needed is that functions that return an owned string get the new type as return type

src/cstringholder.rs
philn: In the unit-tests of the gstreamer subcrate I had to manually create CStringHolders.

src/lib.rs
```diff
@@ -176,7 +176,9 @@ pub mod signal;
 pub mod source;
 pub use source::*;
 mod time_val;
+#[macro_use]
 pub mod translate;
```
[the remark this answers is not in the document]
philn: If I remove this line I get these errors:
```
   Compiling glib v0.6.0 (/home/phil/dev/rust/glib)                                                                                                                                                                
error: cannot find macro `impl_from_glib_container_as_vec_string!` in this scope                                                                                                                                   
   --> src/cstringholder.rs:154:1                                                                                                                                                                                  
    |                                                                                                                                                                                                              
154 | impl_from_glib_container_as_vec_string!(CStringHolder, *const c_char);                                                                                                                                       
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                       
    |                                                                                                                                                                                                              
    = help: have you added the `#[macro_use]` on the module/import?                                                                                                                                                
                                                                                                                                                                                                                   
error: cannot find macro `impl_from_glib_container_as_vec_string!` in this scope                                                                                                                                   
   --> src/cstringholder.rs:155:1                                                                                                                                                                                  
    |                                                                                                                                                                                                              
155 | impl_from_glib_container_as_vec_string!(CStringHolder, *mut c_char);                                                                                                                                         
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                       
    |                                                                                                                                                                                                              
    = help: have you added the `#[macro_use]` on the module/import?                                                                                                                                                
                                                                                                                                                                                                                   
error: cannot find macro `impl_from_glib_container_as_vec_string!` in this scope                                                                                                                                   
   --> src/cstringholder.rs:154:1                                                                                                                                                                                  
    |                                                                                                                                                                                                              
154 | impl_from_glib_container_as_vec_string!(CStringHolder, *const c_char);                                                                                                                                       
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                       
    |                                                                                                                                                                                                              
    = help: have you added the `#[macro_use]` on the module/import?                                                                                                                                                
                                                                                                                                                                                                                   
error: cannot find macro `impl_from_glib_container_as_vec_string!` in this scope                                                                                                                                   
   --> src/cstringholder.rs:155:1                                                                                                                                                                                  
    |                                                                                                                                                                                                              
155 | impl_from_glib_container_as_vec_string!(CStringHolder, *mut c_char);                                                                                                                                         
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                       
    |                                                                                                                                                                                                              
    = help: have you added the `#[macro_use]` on the module/import?    
```

src/translate.rs
```diff
@@ -1411,6 +1412,7 @@ macro_rules! impl_from_glib_container_as_vec_string {
 
             unsafe fn from_glib_container_num_as_vec(ptr: *mut $ffi_name, num: usize) -> Vec<Self> {
                 let res = FromGlibContainerAsVec::from_glib_none_num_as_vec(ptr, num);
+                use ffi as glib_ffi;
```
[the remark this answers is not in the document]
philn: Not needed indeed, will remove these changes :)

src/cstringholder.rs
philn: Yeah let's remove it, doesn't seem useful indeed :)

src/cstringholder.rs
philn: Hum the from_glib_none impl is needed by impl_from_glib_container_as_vec_string

sdroege: Hrm, `unimplemented!()` then. This should never be called :)

src/translate.rs
philn: Yeah, I forgot to regenerate the bindings :)

src/lib.rs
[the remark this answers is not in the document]
EPashkin: Sorry, I mistakes, it [macro_export] in translate.rs unneeded.

EPashkin:
Note: building with rust 1.28.0 produce error:
```
error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
  --> D:/eap/rust/0/glib\src\cstringholder.rs:44:15
   |
44 |         match std::str::from_utf8(bytes) {
   |               ^^^
```

philn:
Hopefully this code won't be part of the next patch iteration :)

philn:
Here's my WIP gstreamer fork adding support for CStringHolder too, https://gitlab.freedesktop.org/philn/gstreamer-rs/commits/master

philn:
https://github.com/philn/pango/commit/2aa6ff7d644b592fa6db8ed268c9192bc3f22a4e

cgwalters:
I only glanced at this but it looks like there's some overlap with https://crates.io/crates/c_utf8 ?  I started using that crate in https://github.com/projectatomic/rpm-ostree/pull/1588

philn:
This was briefly discussed on IRC, the outcome was that for this specific task we would prefer to avoid adding a new external dependency.

philn:
PR updated. I suspect CI might fail again though, will keep an eye on it :)

EPashkin:
Maybe in not_owned case use `CString` field as internal storage?

src/cstringholder.rs
```diff
@@ -0,0 +1,293 @@
...
+        };
+        let ptr = slice.as_ptr();
+        assert!(!ptr.is_null());
+        Self { ptr, length: s.len() - 1, owned: false }
+    }
+
+    pub fn from_string(s: std::string::String) -> Self {
+        let ptr = s.as_ptr();
+        let length = s.len();
+        Self { ptr: ptr as *const c_char, length, owned: false }
```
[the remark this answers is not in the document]
philn: I added this function for cases where a signal handler returns a CStringHolder, like GtkScale::format-value. It's not right indeed :( I wonder if for those cases we should simply don't use CStringHolder.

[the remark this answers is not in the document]
EPashkin: `CStringHolder` IMHO can used in converting C to Rust, otherwise better use `String`, and its `.to_glib_xxx()`,
so better don't use it in signal returns

EPashkin:
@philn Reason to add `CStringHolder` is minimize memory copy when receiving C string?

src/cstringholder.rs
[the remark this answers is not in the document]
sdroege: This code is unsafe: `s` will be freed at the end of the function and the pointer stored in `Self` will be pointing at freed memory.

src/cstringholder.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    pub fn from(s: &[u8]) -> Self {
+        let slice = match CStr::from_bytes_with_nul(s) {
+            Ok(s) => s,
+            Err(e) => panic!(e),
+        };
+        let ptr = slice.as_ptr();
+        assert!(!ptr.is_null());
+        Self { ptr, length: s.len() - 1, owned: false }
```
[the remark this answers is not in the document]
sdroege: `CStringHolder` can only really be implemented in a useful way of `from_glib_full`

sdroege:
> Maybe in not_owned case use `CString` field as internal storage?

`CString` would involve a copy, `CStr` would be unsafe. Strings that are `transfer none` we must always directly copy to new memory.

src/cstringholder.rs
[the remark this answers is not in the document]
EPashkin: IMHO `from_glib_borrow` also can be safely implemented for signal string parameters.
Problem with string, returned from function as `from_glib_none`, I don't see safe way to remove copy.

EPashkin:
Assuming that `CStringHolder` need to `minimize memory copy when receiving C string`
IMHO it need implemented like `AnyBox`: https://github.com/gtk-rs/glib/blob/master/src/boxed.rs#L259-L263 with 3 separated cases:
for `from_glib_full` it store pointer and drop it,
for `from_glib_borrow` it store pointer and not drop it,
for `from_glib_none` it store `String` or `CString` with copy data (this is bad because too many string returned that way).

EPashkin:
Name `CStringHolder` IMHO misleading, something like `GStr(ing)?Holder` is better.

src/cstringholder.rs
```diff
@@ -0,0 +1,299 @@
...
+
+#[derive(Hash)]
+pub struct CStringHolder {
+    ptr: *const c_char,
+    length: usize,
+    owned: bool,
+}
+
+impl CStringHolder {
+    pub fn new(ptr: *const c_char) -> Self {
```
tmiasko: This should be marked as unsafe since it dereferences ptr.

sdroege:
What @EPashkin said in https://github.com/gtk-rs/glib/pull/389#issuecomment-441459714 seems like the best way forward here. I also agree that the name is a bit suboptimal. I would've called it `GString`.

philn:
Are you fine with that name (GString) @EPashkin ? I don't plan to rename this more than one time.

EPashkin:
@philn I fine with `GString` ;)
@GuillaumeGomez And you?

GuillaumeGomez:
Fine for me too.

sdroege:
`GString` has only the disadvantage that this is the name of `glib::String` in C. But I'm ok with that

src/cstringholder.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+impl<'a> ToGlibPtr<'a, *const c_char> for CStringHolder {
+    type Storage = CString;
+
+    #[inline]
+    fn to_glib_none(&'a self) -> Stash<'a, *const c_char, Self> {
+        unsafe {
+            let tmp = CString::from_raw(self.ptr as *mut i8);
+            Stash(tmp.as_ptr(), tmp)
```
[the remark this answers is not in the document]
philn: I don't understand what you mean there :) The stash would expect a `CStringHolder` then, not a reference. Also I'm not sure how the storage can be changed depending on `owned` value.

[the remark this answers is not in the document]
EPashkin: You right, storage can't changed depending on `owned` value, sorry.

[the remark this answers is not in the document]
EPashkin: Actually it can be something like `Result<&CStringHolder, CString>` or tupple with 2 options, but not sure about `&CStringHoder` part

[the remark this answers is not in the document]
philn: Seems to work with a tuple

philn:
I don't know if anyone noticed the PR is updated or if everyone is already tired of reviewing this PR :)

GuillaumeGomez:
The second of course! And I didn't receive notifications before you commented. So when you update, please comment as well. :D

src/value.rs
```diff
@@ -817,16 +818,16 @@ impl SetValueOptional for str {
...
+        Some(<Vec<GString> as FromValue>::from_value(value))
     }
 }
 
-impl<'a> FromValue<'a> for Vec<String> {
+impl<'a> FromValue<'a> for Vec<GString> {
     unsafe fn from_value(value: &'a Value) -> Self {
         let ptr = gobject_ffi::g_value_get_boxed(value.to_glib_none().0) as *const *const c_char;
-        FromGlibPtrContainer::from_glib_none(ptr)
+        GString::from_ptr_container(ptr)
```
EPashkin: There really needed to remove implementations for `Vec<String>` ?

src/translate.rs
```diff
@@ -1759,7 +1765,7 @@ mod tests {
         let ptr: *mut *mut c_char = stash.0;
         let ptr_copy = unsafe { glib_ffi::g_strdupv(ptr) };
 
-        let actual: Vec<String> = unsafe { FromGlibPtrContainer::from_glib_full(ptr_copy) };
+        let actual: Vec<GString> = unsafe { FromGlibPtrContainer::from_glib_full(ptr_copy) };
```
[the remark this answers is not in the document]
EPashkin: *gstring_array

EPashkin:
Looks good, but I still prefer use `from_glib_full` instead `GString::new` etc. in manual code.

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+
+use gobject_ffi;
+use ffi as glib_ffi;
+use value::{FromValueOptional, Value};
+
+#[derive(Hash)]
+pub struct GString {
+    s: Option<CString>,
+    ptr: *const c_char,
+    length: usize
```
EPashkin: Maybe better use `enum` instead `struct` ?
I do not insist.

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+            }
+            res
+        }
+    }
+
+    pub fn as_str(&self) -> &str {
+        let cstr = match self.s {
+            Some(ref s) => s.as_c_str(),
+            None => unsafe {
+                assert!(!self.ptr.is_null());
```
EPashkin: Is was checked in `new` already

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+                assert!(self.length > 0);
```
EPashkin: No empty string?
Anyway it IMHO need be in `new` too

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+                CStr::from_bytes_with_nul_unchecked(bytes)
+            }
+        };
+        cstr.to_str().unwrap()
+    }
+
+    pub fn to_cstring(&self) -> CString {
+        if let Some(ref s) = self.s {
+            let bytes = s.as_bytes_with_nul();
+            unsafe { CString::from_vec_unchecked(bytes.to_vec()) }
```
EPashkin: Why not use `.clone()`?

EPashkin:
@philn Hm, seems you lost usage in trampolines aka `from_glib_borrow`.

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    #[test]
+    fn test_holder() {
+        let data = CString::new("foo").unwrap();
+        let ptr = data.into_raw();
+
+        unsafe {
+            let holder = GString::new(ptr);
+            assert_eq!(holder.as_str(), "foo");
+            let foo: Box<str> = holder.into();
```
EPashkin: If `impl From<GString> for Box<str>` called on "owned ptr" case, then seems it needed `mem::forget` too,
maybe `impl From<GString> for String` too, but not sure

EPashkin:
Found other problem:
We have https://gtk-rs.org/docs/gtk/trait.ScaleExt.html#tymethod.connect_format_value:
signal returning string (Ex. https://github.com/gtk-rs/examples/blob/master/src/bin/gtktest.rs#L78-L81),
with https://github.com/gtk-rs/gir/pull/666 it will be changed to return `GString`,
so closure need be more complex, as seems no conversion from nonstatic `String` to `GString`.

Also I don't see way to prevent this change in gir as it

philn:
I have a gir change for signal trampolines returning String...

EPashkin:
@sdroege, @GuillaumeGomez, @philn  Maybe it too late,
but IMHO worth of using this class for all function string results is arguable,
as it add extra problem to users (usually need extra call) and give benefit (no memory copy) in about half usages (Ex. in gtk-3.0.gir: 112none, const char* vs 60full, char*).
`GString` helps as trampoline parameters for sure, I found only 12 cases of it in gtk.

So we better decide use it/not/in which cases before @philn do more works on it.

For me pros slightly less that cons but feels almost equals.

sdroege:
I'll review this another time tomorrow :) sorry for the delay

philn:
@EPashkin please look at the broader picture. The GTK bindings are not the only users of the GLib bindings, consider the GStreamer bindings for instance.

EPashkin:
@philn Problem that I don't know which benefit from this class for GStreamer too.
Can you tell more?

sdroege:
> but IMHO worth of using this class for all function string results is arguable,
> as it add extra problem to users (usually need extra call) and give benefit (no memory copy)

There will always be a memory copy in case of `transfer none` return values, that's correct. We only prevent copies in `transfer full` cases (and in callbacks as those borrow the strings).

What do you mean with users have to call functions? As long as they use it in places where a `&str` is required, nothing should be needed (as this here is `Deref<Target=str>`). If they need a proper `String` then `into()` or similar is required, that's correct, but that seems like an exceptional case.

src/lib.rs
```diff
@@ -183,7 +183,9 @@ pub mod signal;
...
+pub mod gstring;
```
sdroege:
```rust
mod gstring;
pub use gstring::GString;
```
seems nicer for users

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+    }
+}
+
+impl fmt::Display for GString {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+impl std::borrow::Borrow<str> for GString {
```
sdroege: This should also implemented `Deref<Target=str>`

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+    }
+}
+
+impl PartialEq<str> for GString {
+    fn eq(&self, other: &str) -> bool {
+        self.as_str() == other
+    }
+}
+
+impl PartialEq<&'static str> for GString {
```
sdroege: This one seems unneeded, but it would probably make sense to also implement `PartialOrd` for `str` and `String` (and implement them the other way around too)

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+}
+
+impl From<GString> for String {
+    #[inline]
+    fn from(holder: GString) -> Self {
+        String::from(holder.as_str())
+    }
+}
+
+impl From<GString> for Box<str> {
```
sdroege: We also want these I guess

```rust
impl From<String> for GString
impl From<Box<str>> for GString
```

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+
+impl StaticType for Vec<GString> {
+    fn static_type() -> Type {
+        unsafe {
+            from_glib(glib_ffi::g_strv_get_type())
+        }
+    }
+}
+
+impl<'a> FromValueOptional<'a> for GString {
```
sdroege: We don't need the other way around here, `SetValue` / `SetValueOptional`?

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+}
+
+impl GString {
+    pub unsafe fn new(ptr: *const c_char) -> Self {
+        assert!(!ptr.is_null());
+        let length = libc::strlen(ptr);
+        Self { s: None, ptr, length }
+    }
+
+    pub fn from(s: &[u8]) -> Self {
```
sdroege: `fn from` and `fn from_cstring` should go away and instead become `From<_>` impls. I think clippy also would warn about this

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+                assert!(!self.ptr.is_null());
+                assert!(self.length > 0);
+                let bytes = std::slice::from_raw_parts(self.ptr as *const u8, self.length + 1);
+                CStr::from_bytes_with_nul_unchecked(bytes)
+            }
+        };
+        cstr.to_str().unwrap()
+    }
+
+    pub fn to_cstring(&self) -> CString {
```
sdroege: And maybe also a `to_string()`?

src/value.rs
philn: I can leave it :)

src/gstring.rs
philn: I'm sorry, I'm not sure to understand what you mean, can you please clarify?

src/gstring.rs
```diff
@@ -0,0 +1,333 @@
...
+}
+
+impl<'a> ToGlibPtr<'a, *const c_char> for GString {
+    type Storage = Option<CString>;
+
+    #[inline]
+    fn to_glib_none(&'a self) -> Stash<'a, *const c_char, Self> {
+        let cstring = self.to_cstring();
+        Stash(cstring.as_ptr(), Some(cstring))
+    }
```
EPashkin: @philn, @sdroege Maybe this implementation can be simplified:
```rust
impl<'a> ToGlibPtr<'a, *const c_char> for GString {
    type Storage = &'a Self;
    #[inline]
    fn to_glib_none(&'a self) -> Stash<'a, *const c_char, Self> {
        Stash(self.as_ptr() as *const _, self)
    }
```

philn: Yes!

src/gstring.rs
sdroege: You probably want `From` impls in both directions

src/gstring.rs
philn: It's still needed for code like:
```
error[E0277]: can't compare `glib::GString` with `&str`                                                                                                                                                            
   --> gstreamer-pbutils/src/encoding_profile.rs:521:9                                                                                                                                                             
    |                                                                                                                                                                                                              
521 |         assert_eq!(audio_profile.get_name().unwrap(), AUDIO_PROFILE_NAME);                                                                                                                                   
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `glib::GString == &str`                                                                                     
    |                                                                                                                                                                                                              
    = help: the trait `std::cmp::PartialEq<&str>` is not implemented for `glib::GString`                                                                             
```

sdroege: Strange, I would've expected the impl just above to cover this case. But in any case, instead of static this should be possible with any lifetime. `impl<'a> PartialEq<&'a str> ....`

philn:
Ok, the PR is updated. I'm not sure I addressed all the comments but I did my best 😂

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+        cstr.to_owned().into()
+    }
+}
+
+impl FromGlibPtrBorrow<*const c_char> for GString {
+    #[inline]
+    unsafe fn from_glib_borrow(ptr: *const c_char) -> Self {
+        GString::new(ptr)
+    }
+}
```
EPashkin: Borrowed `ptr` can't be `g_free`d

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+
+impl std::borrow::Borrow<str> for GString {
+    fn borrow(&self) -> &str {
+        self.as_str()
+    }
+}
+
+impl Ord for GString {
+    fn cmp(&self, other: &GString) -> Ordering {
+        self.as_str().bytes().cmp(other.as_str().bytes())
```
EPashkin: Just comparing .as_str() not work?

src/gstring.rs
EPashkin: Missed `let ptr_copy = glib_ffi::g_strdup(ptr);` in this test

src/lib.rs
EPashkin: Still `pub mod`

src/lib.rs
```diff
@@ -183,7 +183,10 @@ pub mod signal;
...
```
sdroege: You can keep the above `mod gstring;` btw if there's nothing that ever would have to be imported from there directly.

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+use value::{FromValueOptional, Value, SetValueOptional, SetValue};
+
+pub struct GString {
+    s: Option<CString>,
+    ptr: *const c_char,
+    length: usize
+}
+
+impl GString {
+    pub unsafe fn new(ptr: *const c_char) -> Self {
```
sdroege: This should probably be `*mut c_char` as we take ownership of the pointer that is passed in

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+    #[inline]
+    fn from(s: Box<str>) -> Self {
+        s.as_bytes().to_vec().into()
+    }
+}
+
+impl From<&'static str> for GString {
+    #[inline]
+    fn from(s: &'static str) -> Self {
+        s.as_bytes().to_vec().into()
```
sdroege: This copies, which we maybe don't want?

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+    fn from(s: GString) -> Self {
+        std::mem::forget(s.ptr);
+        Box::from(s.as_str())
+    }
+}
+
+impl From<String> for GString {
+    #[inline]
+    fn from(s: String) -> Self {
+        GString::from(s.as_bytes().to_vec())
```
sdroege: This copies, which we maybe don't want?

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+}
+
+impl From<Box<str>> for GString {
+    #[inline]
+    fn from(s: Box<str>) -> Self {
+        s.as_bytes().to_vec().into()
```
sdroege: This copies, which we maybe don't want?

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+    #[inline]
+    fn from(s: &'static str) -> Self {
+        s.as_bytes().to_vec().into()
+    }
+}
+
+impl From<Vec<u8>> for GString {
+    #[inline]
+    fn from(s: Vec<u8>) -> Self {
+        let cstr = unsafe { CString::from_vec_unchecked(s) };
```
sdroege: This seems unsafe. You could create strings with this that are not NUL-terminated

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+        let cstring = CString::from(c);
+        Self::from(cstring)
+    }
+}
+
+impl FromGlibPtrFull<*const c_char> for GString {
+    #[inline]
+    unsafe fn from_glib_full(ptr: *const c_char) -> Self {
+        let dst = std::ptr::NonNull::dangling();
+        std::ptr::write(dst.as_ptr(), std::ptr::read(ptr));
```
sdroege: `let dst = std::ptr::NonNull::new(ptr).unwrap()`?

sdroege: Also why the indirection via `NonNull` if you store it as a raw pointer anyway?

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+    unsafe fn from_glib_none(ptr: *mut i8) -> Self {
+        let cstr = CStr::from_ptr(ptr);
+        cstr.to_owned().into()
+    }
+}
+
+impl FromGlibPtrBorrow<*const c_char> for GString {
+    #[inline]
+    unsafe fn from_glib_borrow(ptr: *const c_char) -> Self {
+        GString::new(ptr)
```
sdroege: You need to mark this somehow as borrowed in the `GString` so that the `Drop` impl is not freeing the pointer. That's why an enum would be best for the representation of it

philn: Yes I started moving into that direction...

src/gstring.rs
```diff
@@ -0,0 +1,440 @@
...
+        if num == 0 || ptr.is_null() {
+            return Vec::new();
+        }
+
+        let mut res = Vec::with_capacity(num);
+        for i in 0..num {
+            let len = libc::strlen(*ptr.add(i));
+            let bytes = std::slice::from_raw_parts(*ptr.add(i) as *const u8, len + 1);
+            let cstr = CStr::from_bytes_with_nul_unchecked(bytes);
+            res.push(Self::from(cstr.to_owned()));
```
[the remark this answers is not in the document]
philn: Actually I think this fn is no longer needed :)

philn:
PR updated!

src/gstring.rs
```diff
@@ -0,0 +1,450 @@
...
+    #[inline]
+    fn from(s: GString) -> Self {
+        Box::from(s.as_str())
+    }
+}
+
+impl From<String> for GString {
+    #[inline]
+    fn from(s: String) -> Self {
+        GString::Borrowed(s.as_ptr() as *const c_char, s.len())
```
EPashkin: Also worry about use dropped String data.

src/gstring.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+}
+
+impl From<Box<str>> for GString {
+    #[inline]
+    fn from(s: Box<str>) -> Self {
+        GString::Borrowed(s.as_ptr() as *const c_char, s.len())
```
EPashkin: Also worry about use dropped Box<str> data.

src/gstring.rs
```diff
@@ -0,0 +1,450 @@
...
+        let res = from_glib_none(ptr);
+        glib_ffi::g_free(ptr as *mut _);
+        res
+    }
+}
+
+impl FromGlibPtrFull<*mut i8> for GString {
+    #[inline]
+    unsafe fn from_glib_full(ptr: *mut i8) -> Self {
+        let res = from_glib_none(ptr);
```
EPashkin: `from_glib_none` implementation involve copy, IMHO better use `GString::new` directly

src/gstring.rs
philn: Not sure we can avoid the copy here.

src/gstring.rs
philn: Same :(

src/gstring.rs
```diff
@@ -0,0 +1,450 @@
...
+    #[inline]
+    fn from(s: GString) -> Self {
+        String::from(s.as_str())
+    }
+}
+
+impl From<GString> for Box<str> {
+    #[inline]
+    fn from(s: GString) -> Self {
+        Box::from(s.as_str())
```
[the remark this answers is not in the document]
philn: Now I convert to a String and then Boxed str

sdroege:
I'll give this another review tomorrow :) Thanks a lot for your patience, @philn

src/gstring.rs
```diff
@@ -0,0 +1,445 @@
...
+use types::{StaticType, Type};
+use libc;
+
+use gobject_ffi;
+use ffi as glib_ffi;
+use value::{FromValueOptional, Value, SetValueOptional, SetValue};
+
+#[derive(Debug)]
+pub enum GString {
+    ForeignOwned(CString),
```
sdroege: Where foreign means Rust-allocated? Ok :)

src/gstring.rs
sdroege: Yes it has to be copied unfortunately as we need to get the `\0` at the end. Nothing we can do about that :(

src/gstring.rs
```diff
@@ -0,0 +1,445 @@
...
+}
+
+impl From<&'static str> for GString {
+    #[inline]
+    fn from(s: &'static str) -> Self {
+        GString::Borrowed(s.as_ptr() as *const c_char, s.len())
+    }
+}
+
+impl From<Vec<u8>> for GString {
```
sdroege: Can this be made generic as `impl<T: Into<Vec<u8>>> From<T> for GString {}` or does that break things?

src/gstring.rs
```diff
@@ -0,0 +1,445 @@
...
+        let cstr = CStr::from_ptr(ptr);
+        cstr.into()
+    }
+}
+
+impl FromGlibPtrNone<*mut u8> for GString {
+    #[inline]
+    unsafe fn from_glib_none(ptr: *mut u8) -> Self {
+        let cstr = CStr::from_ptr(ptr as *mut i8);
+        cstr.to_owned().into()
```
sdroege: Why not just `cstr.into()`?

src/gstring.rs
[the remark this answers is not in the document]
sdroege: :+1:

sdroege:
Looks good to me now, just some cosmetic things above.

EPashkin:
@philn All nightly gtk3.4 fails due missing update gir submodule,
also gir version in version.txt better be "c5498ac"

There 2 legit error in 1.28.0 build:
```
error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
  --> src/gstring.rs:72:6
   |
72 | impl std::borrow::Borrow<str> for GString {
   |      ^^^
error[E0106]: missing lifetime specifier
   --> src/gstring.rs:229:11
    |
229 | impl From<&CStr> for GString {
    |           ^ expected lifetime parameter
```

src/gstring.rs
philn: I tried this but then the `FromGlibPtrNone` impls can't rely on `CStr`:
```
error[E0277]: the trait bound `std::vec::Vec<u8>: std::convert::From<&std::ffi::CStr>` is not satisfied                                                                                                            
   --> src/gstring.rs:270:14                                                                                                                                                                                       
    |                                                                                                                                                                                                              
270 |         cstr.into()                                                                                                                                                                                          
    |              ^^^^ the trait `std::convert::From<&std::ffi::CStr>` is not implemented for `std::vec::Vec<u8>`                                                                                                 
    |                                                                                                                                                                                                              
              <std::vec::Vec<T> as std::convert::From<std::boxed::Box<[T]>>>                                                                                                                                       
              <std::vec::Vec<u8> as std::convert::From<&'a str>>                                                                                                                                                   
              <std::vec::Vec<u8> as std::convert::From<std::ffi::CString>>                                                                                                                                         
              <std::vec::Vec<u8> as std::convert::From<std::string::String>>                                                                                                                                       
            and 5 others                                                                                                                                                                                           
    = note: required because of the requirements on the impl of `std::convert::Into<std::vec::Vec<u8>>` for `&std::ffi::CStr`                                                                                      
    = note: required because of the requirements on the impl of `std::convert::From<&std::ffi::CStr>` for `gstring::GString`                                                                                       
    = note: required because of the requirements on the impl of `std::convert::Into<gstring::GString>` for `&std::ffi::CStr`        
```

philn:
@EPashkin OK! I had to force a clean gir build to have the right version.txt :)

src/gstring.rs
philn: Maybe `Native`?

src/gstring.rs
philn: I could use `CString::from_raw` but its documentation doesn't recommend it :)

sdroege: Ok then :)

philn:
OK I think CI will pass this time!

GuillaumeGomez:
Thanks a lot!

PR #393, opened 2018-11-20
Add a second constructor to ObjectSubclass that provides the class st…

sdroege:
…ruct

This one is optional but can be useful in special cases where the class
struct is necessary for getting additional context about the type that
is being created.


sdroege:
@GuillaumeGomez All green, please merge :)

GuillaumeGomez:
Thanks! :)

PR #400, opened 2018-11-28
Rename glib_boxed_get_type! to glib_boxed_type! for consistency

sdroege:
We might want to add more things at a later time than the get_type()
function, and it's also called like the trait in case of ObjectSubclass


EPashkin:
👍

sdroege:
@GuillaumeGomez all green :)

GuillaumeGomez:
Thanks!

PR #401, opened 2018-11-28
Require fewer trait bounds for WeakRef and the ObjectExt trait

sdroege:
This is a very first part of https://github.com/gtk-rs/gir/issues/670


src/object.rs
```diff
@@ -1118,14 +1117,22 @@ impl<T: IsA<Object> + SetValue> ObjectExt for T {
...
+                gobject_ffi::g_value_init(v.to_glib_none_mut().0, self.get_type().to_glib());
+                gobject_ffi::g_value_set_object(v.to_glib_none_mut().0, self.to_glib_none().0);
+                v
+            };
             let args = if args.len() < 10 {
-                for (i, arg) in iter::once(&(self as &ToValue)).chain(args).enumerate() {
-                    s_args[i] = arg.to_value();
+                s_args[0] = self_v;
+                for (i, arg) in args.iter().enumerate() {
+                    s_args[i+1] = arg.to_value();
```
GuillaumeGomez: Shouldn't calling `.all()` instead of looping on it be more rusty?

GuillaumeGomez: I think it'd work as follows:

```rust
args.iter().enumerate().for_each(|(i, arg)| { s_args[i+1] = arg.to_value(); });
```

(Yes, I know, I used `for_each` instead haha)

sdroege: `for_each` is not really idiomatic

GuillaumeGomez: It's just nitpicking. Either way is fine. Why isn't it idiomatic btw?

sdroege: For imperative code (which `for_each` is), `for` loops are generally preferred

GuillaumeGomez: Ok, good to know! Then it's all good for me. I'll merge once CIs passed.

src/object.rs
```diff
@@ -1228,9 +1235,9 @@ unsafe impl IsClassFor for ObjectClass {
 unsafe impl Send for ObjectClass {}
 unsafe impl Sync for ObjectClass {}
 
-pub struct WeakRef<T: IsA<Object> + ?Sized>(Box<gobject_ffi::GWeakRef>, PhantomData<*const T>);
+pub struct WeakRef<T: IsA<Object>>(Box<gobject_ffi::GWeakRef>, PhantomData<*const T>);
```
EPashkin: This change really "require fewer" ? 😉
Or unsized types better not used now?

sdroege: Indeed that one is not really fewer :) It can't really work right now for unsized/dynamically sized types (e.g. `IsA` is not object-safe, and various other issues). I have fixing that on my list but that's still a longer way.

sdroege: Updated the commit message with that

EPashkin:
@sdroege Thanks, 👍

PR #402, opened 2018-11-29
Replace unwrap calls with "more explicit" expect

GuillaumeGomez:
Part of https://github.com/gtk-rs/gtk/issues/743.

cc @sdroege @EPashkin

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -329,7 +329,7 @@ impl<'a> ToGlibPtr<'a, *const c_char> for str {
 
     #[inline]
     fn to_glib_none(&'a self) -> Stash<'a, *const c_char, Self> {
-        let tmp = CString::new(self).unwrap();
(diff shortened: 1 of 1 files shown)


EPashkin:
@GuillaumeGomez  Thanks, 👍

sdroege:
Instead of "invalid string" you could also be more specific as this *only* happens if there's a "\0" somewhere in the string before the end. Otherwise seems fine to me

GuillaumeGomez:
Are you sure this is the only case where it could fail?

sdroege:
Yes: https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new

It's part of the API, it can't possibly fail for anything but a `NulError`:
```rust
pub fn new<T: Into<Vec<u8>>>(t: T) -> Result<CString, NulError>
```

GuillaumeGomez:
Ok then! I'll update the sentence when I can.

GuillaumeGomez:
Updated.

sdroege:
:+1:

PR #403, opened 2018-12-04
Add function to get the implementation from an instance

sdroege:
This allows to get the implementation safely without going via the
instance struct first by directly calling it on the wrapper type.


src/subclass/types.rs
```diff
@@ -271,6 +272,23 @@ pub trait ObjectSubclass: ObjectImpl + Sized + 'static {
...
+    /// Returns the implementation from an instance.
+    ///
+    /// Panics if called on an object of the wrong type.
+    fn get_impl_from_instance(obj: &Self::ParentType) -> &Self {
+        unsafe {
+            let data = Self::type_data();
+            let type_ = data.as_ref().get_type();
+            assert_ne!(type_, Type::Invalid);
+
+            assert!(obj.get_type().is_a(&type_));
```
GuillaumeGomez: Wouldn't it be better/simpler to use one `assert` instead of two? Like:

```rust
assert!(type_ != Type::Invalid && obj.get_type().is_a(&type_));
```

src/subclass/types.rs
```diff
@@ -16,6 +16,7 @@ use translate::*;
 use wrapper::Wrapper;
 use {IsA, IsClassFor, Object, StaticType, Type};
 
+use super::super::object::ObjectExt;
```
GuillaumeGomez: That's the first time I see `super::super`. Just asking but I guess the answer is no: is it possible to not import it this way? It seems very werid?

sdroege: You could also do `use ::object::ObjectExt`. Do you prefer that?

src/subclass/types.rs
sdroege: It's two different things we're asserting here: 1) that the type is already properly registered, 2) that the object passed in is of the correct type.

Making it a single assertion would make it a guessing game which of the two cases it is if the assertion ever fails.

GuillaumeGomez: That's a good point. Nevermind then!

src/subclass/types.rs
sdroege: Changed that

sdroege: New version pushed that also changes the name of the function (to be shorter) and makes it more generic

GuillaumeGomez: All better then! Thanks for doing the change. :)

GuillaumeGomez:
Looks good to me!

cc @EPashkin

sdroege:
@GuillaumeGomez all green :)

GuillaumeGomez:
Just waiting for @EPashkin to take a look first in case I missed something.

PR #410, opened 2018-12-14
variant: Implement Value traits manually

sdroege:
This is not a boxed type and we have to use specific functions for
storing/retrieving from a Value

---

CC @jsparber


GuillaumeGomez:
👍

EPashkin:
@sdroege Thanks for noticing this, 👍

PR #411, opened 2018-12-16
gstring: Specify a lifetime for the From<str> impl

philn:
Needed for https://github.com/gtk-rs/gio/pull/177

### DIFF
--- a/src/gstring.rs
+++ b/src/gstring.rs
@@ -206,10 +206,10 @@ impl From<Box<str>> for GString {
     }
 }
 
-impl From<&'static str> for GString {
+impl<'a> From<&'a str> for GString {
     #[inline]
-    fn from(s: &'static str) -> Self {
-        GString::Borrowed(s.as_ptr() as *const c_char, s.len())
+    fn from(s: &'a str) -> Self {
+        s.as_bytes().to_vec().into()
     }
 }
 


src/gstring.rs
```diff
@@ -206,9 +206,9 @@ impl From<Box<str>> for GString {
...
 }
 
-impl From<&'static str> for GString {
+impl<'a> From<&'a str> for GString {
     #[inline]
-    fn from(s: &'static str) -> Self {
+    fn from(s: &'a str) -> Self {
         GString::Borrowed(s.as_ptr() as *const c_char, s.len())
     }
 }
```
EPashkin: This is wrong, as we outlive `'a` for sure.
Can you add other From with `.to_owned()` conversion instead, if rust allow?

philn: In addition to `impl From<&'static str> for GString`? The conflicting impl isn't allowed.

src/gstring.rs
```diff
@@ -206,9 +206,9 @@ impl From<Box<str>> for GString {
...
 
-impl From<&'static str> for GString {
+impl<'a> From<&'a str> for GString {
```
sdroege: You need to add a lifetime to GString too then,otherwise this is unsafe

src/gstring.rs
sdroege: How did that work without copying before anyway. We just missed a mistake here before :) We would have to copy it anyway because we need to add a `'\0'` at the end.

src/gstring.rs
philn: Hum this would need another gir codegen round it seems

sdroege: Yes, but as we need to copy anyway (to add the `NUL` terminator), there's not much point in this. The previous code was wrong, it omitted the `NUL` terminator so `to_glib_none()` would be broken.

src/gstring.rs
philn: Alright, PR updated accordingly (I think 😅)

sdroege:
Looks good to me

GuillaumeGomez:
Looks good to me as well! What about you @EPashkin ?

EPashkin:
@philn Thanks
@GuillaumeGomez 👍

PR #419, opened 2018-12-31
BoolError: allow owning the message

fengalin:
`BoolError` can currently only be built from a `&'static str`.
This prevents the user from dynamically building the message.

Thanks to @sdroege for [suggesting to use `Cow`](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/merge_requests/196#note_98010).

It's also an opportunity to keep track of the location where the error was built.

This causes breaking changes in some callers but I think previous implementation was not future-proof, so the change should be done IMHO.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,6 +6,7 @@
 
 use std::ffi::CStr;
 use Quark;
+use std::borrow::Cow;
 use std::error;
 use std::fmt;
 use std::str;
@@ -132,13 +133,78 @@ pub trait ErrorDomain: Copy {
 }
 
 /// Generic error used for functions that fail without any further information
+#[macro_export]
+macro_rules! glib_bool_error(
+// Plain strings
+    ($msg:expr) =>  {
+        $crate::BoolError::new($msg, file!(), module_path!(), line!())
+    };
+
+// Format strings
+    ($($msg:tt)*) =>  { {
+        $crate::BoolError::new(format!($($msg)*), file!(), module_path!(), line!())
+    }};
+);
+
+#[macro_export]
+macro_rules! glib_result_from_gboolean(
+// Plain strings
+    ($ffi_bool:expr, $msg:expr) =>  {
+        $crate::BoolError::from_glib($ffi_bool, $msg, file!(), module_path!(), line!())
+    };
+
+// Format strings
+    ($ffi_bool:expr, $($msg:tt)*) =>  { {
+        $crate::BoolError::from_glib(
+            $ffi_bool,
+            format!($($msg)*),
+            file!(),
+            module_path!(),
+            line!(),
+        )
+    }};
+);
+
 #[derive(Debug)]
-pub struct BoolError(pub &'static str);
+pub struct BoolError {
+    pub message: Cow<'static, str>,
+    #[doc(hidden)]
+    pub filename: &'static str,
+    #[doc(hidden)]
+    pub function: &'static str,
+    #[doc(hidden)]
+    pub line: u32,
+}
 
 impl BoolError {
-    pub fn from_glib(b: glib_ffi::gboolean, s: &'static str) -> Result<(), Self> {
+    pub fn new<Msg: Into<Cow<'static, str>>>(
+        message: Msg,
+        filename: &'static str,
+        function: &'static str,
+        line: u32,
+    ) -> Self {
+        BoolError {
+            message: message.into(),
+            filename,
+            function,
+            line,
+        }
+    }
+
+    pub fn from_glib<Msg: Into<Cow<'static, str>>>(
+        b: glib_ffi::gboolean,
+        message: Msg,
+        filename: &'static str,
+        function: &'static str,
+        line: u32,
+    ) -> Result<(), Self> {
         match b {
-            glib_ffi::GFALSE => Err(BoolError(s)),
+            glib_ffi::GFALSE => Err(BoolError::new(
+                message,
+                filename,
+                function,
+                line,
+            )),
             _ => Ok(()),
         }
     }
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -133,25 +134,59 @@ pub trait ErrorDomain: Copy {
 
 /// Generic error used for functions that fail without any further information
 #[derive(Debug)]
-pub struct BoolError(pub &'static str);
+pub struct BoolError(Cow<'static, str>);
```
sdroege: I would still keep the field public

src/error.rs
```diff
@@ -133,25 +134,59 @@ pub trait ErrorDomain: Copy {
...
 
 impl BoolError {
     pub fn from_glib(b: glib_ffi::gboolean, s: &'static str) -> Result<(), Self> {
```
sdroege: By using the right trait, this should be possible to have with a single function. If you pass in a owned string, it will use that variant, otherwise the borrowed variant

src/error.rs
```diff
@@ -133,25 +134,66 @@ pub trait ErrorDomain: Copy {
 
 /// Generic error used for functions that fail without any further information
 #[derive(Debug)]
-pub struct BoolError(pub &'static str);
+pub struct BoolError(pub Cow<'static, str>);
```
sdroege: Maybe while we extend this, we could also add a field for `file!()`, `module_path!()` and `line!()` and make a constructor-macro that fills that in?

sdroege: @GuillaumeGomez @EPashkin opinions?

fengalin: Good idea!

fengalin: See the new commit. I'm not convinced by the macro name: `glib_result_from_glib`. Any idea? I though `glib_bool_error_result_from_glib` was a bit too long. Maybe `glib_result_from_glib_bool`?

sdroege: `glib_bool_error!(message)` and `glib_bool_error_from_gboolean!(message, b)` maybe?

src/error.rs
```diff
@@ -132,26 +133,122 @@ pub trait ErrorDomain: Copy {
...
+    }};
+);
+
 #[derive(Debug)]
-pub struct BoolError(pub &'static str);
+pub struct BoolError {
+    pub message: Cow<'static, str>,
+    pub(crate) filename: &'static str,
+    pub(crate) function: &'static str,
+    pub(crate) line: u32,
```
sdroege: You need to make these public or otherwise the macro will not work when called from other crates.

You can make them ignored for the docs though :)

src/error.rs
fengalin: I would tend to use `glib_result_from_gboolean` instead of `glib_bool_error_from_gboolean` since the return value is a `Result`.

Why do you prefer putting the boolean as the last argument?

sdroege: I don't mind either way :)

src/error.rs
```diff
@@ -132,26 +133,125 @@ pub trait ErrorDomain: Copy {
...
+    fn test_bool_error() {
+        use std::error::Error;
+
+        let from_static_msg = glib_bool_error!("Static message");
+        assert_eq!(from_static_msg.description(), "Static message");
+
+        let from_dynamic_msg = glib_bool_error!("{} message", "Dynamic");
+        assert_eq!(from_dynamic_msg.description(), "Dynamic message");
+
+        let static_res = glib_result_from_gboolean!(glib_ffi::GFALSE, "Static message");
```
EPashkin: Please add test for `glib_ffi::GTRUE` just for completeness

EPashkin:
@fengalin  Thanks, other that small nit all looks good for me.

src/error.rs
fengalin: Yes, you're right!

EPashkin:
IMHO you need only change this https://github.com/gtk-rs/gir/blob/master/src/codegen/translate_from_glib.rs#L72-L82

fengalin:
> IMHO you need only change this https://github.com/gtk-rs/gir/blob/master/src/codegen/translate_from_glib.rs#L72-L82

Yes, that's it. Thanks @EPashkin

fengalin:
@sdroege: `gstreamer-rs` is currently @https://github.com/gtk-rs/gir/commit/811e7118a5ac275a2ff912be72930285ca84a1d3. I see there are commits such as [Generate code using GString instead of String ](https://github.com/gtk-rs/gir/commit/3fef5be81ec8a1f4328956e94f7349ca9ea59658) after this. I presume, this would lead to issues if I update Gir. What do you think?

sdroege:
> @sdroege: `gstreamer-rs` is currently @[gtk-rs/gir@811e711](https://github.com/gtk-rs/gir/commit/811e7118a5ac275a2ff912be72930285ca84a1d3). I see there are commits such as [Generate code using GString instead of String ](https://github.com/gtk-rs/gir/commit/3fef5be81ec8a1f4328956e94f7349ca9ea59658) after this. I presume, this would lead to issues if I update Gir. What do you think?

That's fine, regenerating with newer gir shouldn't cause any problems.

fengalin:
> That's fine, regenerating with newer gir shouldn't cause any problems.

Yes indeed, it's fine. I checked `gstreamer-rs/gstreamer` with my modifications and unit tests are all good.

@EPashkin, it should be ready to merge now. See also https://github.com/gtk-rs/gir/pull/687 & https://github.com/gtk-rs/gtk/pull/751.

EPashkin:
@fengalin Yes, i checked all 3 PRs and it works fine, thanks.
@GuillaumeGomez 👍 for this.

GuillaumeGomez:
Compilation errors seem to be expected. Code looks good.

@fengalin: Thanks!

PR #421, opened 2019-01-07
Clean up glib_wrapper! macro and IsA<_>, Wrapper traits and their impls

sdroege:
This PR is still WIP as it also requires a corresponding [PR for gir](https://github.com/gtk-rs/gir/pull/689). The changes are
* Rename `Wrapper` trait to `ObjectType` to be more descriptive and add more useful API around it
* `glib_wrapper!` does not need FFI types for the parent classes in other crates anymore
* `Downcast` trait became `CanDowncast`, and the `Cast` trait impl was cleaned up
* Generate Rust class structs for class types in `glib_wrapper!`, and distinguish between classes and interfaces
* `IsA<_>` is cleaned up and does not require `ToGlibPtr` impls for everything anymore (that's why FFI types were needed). Instead `IsA<T>` requires `AsRef<T>`, i.e. we can always get back to a `&T` which then has all the traits we care about implemented.
* `IsA<Object>` and `ObjectType` (former `Wrapper`) trait bounds are equivalent, i.e. `IsA<Object>` is implemented for every `ObjectType`. They can be used interchangeably.

### DIFF
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -190,35 +190,6 @@
 /// }
 /// ```
 ///
-/// #### Wrapping objects with parents/interfaces from different crates
-///
-/// Implementing types whose parents or interfaces come from different
-/// crates requires specifying the wrapped names of the
-/// parents/interfaces and their FFI counterparts as well.  Note that
-/// these must be specified inside square brackets "`[]`", as
-/// comma-delimited pairs like
-/// "`crate_name::WrappedName => ffi_crate_name::Name`".
-///
-/// Here, note that the parent class for `ffi::GtkApplication` is
-/// `gio::Application`, which is the Rust wrapper for
-/// `gio_ffi::GApplication`.  Similarly, the `ActionGroup` and
-/// `ActionMap` interfaces, and their corresponding ffi structs, come
-/// from different crates.
-///
-/// ```ignore
-/// glib_wrapper! {
-///     pub struct Application(Object<ffi::GtkApplication, ffi::GtkApplicationClass>): [
-///         gio::Application => gio_ffi::GApplication,
-///         gio::ActionGroup => gio_ffi::GActionGroup,
-///         gio::ActionMap   => gio_ffi::GActionMap,
-///     ];
-///
-///     match fn {
-///         get_type => || ffi::gtk_application_get_type(),
-///     }
-/// }
-/// ```
-///
 /// #### Non-derivable classes
 ///
 /// By convention, GObject implements "final" classes, i.e. those who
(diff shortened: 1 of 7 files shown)


EPashkin:
Seems some `impl AsRef<$name>` is missing:
```
error[E0277]: the trait bound `auto::fontset_simple::FontsetSimple: std::convert::AsRef<glib::Object>` is not satisfied
  --> D:/eap/rust/0/pango\src\auto\fontset_simple.rs:13:1
   |
13 | / glib_wrapper! {
14 | |     pub struct FontsetSimple(Object<ffi::PangoFontsetSimple, ffi::PangoFontsetSimpleClass, FontsetSimpleClass>): Fontset;
15 | |
16 | |     match fn {
17 | |         get_type => || ffi::pango_fontset_simple_get_type(),
18 | |     }
19 | | }
   | |_^ the trait `std::convert::AsRef<glib::Object>` is not implemented for `auto::fontset_simple::FontsetSimple`
   |
             <auto::fontset_simple::FontsetSimple as std::convert::AsRef<auto::fontset::Fontset>>
             <auto::fontset_simple::FontsetSimple as std::convert::AsRef<auto::fontset_simple::FontsetSimple>>
             <auto::fontset_simple::FontsetSimple as std::convert::AsRef<glib::object::ObjectRef>>
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

EPashkin:
Good thing that Interface has its tag, but seems it lost it FFI link. Or it don't need anymore?
```diff
+++ b/src/auto/editable.rs
@@ -10,7 +10,7 @@ use std::fmt;
 use std::mem;
 
 glib_wrapper! {
-    pub struct Editable(Object<ffi::GtkEditable, ffi::GtkEditableInterface>);
+    pub struct Editable(Interface<ffi::GtkEditable>);
```

sdroege:
> Good thing that Interface has its tag, but seems it lost it FFI link. Or it don't need anymore?

For the class/iface struct that's not needed anymore. For interfaces there is currently nothing the macro could do with that.

From doing the bindings parts for implementing interfaces, I don't think anything will be needed here either. See e.g. https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/blob/master/gstreamer/src/subclass/uri_handler.rs

EPashkin:
Thanks  for explanation.

sdroege:
> Seems some `impl AsRef<$name>` is missing:

In some cases the macro does not expand to the part that implements `AsRef<Object>`, yes.

So generally we have another problem now though, that needs some further changes. We assume that the first parent type given is the direct parent class. Which is fine except for types that inherit directly from `glib::Object` (which is always omitted!) and implement an interface.

I think we'll need to always list `glib::Object` explicitly now. But that should be fine and nicer anyway as it makes this more explicit.

Will change it accordingly.

sdroege:
> Will change it accordingly.

Which is not trivial because we have to special-case `glib::Object` in the unwrapping of the parent types. Maybe the macro should distinguish `extends` (classes) and `implements` (interfaces) like in Java.

```
glib_wrapper! {
     pub struct Foo(Object<ffi::Foo, ffi::FooClass>) extends Bar, Baz, implements Meh;
     [...]
```

or similar.

@GuillaumeGomez @EPashkin do you have any opinions?

EPashkin:
IMHO extends/implements is better that direct adding `glib::Object`

EPashkin:
Not sure that will solve problem with object pango::`FontsetSimple`, it has only one parent object `Fontset` that don't have parents at all.

sdroege:
> Not sure that will solve problem with object pango::`FontsetSimple`

It doesn't, that's a different problem but I'll fix that at the same time as the other one I mentioned above :) Same code paths in the macro need refactoring.

> it has only one parent object `Fontset` that don't have parents at all.

Well, it has `glib::Object` as parent. Everything does :)

sdroege:
This is now ready too

src/wrapper.rs
```diff
@@ -119,18 +119,17 @@
 ///
 /// ### Object
 ///
-/// Objects -- classes and interfaces.  Note that the class name, if
-/// available, must be specified after the $foreign type; see below
-/// for [non-derivable classes][#non-derivable-classes].
+/// Objects -- classes.  Note that the class name, if available, must be specified after the
```
GuillaumeGomez: Two whitespaces in `classes.  Note`. Is it expected? Otherwise please remove one or start a new paragraph.

GuillaumeGomez:
Looks good to me. Just a little nit but it's just a detail for let's just merge for now.

PR #423, opened 2019-01-12
GString: Avoid copy for From<GString> for String impl

### LINKED ISSUE #422: GString: erroneous string copies in current implementation
```
[13:17:42] <slomo> philn: From<GString> for String currently copies the CString (in case of ForeignOwned), instead of just converting the CString into a String without copying
[13:18:57] <slomo> philn: similarly impl From<String> for GString can probably also happen without copying
```

philn:
Fixes #422

### DIFF
--- a/src/gstring.rs
+++ b/src/gstring.rs
@@ -21,7 +21,7 @@ use value::{FromValueOptional, Value, SetValueOptional, SetValue};
 
 #[derive(Debug)]
 pub enum GString {
-    ForeignOwned(CString),
+    ForeignOwned(Option<CString>),
     Borrowed(*const c_char, usize),
     Owned(*mut c_char, usize)
 }
@@ -48,7 +48,7 @@ impl GString {
                 CStr::from_bytes_with_nul_unchecked(bytes)
             }
             GString::ForeignOwned(cstring) => {
-                cstring.as_c_str()
+                cstring.as_ref().expect("ForeignOwned shouldn't be empty").as_c_str()
             }
         };
         cstr.to_str().unwrap()
@@ -179,7 +179,12 @@ impl Deref for GString {
 
 impl From<GString> for String {
     #[inline]
-    fn from(s: GString) -> Self {
+    fn from(mut s: GString) -> Self {
+        if let GString::ForeignOwned(ref mut cstring) = s {
+            if let Ok(s) = cstring.take().expect("ForeignOwned shouldn't be empty").into_string() {
+                return s;
+            }
+        }
         String::from(s.as_str())
     }
 }
@@ -195,7 +200,7 @@ impl From<GString> for Box<str> {
 impl From<String> for GString {
     #[inline]
     fn from(s: String) -> Self {
-        s.as_bytes().to_vec().into()
+        s.into_bytes().into()
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/gstring.rs
```diff
@@ -180,7 +180,16 @@ impl Deref for GString {
...
-        String::from(s.as_str())
+        if let GString::ForeignOwned(ref cstring) = s {
+            if let Ok(s) = String::from_utf8(cstring.as_bytes().to_vec()) {
+                s
+            } else {
+                String::from(s.as_str())
+            }
+        } else {
+            String::from(s.as_str())
+
```
EPashkin: Unneeded empty line.

EPashkin:
@philn Thansk, LGFM
Only possible change it do early return in nocopy case and use one call `String::from(s.as_str())`
but IMHO rust can optimize it yourself

philn:
Thanks, I updated the PR, code looks simpler now indeed!

src/gstring.rs
```diff
@@ -180,6 +180,11 @@ impl Deref for GString {
 impl From<GString> for String {
     #[inline]
     fn from(s: GString) -> Self {
+        if let GString::ForeignOwned(ref cstring) = s {
+            if let Ok(s) = String::from_utf8(cstring.as_bytes().to_vec()) {
```
EPashkin: Hm, I missed that we have buildin method
https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_string

philn: Sorry, I don't follow.

philn:
```
error[E0507]: cannot move out of borrowed content                                                                                                                                                                  
   --> src/gstring.rs:184:28                                                                                                                                                                                       
    |                                                                                                                                                                                                              
184 |             if let Ok(s) = cstring.into_string() {                                                                                                                                                           
    |                            ^^^^^^^ cannot move out of borrowed content                                                                                                                                       
                                                                                                                                                                                                                   
```

EPashkin: We want remove copy in foreign owned case.
As [`slice::to_vec`](https://doc.rust-lang.org/std/primitive.slice.html#method.to_vec) make copy,
we need get inner slice and give it to new String with transfer ownership.

sdroege: Yeah, remove the `ref` keyword here

philn: That's not possible because GString implements the Drop trait :/

sdroege: Then let's do the standard workaround for that :) Store the `CString` in an `Option<_>` and use `Option::take()` here, and in all other places use `as_ref().unwrap()` when accessing the `Option<_>`

src/gstring.rs
```diff
@@ -48,7 +48,7 @@ impl GString {
                 CStr::from_bytes_with_nul_unchecked(bytes)
             }
             GString::ForeignOwned(cstring) => {
-                cstring.as_c_str()
+                cstring.as_ref().unwrap().as_c_str()
```
EPashkin: Maybe use `expect`?

GuillaumeGomez: > Maybe use `expect`?

I approve!

philn:
CI is green and comments addressed :)

EPashkin:
@philn Thanks, 👍

GuillaumeGomez:
Thanks!

PR #425, opened 2019-01-16
implement DerefMut for glib_object_wrapper's rust_class

vhdirk:

PR #426, opened 2019-01-16
Implement DerefMut for class structs too

sdroege:
This is needed for subclassing


GuillaumeGomez:
Is it the same as #425?

sdroege:
> Is it the same as #425?

Yeah but more correct (it compiles), and includes some more cleanup

sdroege:
@GuillaumeGomez all green :)

GuillaumeGomez:
Then I merge. Thanks!

PR #428, opened 2019-01-21
generate from_glib_borrow implementation for const pointers

GuillaumeGomez:
cc @sdroege @EPashkin

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -206,6 +206,14 @@ macro_rules! glib_boxed_wrapper {
         }
 
         #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $crate::translate::from_glib_borrow(ptr as *mut $ffi_name)
+            }
+        }
+
+        #[doc(hidden)]
         impl $crate::translate::FromGlibContainerAsVec<*mut $ffi_name, *mut *mut $ffi_name> for $name {
             unsafe fn from_glib_none_num_as_vec(ptr: *mut *mut $ffi_name, num: usize) -> Vec<Self> {
                 if num == 0 || ptr.is_null() {
--- a/src/object.rs
+++ b/src/object.rs
@@ -532,6 +532,15 @@ macro_rules! glib_object_wrapper {
         }
 
         #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $crate::translate::from_glib_borrow(ptr as *mut $ffi_name)
+            }
+        }
+
+        #[doc(hidden)]
         impl $crate::translate::FromGlibContainerAsVec<*mut $ffi_name, *mut *mut $ffi_name> for $name {
             unsafe fn from_glib_none_num_as_vec(ptr: *mut *mut $ffi_name, num: usize) -> Vec<Self> {
                 if num == 0 || ptr.is_null() {
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -185,6 +185,14 @@ macro_rules! glib_shared_wrapper {
         }
 
         #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $crate::translate::from_glib_borrow(ptr as *mut $ffi_name)
+            }
+        }
+
+        #[doc(hidden)]
         impl $crate::translate::FromGlibContainerAsVec<*mut $ffi_name, *mut *mut $ffi_name> for $name {
             unsafe fn from_glib_none_num_as_vec(ptr: *mut *mut $ffi_name, num: usize) -> Vec<Self> {
                 if num == 0 || ptr.is_null() {


src/boxed.rs
```diff
@@ -205,6 +205,14 @@ macro_rules! glib_boxed_wrapper {
             }
         }
 
+        #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $name($crate::translate::from_glib_borrow(ptr as usize as *mut $ffi_name))
```
sdroege: Why the cast via `usize`? That shouldn't be needed at all, just the cast to a mutable pointer should be needed

sdroege: Also everywhere else

src/boxed.rs
```diff
@@ -391,6 +399,17 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> FromGlibPtrBorrow<*mut T> for Boxed<
     }
 }
 
+impl<T: 'static, MM: BoxedMemoryManager<T>> FromGlibPtrBorrow<*const T> for Boxed<T, MM> {
```
sdroege: This shouldn't be needed as you always go via the `*mut` variant anyway

sdroege:
Looks good to me otherwise (same thing in all 3 files)

GuillaumeGomez:
Like this?

src/shared.rs
```diff
@@ -400,3 +408,15 @@ impl<T: 'static, MM: SharedMemoryManager<T>> FromGlibPtrBorrow<*mut T> for Share
         }
     }
 }
+
+impl<T: 'static, MM: SharedMemoryManager<T>> FromGlibPtrBorrow<*const T> for Shared<T, MM> {
```
sdroege: Shouldn't be needed

src/shared.rs
```diff
@@ -184,6 +184,14 @@ macro_rules! glib_shared_wrapper {
             }
         }
 
+        #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $name($crate::translate::from_glib_borrow(ptr as *mut $ffi_name))
```
sdroege: I believe you can also do
```
$crate::translate::from_glib_borrow(ptr as *mut $ffi_name)
```
here and in the other places. Without the `$name(...)` part.

GuillaumeGomez: Isn't it what I did?

GuillaumeGomez: Oh nevermind.

GuillaumeGomez:
Updated.

src/boxed.rs
```diff
@@ -205,6 +205,14 @@ macro_rules! glib_boxed_wrapper {
             }
         }
 
+        #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $name($crate::translate::from_glib_borrow(ptr as *mut $ffi_name))
```
sdroege: You can also remove the `$name(...)` part here :)

src/object.rs
```diff
@@ -531,6 +531,17 @@ macro_rules! glib_object_wrapper {
...
         }
 
+        #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                debug_assert!($crate::types::instance_of::<Self>(ptr as *const _));
+                $name($crate::translate::from_glib_borrow(ptr as *mut _),
+                      ::std::marker::PhantomData)
```
sdroege: And here. And the assertion is already in the function you'd call then, and could also be removed

src/boxed.rs
GuillaumeGomez: It's everywhere! :o

src/object.rs
```diff
@@ -531,6 +531,15 @@ macro_rules! glib_object_wrapper {
             }
         }
 
+        #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $crate::translate::from_glib_borrow(ptr as *mut _)
```
sdroege: This could also be an explicit cast to `*mut $ffi_name`, but otherwise all good :) Please merge after changing that

GuillaumeGomez:
I removed stuff and updated the `$crate_name` stuff.

src/object.rs
GuillaumeGomez: You order, I execute!

sdroege:
:+1:

src/object.rs
GuillaumeGomez: No, it needs to be `as _`.

src/object.rs
```diff
@@ -526,11 +526,20 @@ macro_rules! glib_object_wrapper {
             #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
             unsafe fn from_glib_borrow(ptr: *mut $ffi_name) -> Self {
                 debug_assert!($crate::types::instance_of::<Self>(ptr as *const _));
-                $name($crate::translate::from_glib_borrow(ptr as *mut _),
+                $name($crate::translate::from_glib_borrow(ptr as *mut $ffi_name),
```
EPashkin: `ptr: *mut $ffi_name`  converted to `*mut $ffi_name` ? seems previous it was something else, IMHO ObjectRef need `*mut $crate::gobject_ffi::GObject` but not sure .

src/object.rs
GuillaumeGomez: Otherwise we get:

```
error[E0277]: the trait bound `object::ObjectRef: translate::FromGlibPtrBorrow<*mut gobject_ffi::GBinding>` is not satisfied
   --> src/object.rs:529:23
    |
529 |                   $name($crate::translate::from_glib_borrow(ptr as *mut $ffi_name),
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `translate::FromGlibPtrBorrow<*mut gobject_ffi::GBinding>` is not implemented for `object::ObjectRef`
    | 
   ::: src/gobject/auto/binding.rs:12:1
    |
12  | / glib_wrapper! {
13  | |     pub struct Binding(Object<ffi::GBinding, BindingClass>);
14  | |
15  | |     match fn {
16  | |         get_type => || ffi::g_binding_get_type(),
17  | |     }
18  | | }
    | |_- in this macro invocation
    |
              <object::ObjectRef as translate::FromGlibPtrBorrow<*const gobject_ffi::GObject>>
              <object::ObjectRef as translate::FromGlibPtrBorrow<*mut gobject_ffi::GObject>>
note: required by `translate::from_glib_borrow`
   --> src/translate.rs:1052:1
    |
1052| pub unsafe fn from_glib_borrow<P: Ptr, T: FromGlibPtrBorrow<P>>(ptr: P) -> T {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0277]: the trait bound `object::ObjectRef: translate::FromGlibPtrBorrow<*mut gobject_ffi::GBinding>` is not satisfied
   --> src/object.rs:529:23
    |
529 |                   $name($crate::translate::from_glib_borrow(ptr as *mut $ffi_name),
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `translate::FromGlibPtrBorrow<*mut gobject_ffi::GBinding>` is not implemented for `object::ObjectRef`
    | 
   ::: src/gobject/auto/binding.rs:12:1
    |
12  | / glib_wrapper! {
13  | |     pub struct Binding(Object<ffi::GBinding, BindingClass>);
14  | |
15  | |     match fn {
16  | |         get_type => || ffi::g_binding_get_type(),
17  | |     }
18  | | }
    | |_- in this macro invocation
    |
              <object::ObjectRef as translate::FromGlibPtrBorrow<*const gobject_ffi::GObject>>
              <object::ObjectRef as translate::FromGlibPtrBorrow<*mut gobject_ffi::GObject>>
note: required by `translate::from_glib_borrow`
   --> src/translate.rs:1052:1
    |
1052| pub unsafe fn from_glib_borrow<P: Ptr, T: FromGlibPtrBorrow<P>>(ptr: P) -> T {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

src/object.rs
sdroege: Yes, the whole line needs to be `$crate::translate::from_glib_borrow(ptr as *const $ffi_name)`. No need to put `$name(...)` around it all.

sdroege:
Now looks all good to me

src/object.rs
GuillaumeGomez: I replaced with `as _`. Seems better.

sdroege: Why? You need a `*mut $ffi_name` here and not some random pointer. It's safer to be explicit here

sdroege: Now it is with `$name(...)` and the cast to `*mut $ffi_type`, and that's the only combination that does not work.

EPashkin: Locally I get this error:
```
error[E0277]: the trait bound `object::ObjectRef: translate::FromGlibPtrBorrow<*mut gobject_ffi::GBinding>` is not satisfied
   --> D:/eap/rust/0/glib\src\object.rs:529:23
    |
529 |                   $name($crate::translate::from_glib_borrow(ptr as *mut $ffi_name),
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `translate::FromGlibPtrBorrow<*mut gobject_ffi::GBinding>` is not implemented for `object::ObjectRef`
    | 
   ::: D:/eap/rust/0/glib\src\gobject\auto\binding.rs:12:1
    |
12  | / glib_wrapper! {
13  | |     pub struct Binding(Object<ffi::GBinding, BindingClass>);
14  | |
15  | |     match fn {
16  | |         get_type => || ffi::g_binding_get_type(),
17  | |     }
18  | | }
    | |_- in this macro invocation
    |
              <object::ObjectRef as translate::FromGlibPtrBorrow<*const gobject_ffi::GObject>>
              <object::ObjectRef as translate::FromGlibPtrBorrow<*mut gobject_ffi::GObject>>
note: required by `translate::from_glib_borrow`
   --> D:/eap/rust/0/glib\src\translate.rs:1052:1
    |
1052| pub unsafe fn from_glib_borrow<P: Ptr, T: FromGlibPtrBorrow<P>>(ptr: P) -> T {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

sdroege: Yes, that's exactly what I'm saying :) Remove the `$name(...)` part and it will work

EPashkin: Removing name IMHO don't help as this is definition from_glib_borrow(*mut $ffi_name) that used by `::from_glib_borrow`

EPashkin: `*mut $crate::gobject_ffi::GObject` IMHO solve problem

sdroege: We might be talking about different functions. I mean the implementation for `*const $ffi_name` that should just call into the one for `*mut $ffi_name`.

The implementation for `*mut $ffi_name` needs the `$name(...)` wrapper and has to be cast to `*mut GObject`.

GuillaumeGomez:
We now have this error:

```
thread 'subclass::object::test::test_signals' has overflowed its stack
```

I can't figure out where it's coming from for the moment so I'll check tomorrow.

src/object.rs
```diff
@@ -526,8 +526,16 @@ macro_rules! glib_object_wrapper {
             #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
             unsafe fn from_glib_borrow(ptr: *mut $ffi_name) -> Self {
                 debug_assert!($crate::types::instance_of::<Self>(ptr as *const _));
-                $name($crate::translate::from_glib_borrow(ptr as *mut _),
-                      ::std::marker::PhantomData)
+                $crate::translate::from_glib_borrow(ptr as *mut $ffi_name)
```
sdroege: The stack overflow comes from here. This impl should've stayed as it was. I was only talking about the new impls in my review comments.

What now happens is that this impl here is calling itself again.

src/object.rs
```diff
@@ -526,8 +526,16 @@ macro_rules! glib_object_wrapper {
...
+            }
+        }
+
+        #[doc(hidden)]
+        impl $crate::translate::FromGlibPtrBorrow<*const $ffi_name> for $name {
+            #[inline]
+            #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+            unsafe fn from_glib_borrow(ptr: *const $ffi_name) -> Self {
+                $crate::translate::from_glib_borrow(ptr as *mut _)
```
sdroege: And this one here can get an explicit cast to `*mut $ffi_type`

GuillaumeGomez:
All green!

sdroege:
:+1:

PR #431, opened 2019-01-25
Add new Object::connect_unsafe(), Object::connect_notify_unsafe() and…

sdroege:
… Closure::new_unsafe()

These allow to use non-Send/non-Sync closures and it's the job of the
caller to ensure that the closures are only used in a safe way.

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -32,10 +32,16 @@ glib_wrapper! {
 
 impl Closure {
     pub fn new<F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static>(callback: F) -> Self {
+        unsafe {
+            Closure::new_unsafe(callback)
+        }
+    }
+
+    pub unsafe fn new_unsafe<F: Fn(&[Value]) -> Option<Value>>(callback: F) -> Self {
         unsafe extern "C" fn marshal<F>(_closure: *mut gobject_ffi::GClosure, return_value: *mut gobject_ffi::GValue,
             n_param_values: c_uint, param_values: *mut gobject_ffi::GValue, _invocation_hint: *mut c_void,
             marshal_data: *mut c_void)
-            where F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static
+            where F: Fn(&[Value]) -> Option<Value>
         {
             let values = slice::from_raw_parts(param_values as *const _, n_param_values as usize);
             let callback: Box<F> = Box::from_raw(marshal_data as *mut _);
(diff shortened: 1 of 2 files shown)


src/closure.rs
```diff
@@ -53,23 +59,21 @@ impl Closure {
...
-            let closure = gobject_ffi::g_closure_new_simple(size, ptr::null_mut());
-            assert_ne!(closure, ptr::null_mut());
-            let callback = Box::new(callback);
-            let ptr: *mut F = Box::into_raw(callback);
-            let ptr: *mut c_void = ptr as *mut _;
-            gobject_ffi::g_closure_set_meta_marshal(closure, ptr, Some(marshal::<F>));
-            gobject_ffi::g_closure_add_finalize_notifier(closure, ptr, Some(finalize::<F>));
-            from_glib_none(closure)
-        }
+        let size = 4 + 4 + 3 * mem::size_of::<*mut c_void>() as u32;
```
GuillaumeGomez: Could you comment what elements of this multiplication are please?

src/object.rs
```diff
@@ -1053,64 +1060,69 @@ impl<T: ObjectType> ObjectExt for T {
...
+                    Some(ret) => {
+                        let valid_type: bool = from_glib(gobject_ffi::g_type_check_value_holds(
+                                mut_override(ret.to_glib_none().0),
+                                return_type.to_glib()));
+                        if !valid_type {
+                            panic!("Signal required return value of type {} but got {}",
+                                   return_type.name(), ret.type_().name());
+                        }
+                        Some(ret)
+                    },
```
GuillaumeGomez: No need for the comma.

src/object.rs
```diff
@@ -1053,64 +1060,69 @@ impl<T: ObjectType> ObjectExt for T {
...
+                    None => {
+                        panic!("Signal required return value of type {} but got None", return_type.name());
+                    },
```
GuillaumeGomez: No need for the comma.

src/closure.rs
sdroege: Sure, but tomorrow :) Note that all your comments are already for the existing code. This code only lost a level of indentation

EPashkin:
@sdroege Thanks, LVGFM

src/closure.rs
sdroege: Done, and also improved the calculation a bit

src/closure.rs
```diff
@@ -53,23 +59,28 @@ impl Closure {
...
-            from_glib_none(closure)
-        }
+        // Due to bitfields we have to do our own calculations here for the size of the GClosure:
+        // - 4: 32 bits in guint bitfields at the beginning
+        // - padding due to alignment needed for the following pointer
+        // - 3 * size_of<*mut c_void>: 3 pointers
+        // We don't store any custom data ourselves in the GClosure
+        let size = 4
+            + (mem::align_of::<*mut c_void>() as u32 - 4)
+            + 3 * mem::size_of::<*mut c_void>() as u32;
```
EPashkin: maybe better use max(4, mem::aligin_of) ?

sdroege: No, that would be wrong. Consider for example 16 byte alignment for pointers, we would have to add 12 bytes of padding here and not 16 (with 16 bytes of padding we would be unaligned again).

Or take 4 byte alignment: no padding would have to be added, but you would add 4 bytes then. Which does not cause problems in our specific case but is not needed either.

EPashkin: I does mean only instead first 2 addendums,
but this also not error prone if "align_of<*void>" < 4 same as your case.
Maybe add assert about this?

sdroege: > I does mean only instead first 2 addendums,

I don't understand what you mean, can you write the whole calculation you think would be better?

We need 4 bytes for the bitfields, N bytes padding, then 3 pointers.

> but this also not error prone if "align_of<*void>" < 4 same as your case.
> Maybe add assert about this?

I'm not aware of any such platform but that's a good idea nonetheless.

sdroege: Ah indeed, that's slightly nicer and then we also don't need the assert. Thanks! Will update when I'm back at the computer.

sdroege: Done :)

PR #438, opened 2019-02-01
Add support for defining new interfaces and other minor things

sdroege:
See individual commits

### DIFF
--- a/src/subclass/interface.rs
+++ b/src/subclass/interface.rs
@@ -11,7 +11,7 @@ use std::ptr;
 
 use super::{InitializingType, Property};
 use translate::*;
-use {StaticType, Type, Value};
+use {IsA, Object, ObjectExt, StaticType, Type, Value};
 
 impl<T: ObjectInterface> InitializingType<T> {
     /// Adds an interface prerequisite for `I` to the type.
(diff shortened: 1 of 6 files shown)


src/subclass/interface.rs
```diff
@@ -0,0 +1,241 @@
...
+                *(self as *mut _ as *mut ffi::GType),
+                name,
+                arg_types,
+                ret_type,
+                accumulator,
+            );
+        }
+    }
+
+    /// Add a new action signal with accumulator to the interface.
```
EPashkin: With accumulator?

EPashkin:
Thanks, found only small nit in docs

src/subclass/interface.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// should be combined with the old value in the first argument. If no further
+    /// signal handlers should be called, `false` should be returned.
+    fn add_signal_with_accumulator<F>(
+        &mut self,
+        name: &str,
+        arg_types: &[Type],
+        ret_type: Type,
+        accumulator: F,
+    ) where
+        F: Fn(&mut Value, &Value) -> bool + Send + Sync + 'static,
```
EPashkin: Maybe use enum { None, TrueHandled, FirstWins, .., Custom(F:Fn())} for accumulator?
Not sure how to handle `g_application_handle_local_options_accumulator` etc,
maybe define trait to get function pointer?

src/subclass/interface.rs
sdroege: What do you mean? :)

Also this is the same documentation as we already had for subclasses before

src/subclass/interface.rs
sdroege: An enum would have the same problem that we had with `Option` for optional callbacks, and `Fn` is a trait for giving a function pointer and passing some data around ;)

What exactly do you see wrong with the way how the function is defined? Also this is how we had it before for subclasses already, no change here.

If you'd like to have a different API for that, maybe let's handle that in a separate PR?

sdroege: `g_application_handle_local_options_accumulator` would be

```rust
|return_acc, handler_return| {
  let value = handler_return.get::<u32>().unwrap();
  *return_acc = value.to_value();
  value < 0
}
```

src/subclass/interface.rs
EPashkin: I missed that `handler` actually `GSignalAccumulator accumulator`, sorry

src/subclass/interface.rs
EPashkin: Yes can handled later, just note about extension

EPashkin:
👍

sdroege:
Ok, let's get this in then? :) @GuillaumeGomez @EPashkin?

I added another small commit on top.

EPashkin:
👍

GuillaumeGomez:
All good for me!

EPashkin:
Still 👍
Thanks @sdroege

PR #445, opened 2019-02-07
Minor subclass API usability improvements

sdroege:


GuillaumeGomez:
The failure for examples can be ignored (there is no more `StatusIcon` now so obviously...).

EPashkin:
👍

PR #446, opened 2019-02-08
Update Travis to Ubuntu Xenial

sdroege:
And add GTK 3.24 builds for Linux and update the minimum version to 3.14.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,12 +17,13 @@ install:
 
 build_script:
   - cargo doc --features "dox"
-  - cargo test
+  - cargo test --features subclassing
+  - cargo test --features subclassing,v2_58
   - mkdir .cargo
   - echo paths = ["."] > .cargo\config
   - git clone -q --depth 50 -b pending https://github.com/gtk-rs/examples _examples
   - cd _examples
   - cargo build
-  - cargo build --features gtk_3_18
+  - cargo build --features gtk_3_24
 
 test: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,49 +1,49 @@
+dist: xenial
 language: rust
 matrix:
   include:
     - os: linux
       rust: nightly
-      env: GTK=3.4
+      env: GTK=3.14 FEATURES=subclassing
     - os: linux
       rust: nightly
-      env: GTK=3.22.30
+      env: GTK=3.24 FEATURES=subclassing,v2_58
     - os: linux
       rust: beta
-      env: GTK=3.4
+      env: GTK=3.14 FEATURES=subclassing
     - os: linux
       rust: beta
-      env: GTK=3.22.30
+      env: GTK=3.24 FEATURES=subclassing,v2_58
     - os: linux
       rust: 1.31.0
-      env: GTK=3.4
+      env: GTK=3.14 FEATURES=subclassing
     - os: linux
       rust: 1.31.0
-      env: GTK=3.22.30
+      env: GTK=3.24 FEATURES=subclassing,v2_58
     - os: osx
       rust: nightly
-      env: GTK=3.4
+      env: GTK=3.14 FEATURES=subclassing
     # - os: osx
     #   rust: nightly
-    #   env: GTK=3.22.30
+    #   env: GTK=3.24 FEATURES=subclassing,v2_58
     - os: osx
       rust: beta
-      env: GTK=3.4
+      env: GTK=3.14 FEATURES=subclassing
     # - os: osx
     #   rust: beta
-    #   env: GTK=3.22.30
+    #   env: GTK=3.24 FEATURES=subclassing,v2_58
     - os: osx
       rust: stable
-      env: GTK=3.4
+      env: GTK=3.14 FEATURES=subclassing
     # - os: osx
     #   rust: stable
-    #   env: GTK=3.22.30
+    #   env: GTK=3.24 FEATURES=subclassing,v2_58
     - os: linux
       rust: nightly
-      env: GTK=3.4 ARM=1 OTHER_TARGET="--target armv7-unknown-linux-gnueabihf"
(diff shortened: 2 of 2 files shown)


sdroege:
We don't really test the GLib 2.50 (GTK 3.22.30) or GLib 2.58 (GTK 3.24) features though. It would be good to do that but I have no idea how to do that via the Travis toml file.

@EPashkin @GuillaumeGomez ?

EPashkin:
IMHO 3.22.30 unneeded as we have 3.24
👍 for this and all in this batch

sdroege:
But how would I do a build with `--features=v2_58`? The problem is that it would mean duplicating all the lines and adding ifs before them if I don't miss anything...

And if I can somehow set variables globally for the run, the problem is that for the default build there would be an empty `--features=` then or none at all, which does not compose well with everything else in there.

EPashkin:
About testing build for GLIB2.50: we can add "gio/v2_58" to last version in examples/Cargo.toml
Note that it only v3_22_30 now https://github.com/gtk-rs/examples/blob/master/Cargo.toml#L28 
and travis also also not changed https://github.com/gtk-rs/examples/blob/master/.travis.yml#L13

sdroege:
Ah so we simple add the corresponding versions of all the crates there? But that doesn't give us `cargo test --features=v2_58` for glib, only building of the examples.

EPashkin:
IMHO don't need add it no all crates, only to examples

sdroege:
> IMHO don't need add it no all crates, only to examples

But we don't run tests for anything but the default features then. At least gio has tests that only run with newer versions, for example, and if we add more tests over time this will become more common.

EPashkin:
If you want add other feature to `cargo test --features "subclassing";` in glib,
then IMHO better just add local environment variable FEATURES and add "v2_58" to it for gtk3.24

sdroege:
So simply always to a build with `--features=subclassing,$OTHERFEATURES` ? I.e. we won't have any build/test without the subclassing feature? Because otherwise we would have an empty features line in case of 2.42.

EPashkin:
> Because otherwise we would have an empty features line in case of 2.42.
Sorry, don't understand that.

In glib I prefer test all with only "subclassing" except 3.24 that do it with "subclassing;v2_58"

EPashkin:
@sdroege Yes, IMHO we need test with subclassing for all cases

EPashkin:
I says about FEATURE environment variable because we have 2 uses of features in .travis.yml.

sdroege:
> @sdroege Yes, IMHO we need test with subclassing for all cases

I was going to also run tests without to ensure that this actually works. If we only ever test with subclassing, we could as well remove the feature at all and enable it by default always :)

What would that `FEATURES` variable contain exactly? `FEATURES=` for 3.14, `FEATURES=--features=v2_58` for 3.24? And you think I should simply remove the 3.22.30 build?

EPashkin:
IMHO `FEATURES=subclassing` for 3.14, `FEATURES=subclassing;v2_58` for 3.24,
usage `cargo test --features "subclassing"` and `... --features "subclassing" $OTHER_TARGET` for other use
Case with `FEATURES=--features=subclassing;v2_58` IMHO also appropriate as we don't have spaces in feature names
Yes, I don't see reason to test `3.22.30` now.

EPashkin:
@GuillaumeGomez You think that we need test 3 version of gtk?

sdroege:
Like this then?

EPashkin:
I thought do it only in script part, but this way we have better control on flags 👍

sdroege:
Ah ffs, Travis of course does not have GLib 2.58 so we can't do that anyway.

sdroege:
That's for someone else to clean up then, this Travis setup is not useful for testing what we're doing and I don't know how that's supposed to work with Travis.

I'll revert back to simply having 2.48 builds, as that is what Travis has.

EPashkin:
@sdroege If you know how to test open source project on docker-ed Gitlab CI for free,
can you try it on some your project?

sdroege:
That's what all the GNOME projects are doing on gitlab.gnome.org.

sdroege:
Can someone kick travis? It's building with the old commit that used 2.58 and did not even schedule a new build for the latest commit.

sdroege:
Maybe it's fine now, let's see.

EPashkin:
@sdroege You have write right on this repo, so IMHO you can kick travis yourself.

GuillaumeGomez:
> @GuillaumeGomez You think that we need test 3 version of gtk?

Two should be fine but 3 doesn't bother me either so let's just do as needed.

appveyor.yml
```diff
@@ -17,12 +17,13 @@ install:
 
 build_script:
   - cargo doc --features "dox"
-  - cargo test
+  - cargo test --feature subclassing
+  - cargo test --feature subclassing,v2_58
```
EPashkin: missed `S` and too new version

EPashkin: Maybe first test without subclassing?

EPashkin: Sorry, was wrong, test with subclassing and default version is needed really

sdroege: The version is fine, I was only missing the `s`. See the log

```
mingw-w64-i686-glib2-2.58.1-1
```

sdroege:
All green now. Is everybody ok with that pattern for now until we figure out a better way for building things?

@EPashkin btw for your question about docker. You basically tell in the `gitlab-ci.yaml` which docker image(s) you want to use, and then you have that and can run things in there. It's literally just adding a line for the image to use in the simple case like ours. There must be something like this for Travis.

GuillaumeGomez:
Nope, all good for me for now.

sdroege:
Ok, then I'll go through the other repos one by one slowly over the next hours or maybe days :)

EPashkin:
@sdroege Thanks.
Can you start with examples, current version of https://github.com/gtk-rs/examples/blob/master/build_travis.sh seems just don't use 3.24.

EPashkin:
Sorry, forgot about "pending" branch

PR #450, opened 2019-02-13
subclassing: move parent invocation fnct in Ext trait

fengalin:
This reduces the risk for the user to override them, or even worse, let the user believe that it has an actual effect to override them.

See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/merge_requests/223#note_112472


sdroege:
> This reduces the risk for the user to override them.

Or worse, the user believing that it has an actual effect to override them :)

fengalin:
> Or worse, the user believing that it has an actual effect to override them :)

Indeed :) Do you want me to amend the commit message?

src/subclass/object.rs
```diff
@@ -320,7 +306,31 @@ unsafe impl<T: ObjectSubclass> IsSubclassable<T> for ObjectClass {
     }
 }
 
-pub trait ObjectImplExt: ObjectImpl + ObjectSubclass {
+pub trait ObjectImplExt {
+    /// Chain up to the parent class' implementation of `glib::Object::constructed()`.
+    ///
+    /// Do not override this, it has no effect.
```
sdroege: Well, it can't be overridden anymore so this part of the docs can go away :)

sdroege:
Looks good to me apart from that line in the docs, thanks!

sdroege:
> Indeed :) Do you want me to amend the commit message?

No that's fine

fengalin:
Note that AppVeyor tests are marked as "succeeded" when the 64bits version is still in progress.

src/subclass/object.rs
fengalin: Done! :)

sdroege:
Thanks, will merge once CI is happy again

PR #453, opened 2019-02-15
0.6

GuillaumeGomez:
We'll need to have this commit in the next release since the whole branch support isn't there yet...

cc @EPashkin @sdroege


sdroege:
These are already in `master`:
* 0acd0a4733f210412a9d3c58db98666a77b5b84a
* 89670b767fba208ecd4fc9f217be8d698df4a857

GuillaumeGomez:
Ah crap, hashes didn't match. My bad!

PR #455, opened 2019-02-16
Add test for checking if MainContext Senders return errors when the R…

sdroege:
…eceiver is removed from the context


GuillaumeGomez:
👍

EPashkin:
Thanks, 👍

Issue #454, opened 2019-02-16
Context passed to Receiver::attach() is not released when removing its source


gdesmott:
@sdroege : you told me the other day that the callback passed to `attach()` should be released when removing the source  but that doesn't seem to be the case according to this test case.

```rust
#[derive(Debug)]
struct Test(u32);

impl Drop for Test {
    fn drop(&mut self) {
        dbg!("dropped");
    }
}

pub fn main() {
    gtk::init().unwrap();

    let (_tx, rx) = glib::MainContext::channel(glib::PRIORITY_DEFAULT);

    let test = Test(1);
    let source = rx.attach(None, move |_u: u32| {
        dbg!(&test);
        glib::Continue(true)
    });

    glib::Source::remove(source).unwrap();

    gtk::main();
}
```

sdroege:
That's bad :) Let's analyze that and transform it into an unit test then. Thanks for noticing!

sdroege:
Your problem is that the sender (`_tx`) is still alive. The sender also has a reference to the underlying `GSource`, and the closure will only be dropped once the `GSource` is finalized. Which will happen once nothing has a reference to it anymore. In your case the reference owned by the main context is gone but the one from the sender not.

If you name it `_` instead of `_tx` it will also already be dropped. Assignments to `_` are dropped immediately, `_foo` is assigning to a variable that you can use but there won't be any warning about it not being used and once it goes out of scope it will be dropped.

sdroege:
Let me know if this answers your question and if you think there's still something wrong here. There were actually tests for these already.

Note that the channel receiver is also automatically removed once the last sender is dropped (as there's no possible way for the closure to be called anymore after that).

gdesmott:
Ah ok, I thought the context was dropped if either the source was removed or all the senders. Wouldn't that make sense from an API pov?

This is a simplified version of my problem as I was looking as a way to break reference cycles. I guess I'll have to drop the sender then.

sdroege:
> Ah ok, I thought the context was dropped if either the source was removed or all the senders. Wouldn't that make sense from an API pov?

That would be ideal but I don't know how to implement that nicely. There's no `GSourceFunc` or anything that tells us when the `GSource` is removed (aka `g_source_destroy()`).

sdroege:
Added another small test for a different variant: https://github.com/gtk-rs/glib/pull/455

When the receiver is removed from the context, all your senders will return errors when trying to send something. Maybe that can help in your case?

[...]

PR #456, opened 2019-02-18
Re-implement MainContext channel around a manual channel

sdroege:
This allows us to ensure that dropping the Receiver and its GSource will
also directly drop the closure, instead of having to wait for all
Senders to disappear too.

We have to use a mutex and a shared struct for this, and as such it
makes sense to directly implement the channel as part of this shared
struct. As the std mpsc channel internally also uses mutexes this should
not cause any considerable performance difference.

At the same time also simplify some more code and add a few more tests.

Fixes https://github.com/gtk-rs/glib/issues/454

----

CC @gdesmott this seems to work now but I'll have to write a few more tests (especially for the sync channel). Please give it a try and let me know how well it works for you. Once the Receiver or its GSource is finalized the closure should be freed now.

### DIFF
--- a/src/main_context_channel.rs
+++ b/src/main_context_channel.rs
@@ -3,11 +3,11 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use std::cell::RefCell;
-use std::marker::PhantomData;
+use std::collections::VecDeque;
 use std::mem;
 use std::ptr;
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc;
+use std::sync::{Arc, Condvar, Mutex};
 
 use Continue;
 use MainContext;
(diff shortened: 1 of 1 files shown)


gdesmott:
My code isn't building with it, I'm not sure why:

```
error[E0277]: `*mut glib_sys::GSource` cannot be sent between threads safely
  --> src/web.rs:30:1
   |
30 | / fn play(web: State<Web>) {
31 | |     let tx = web.tx.lock().unwrap();
32 | |     tx.send(Message::Play).unwrap();
33 | | }
   | |_^ `*mut glib_sys::GSource` cannot be sent between threads safely
   |
   = help: within `glib::main_context_channel::ChannelInner<karapulse::Message>`, the trait `std::marker::Send` is not implemented for `*mut glib_sys::GSource`
   = note: required because it appears within the type `std::option::Option<*mut glib_sys::GSource>`
   = note: required because it appears within the type `glib::main_context_channel::ChannelInner<karapulse::Message>`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Mutex<glib::main_context_channel::ChannelInner<karapulse::Message>>`
   = note: required because it appears within the type `(std::sync::Mutex<glib::main_context_channel::ChannelInner<karapulse::Message>>, std::option::Option<glib::main_context_channel::ChannelBound>)`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<(std::sync::Mutex<glib::main_context_channel::ChannelInner<karapulse::Message>>, std::option::Option<glib::main_context_channel::ChannelBound>)>`
   = note: required because it appears within the type `glib::main_context_channel::Channel<karapulse::Message>`
   = note: required because it appears within the type `std::option::Option<glib::main_context_channel::Channel<karapulse::Message>>`
   = note: required because it appears within the type `glib::main_context_channel::Sender<karapulse::Message>`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Mutex<glib::main_context_channel::Sender<karapulse::Message>>`
   = note: required because it appears within the type `web::Web`
   = note: required by `rocket::request::state::State`
```

Code is at https://github.com/gdesmott/karapulse/tree/cdg if you want to test.

sdroege:
Oops, please try again :)

gdesmott:
Ah, that works now. And it does to break my ref cycle as I was hoping!
Thanks a lot, you're making my life so much easier. :+1:

sdroege:
Compiles here now.

@gdesmott In this [function](https://github.com/gdesmott/karapulse/blob/cdg/src/player.rs#L19) you could use the new `Bus::add_watch_local()` instead of the signal to not require the closure to be `Send+Sync`. Maybe that helps with something?

gdesmott:
Oh, I didn't know about this API, I'll give it a shot.
Do I still have to call `remove_signal_watch()` once I'm done with it?

I struggled quiet a bit to get the threading/callback model working (see for example the inner pattern in Karapulse which I'm not so sure about) so any recommendation about how to do it properly is really welcome. :)

sdroege:
>  Do I still have to call `remove_signal_watch()` once I'm done with it?

`Bus::remove_watch()`. That API in GStreamer is quite weird on the C level :)

> I struggled quiet a bit to get the threading/callback model working (see for example the inner pattern in Karapulse which I'm not so sure about) so any recommendation about how to do it properly is really welcome. :)

Ask on IRC if you have specific questions :)

src/main_context_channel.rs
```diff
@@ -18,74 +18,269 @@ use SourceId;
...
+                cond: Condvar::new(),
+            }),
+        )))
+    }
+
+    fn send(&self, t: T) -> Result<(), mpsc::SendError<T>> {
+        let mut inner = (self.0).0.lock().unwrap();
+
+        // If we have a bounded channel then we need to wait here until enough free space is
+        // available or the receiver disappears
```
GuillaumeGomez: You have an issue with dots, don't you? :p

GuillaumeGomez:
Looks good to me, thanks!

src/main_context_channel.rs
```diff
@@ -18,74 +18,269 @@ use SourceId;
...
+            // Receiver still running
+            Some(_) => false,
+        }
+    }
+
+    fn set_ready_time(&mut self, ready_time: i64) {
+        if let Some(source) = self.source {
+            if !source.is_null() {
+                unsafe {
+                    glib_ffi::g_source_set_ready_time(source, ready_time);
```
EPashkin: Maybe this function can return `true` if it really set time?

sdroege: It could but nothing would make use of that boolean :)

EPashkin: Agree after use check

EPashkin:
@sdroege Thanks, 👍

sdroege:
This should be good to go now. I've added quite a few more tests for the sync channel and various special cases, and also cleaned up some code.

sdroege:
@EPashkin @GuillaumeGomez opinions? :)

GuillaumeGomez:
I already approved it and I maintain my position. :)

sdroege:
Then let's get it in, thanks :)

PR #457, opened 2019-02-21
For setting properties and the return value of signal handlers relax …

sdroege:
…the type checks

Instead of requiring that the type of the Value itself has to be
compatible, we are checking for Object subclasses also if the type of
the contained object is compatible and if it is make sure that
everything works correctly.

This can happen for example if a button is stored in a gtk::Widget typed
variable and then converted to a glib::Value. The Value will have the
type of Widget and not Button, and type checks would fail.


sdroege:
CC @GuillaumeGomez, @EPashkin, @manishearth

GuillaumeGomez:
👍

PR #460, opened 2019-02-21
[release] merging crate-release-update into crate

GuillaumeGomez:
cc @GuillaumeGomez @EPashkin @sdroege


Cargo.toml
```diff
@@ -21,27 +21,25 @@ travis-ci = { repository = "gtk-rs/glib" }
 name = "glib"
 
 [dependencies]
-lazy_static = "1.0"
+lazy_static = "1.2"
```
sdroege: Why? Do we fail to compile with `1.0` now?

sdroege: Part of 68a03649d6ff9f91ef46f879d30d84bd374b36e8 from master. Has no justification but it's in master so will be fine ;)

CC @EPashkin

GuillaumeGomez: No clue. It's just a mirror of what we have on master after all.

sdroege:
:+1:

PR #463, opened 2019-02-22
Filename from uri fix

philn:


GuillaumeGomez:
👍

src/utils.rs
```diff
@@ -229,4 +229,17 @@ mod tests {
         check_setenv("Test");
         check_setenv("Тест"); // "Test" in Russian
     }
+
+    #[test]
+    fn test_filename_from_uri() {
+        use gstring::GString;
+        use std::path::PathBuf;
+        let uri: GString = "file:///foo/bar.txt".into();
```
sdroege: Maybe also add a test with a hostname for completeness?

philn: In which case can hostname be non-null? I tried an http uri but:
`Error { domain: g_convert_error, code: 4, message: "The URI “http://gtk-rs.org/index.html” is not an absolute URI using the “file” scheme" }`

sdroege: I don't know :) Maybe check the code of the C function?

philn: Yes the PR is updated now :)

sdroege:
Perfect :)

EPashkin:
Thanks 👍

PR #464, opened 2019-02-24
Regen

GuillaumeGomez:
cc @EPashkin @sdroege


EPashkin:
👍

src/auto/functions.rs
```diff
@@ -47,9 +47,7 @@ pub fn assertion_message_cmpstr(domain: &str, file: &str, line: i32, func: &str,
     }
 }
 
-pub fn assertion_message_expr<'a, 'b, P: Into<Option<&'a str>>, Q: Into<Option<&'b str>>>(domain: P, file: &str, line: i32, func: &str, expr: Q) {
-    let domain = domain.into();
-    let expr = expr.into();
+pub fn assertion_message_expr(domain: Option<&str>, file: &str, line: i32, func: &str, expr: Option<&str>) {
```
sdroege: Why do we even have this function in the bindings? Someone should review all the global functions we have generated here...

src/auto/functions.rs
```diff
@@ -1218,8 +1208,7 @@ pub fn test_set_nonfatal_assertions() {
     }
 }
 
-pub fn test_skip<'a, P: Into<Option<&'a str>>>(msg: P) {
-    let msg = msg.into();
+pub fn test_skip(msg: Option<&str>) {
```
sdroege: Same for this for example. We generate lots of functions that are useless in Rust, not usable as-is from Rust, or are simply wrong :)

src/value_array.rs
```diff
@@ -28,8 +28,7 @@ impl ValueArray {
         }
     }
 
-    pub fn append<'a, P: Into<Option<&'a Value>>>(&mut self, value: P) {
-        let value = value.into();
+    pub fn append(&mut self, value: Option<&Value>) {
```
sdroege: This should not be `nullable`. While the C API allows it, it is unsafe. Same for the other functions below

sdroege:
Looks good to me apart from the value array stuff. And please create an issue for reviewing the global functions here and removing most of them.

GuillaumeGomez:
I removed the two useless functions you pointed and removed the `Option` from the array function.

src/value_array.rs
```diff
@@ -28,7 +28,7 @@ impl ValueArray {
         }
     }
 
-    pub fn append(&mut self, value: Option<&Value>) {
+    pub fn append(&mut self, value: &Value) {
```
sdroege: Also for `insert` and `prepend` below

Gir.toml
```diff
@@ -93,6 +93,12 @@ status = "generate"
     #Quark
     ignore = true
     [[object.function]]
+    name = "assertion_message_expr"
+    ignore = true # function is useless
+    [[object.function]]
+    name = "test_skip"
```
sdroege: You could make this `pattern = "test_*"` I guess. The GLib unit tests API

GuillaumeGomez:
Updated.

sdroege:
:+1:

src/source.rs
```diff
@@ -270,7 +270,7 @@ where F: FnMut() -> Continue + 'static {
 /// process exits.
 ///
 /// `func` will be called when `pid` exits
-pub fn child_watch_add<'a, N: Into<Option<&'a str>>, F>(pid: Pid, func: F) -> SourceId
+pub fn child_watch_add<F>(pid: Pid, func: F) -> SourceId
```
EPashkin: Strange that there was unused constraint N but rust don't nag about it

EPashkin:
👍 after `insert` and `prepend` has option

GuillaumeGomez:
> 👍 after `insert` and `prepend` has option

I removed them on purpose after @sdroege asked for it so I'll keep it this way.

@sdroege: It's ready on my side so if there's nothing more missing, please merge. :)

sdroege:
> +1 after `insert` and `prepend` has option

@EPashkin what do you mean?

EPashkin:
Sorry, I made mistake Option need be removed and this already done.

EPashkin:
Lets merge?

GuillaumeGomez:
Yup!

PR #468, opened 2019-02-26
Fix some clippy warnings

EPashkin:
cc @GuillaumeGomez , @sdroege

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -82,6 +82,7 @@ impl Closure {
         from_glib_none(closure)
     }
 
+    #[allow(clippy::redundant_closure)]
     pub fn invoke(&self, values: &[&ToValue]) -> Option<Value> {
         let mut result = unsafe { Value::uninitialized() };
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -83,12 +83,6 @@ impl Error {
             })
         }
     }
-
-    // backcompat shim
-    #[cfg_attr(feature = "cargo-clippy", allow(not_unsafe_ptr_arg_deref))]
-    pub fn wrap(ptr: *mut glib_ffi::GError) -> Error {
-        unsafe { from_glib_full(ptr) }
-    }
 }
 
 impl fmt::Display for Error {
--- a/src/main_context.rs
+++ b/src/main_context.rs
@@ -4,7 +4,6 @@
 
 use ffi;
 use translate::*;
-use std::mem::transmute;
 use std::mem;
 use ffi as glib_ffi;
 use ffi::{gpointer, gboolean};
@@ -96,9 +95,8 @@ impl MainContext {
     }
 }
 
-#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
 unsafe extern "C" fn trampoline<F: FnOnce() + 'static>(func: gpointer) -> gboolean {
-    let func: &mut Option<F> = transmute(func);
+    let func: &mut Option<F> = &mut *(func as *mut Option<F>);
     let func = func.take().expect("MainContext::invoke() closure called multiple times");
     func();
     glib_ffi::G_SOURCE_REMOVE
--- a/src/object.rs
+++ b/src/object.rs
@@ -343,7 +343,7 @@ pub trait Cast: ObjectType {
         // same representation except for the name and the phantom data
         // type. IsA<> is an unsafe trait that must only be implemented
         // if this is a valid wrapper type
-        mem::transmute(self)
+	&*(self as *const Self as *const T)
     }
 }
 
@@ -532,7 +532,7 @@ macro_rules! glib_object_wrapper {
         #[doc(hidden)]
         impl $crate::translate::FromGlibPtrNone<*mut $ffi_name> for $name {
             #[inline]
-            #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+            #[allow(clippy::cast_ptr_alignment)]
             unsafe fn from_glib_none(ptr: *mut $ffi_name) -> Self {
                 debug_assert!($crate::types::instance_of::<Self>(ptr as *const _));
                 $name($crate::translate::from_glib_none(ptr as *mut _), ::std::marker::PhantomData)
@@ -542,7 +542,7 @@ macro_rules! glib_object_wrapper {
         #[doc(hidden)]
         impl $crate::translate::FromGlibPtrNone<*const $ffi_name> for $name {
             #[inline]
-            #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+            #[allow(clippy::cast_ptr_alignment)]
             unsafe fn from_glib_none(ptr: *const $ffi_name) -> Self {
                 debug_assert!($crate::types::instance_of::<Self>(ptr as *const _));
                 $name($crate::translate::from_glib_none(ptr as *mut _), ::std::marker::PhantomData)
@@ -552,7 +552,7 @@ macro_rules! glib_object_wrapper {
         #[doc(hidden)]
         impl $crate::translate::FromGlibPtrFull<*mut $ffi_name> for $name {
             #[inline]
-            #[cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+            #[allow(clippy::cast_ptr_alignment)]
             unsafe fn from_glib_full(ptr: *mut $ffi_name) -> Self {
                 debug_assert!($crate::types::instance_of::<Self>(ptr as *const _));
                 $name($crate::translate::from_glib_full(ptr as *mut _), ::std::marker::PhantomData)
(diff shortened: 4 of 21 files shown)


src/translate.rs
```diff
@@ -189,9 +189,7 @@ impl ToGlib for () {
     type GlibType = ();
 
     #[inline]
-    fn to_glib(&self) -> () {
```
GuillaumeGomez: Why the `/*-> ()*/`?

EPashkin: Because clippy not like returning unit both in declaration and body

EPashkin: to_glib usually return something so I decide just comment

sdroege:
Should probably also run clippy as part of CI for glib at least.

src/error.rs
```diff
@@ -85,7 +85,7 @@ impl Error {
     }
 
     // backcompat shim
-    #[cfg_attr(feature = "cargo-clippy", allow(not_unsafe_ptr_arg_deref))]
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::not_unsafe_ptr_arg_deref))]
```
sdroege: AFAIU this can be simply `#[clippy::blablabla]` now without the `cfg_attr` stuff?

src/main_context.rs
```diff
@@ -96,7 +96,7 @@ impl MainContext {
     }
 }
 
-#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
+#[cfg_attr(feature = "cargo-clippy", allow(clippy::transmute_ptr_to_ref))]
 unsafe extern "C" fn trampoline<F: FnOnce() + 'static>(func: gpointer) -> gboolean {
     let func: &mut Option<F> = transmute(func);
```
sdroege: This could simply be a cast: `let func: &mut Option<F> = &*(func as *mut Option<F>)`

src/object.rs
```diff
@@ -337,6 +337,7 @@ pub trait Cast: ObjectType {
     /// Casts to `&T` unconditionally.
     ///
     /// Panics if compiled with `debug_assertions` and the instance doesn't implement `T`.
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::transmute_ptr_to_ptr))]
```
sdroege: Here we should be able to cast too then, or not?

src/source.rs
```diff
@@ -98,7 +98,7 @@ impl Drop for CallbackGuard {
     }
 }
 
-#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
+#[cfg_attr(feature = "cargo-clippy", allow(clippy::transmute_ptr_to_ref))]
 unsafe extern "C" fn trampoline<F: FnMut() -> Continue + 'static>(func: gpointer) -> gboolean {
     let func: &RefCell<F> = transmute(func);
```
sdroege: Can be a simple cast

src/source.rs
```diff
@@ -113,7 +113,7 @@ fn into_raw<F: FnMut() -> Continue + 'static>(func: F) -> gpointer {
     Box::into_raw(func) as gpointer
 }
 
-#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
+#[cfg_attr(feature = "cargo-clippy", allow(clippy::transmute_ptr_to_ref))]
 unsafe extern "C" fn trampoline_child_watch<F: FnMut(Pid, i32) + 'static>(pid: glib_ffi::GPid, status: i32, func: gpointer) {
     let func: &RefCell<F> = transmute(func);
```
sdroege: Can be a simple cast

src/subclass/object.rs
```diff
@@ -296,6 +296,7 @@ pub unsafe trait ObjectClassSubclassExt: Sized + 'static {
 unsafe impl ObjectClassSubclassExt for ObjectClass {}
 
 unsafe impl<T: ObjectSubclass> IsSubclassable<T> for ObjectClass {
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::cast_ref_to_mut))]
     fn override_vfuncs(&mut self) {
         unsafe {
             let klass = &mut *(self as *const Self as *mut gobject_ffi::GObjectClass);
```
sdroege: Why do we even cast to a `*const Self`? Just make this a `*mut Self` and it will all be fine :)

src/value.rs
```diff
@@ -138,6 +138,7 @@ impl Value {
     ///
     /// Returns `Some(&TypedValue<T>)` if the value carries a type corresponding
     /// to `T` and `None` otherwise.
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::transmute_ptr_to_ptr))]
```
sdroege: Should be possible to use a cast here and below?

src/translate.rs
sdroege: Fine with me :)

sdroege: Why do we need this impl though?

src/error.rs
EPashkin: IMHO 1.31.0 still don't support it, will check at evening.
Thanks for reminder

src/object.rs
EPashkin: Maybe yes, now not sure why I decide ignoring.

src/translate.rs
EPashkin: Examples compiles without it, so it will be removed

EPashkin:
Updated

src/error.rs
```diff
@@ -85,7 +85,7 @@ impl Error {
     }
 
     // backcompat shim
-    #[cfg_attr(feature = "cargo-clippy", allow(not_unsafe_ptr_arg_deref))]
+    #[allow(clippy::not_unsafe_ptr_arg_deref)]
```
sdroege: We should probably remove this function? It says backwards compatibility :)

src/closure.rs
```diff
@@ -82,6 +82,7 @@ impl Closure {
         from_glib_none(closure)
     }
 
+    #[allow(clippy::redundant_closure)]
     pub fn invoke(&self, values: &[&ToValue]) -> Option<Value> {
```
sdroege: What does this mean exactly, what redudant closure and why can't we fix that?

sdroege:
Looks good to me otherwise, thanks! Can we run clippy as part of CI now or are there still warnings left?

src/error.rs
EPashkin: Yes, it not used

src/closure.rs
EPashkin: It wan't change `.map(|v| v.to_value())` to function but if I change it, then will be error.
```
error[E0277]: the trait bound `dyn value::ToValue: value::SetValue` is not satisfied
  --> D:/eap/rust/0/glib\src\closure.rs:98:22
   |
98 |                 .map(ToValue::to_value)
   |                      ^^^^^^^^^^^^^^^^^ the trait `value::SetValue` is not implemented for `dyn value::ToValue`
   |
   = note: required because of the requirements on the impl of `value::SetValue` for `&dyn value::ToValue`
   = note: required because of the requirements on the impl of `value::ToValue` for `&dyn value::ToValue`
note: required by `value::ToValue::to_value`
  --> D:/eap/rust/0/glib\src\value.rs:594:5
   |
594|     fn to_value(&self) -> Value;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Same problem for second `redundant_closure`
```
error[E0277]: the trait bound `std::vec::Vec<std::string::String>: translate::ToGlibPtr<'_, _>` is not satisfied
   --> D:/eap/rust/0/glib\src\value.rs:871:46
    |
871 |         let ptr: *mut *mut c_char = this.map(ToGlibPtr::to_glib_full).unwrap_or(ptr::null_mut());
    |                                              ^^^^^^^^^^^^^^^^^^^^^^^ the trait `translate::ToGlibPtr<'_, _>` is not implemented for `std::vec::Vec<std::string::String>`
    |
note: required by `translate::ToGlibPtr::to_glib_full`
   --> D:/eap/rust/0/glib\src\translate.rs:259:5
    |
259 |     fn to_glib_full(&self) -> P {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0277]: the trait bound `std::vec::Vec<std::string::String>: translate::ToGlibPtr<'_, _>` is not satisfied
   --> D:/eap/rust/0/glib\src\value.rs:871:37
    |
871 |         let ptr: *mut *mut c_char = this.map(ToGlibPtr::to_glib_full).unwrap_or(ptr::null_mut());
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `translate::ToGlibPtr<'_, _>` is not implemented for `std::vec::Vec<std::string::String>`
```
```

EPashkin:
Added clippy check on travis

src/closure.rs
sdroege: Does this need a `as_ref()` maybe? But fine with me

sdroege:
> Added clippy check on travis

Please use stable clippy

EPashkin:
I prefer use last version, you against "beta" too ?

sdroege:
> I prefer use last version, you against "beta" too ?

The problem is that nightly (and beta?) is often released without clippy or other tools because of breakages and that would then make the CI fail for no good reason. Otherwise I'd also prefer nightly to have the latest checks.

EPashkin:
As beta is next stable so it must have clippy, I try change to it

src/closure.rs
EPashkin: Not works too.

sdroege: :man_shrugging:

EPashkin:
Maybe we need test both beta and nightly but change script that it don't fails on missing component?

sdroege:
> Maybe we need test both beta and nightly but change script that it don't fails on missing component?

I don't think that's worth it. clippy is not changing that often :) Let's merge this as is?

EPashkin:
Yes, we can't fix failing nightly clippy, so lets merge current version if @GuillaumeGomez don't have better solution.

GuillaumeGomez:
Like I said, I'm fine as is.

PR #470, opened 2019-03-05
Remove xx_sys to xx_ffi renaming

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/685

cc @GuillaumeGomez, @sdroege


GuillaumeGomez:
👍

src/auto/checksum.rs
```diff
@@ -3,43 +3,43 @@
 // DO NOT EDIT
 
 use ChecksumType;
-use ffi;
+use sys;
```
sdroege: If we're changing this anyway, shouldn't we directly name this `glib_sys` instead? That way there's no renaming necessary at all even for the "local" FFI bindings and it would be more consistent.

EPashkin: Currently I think that all concrete classes better be almost same as generated code (use `sys` for current FFI-crate).
In glib has 2 main parts: currently glib-based object has sys+gobject_sys,
only gobject-based classes has glib_sys+gobject_sys to minimize confusion.
If we want use `xx_sys` instead `sys` in generated code too, it maybe need loot changes in gir.

sdroege: I'd prefer to go without `sys` and rename it all to `xxx_sys`.

As with Rust 2018 the `extern crate` are not required anymore, it would be really nice if we wouldn't have to do renames. It simplifies things a lot :)

EPashkin: Ok I try change this too.

sdroege: Great, thanks :)

EPashkin:
Now always glib_sys/gobject_sys

sdroege:
> Now always glib_sys/gobject_sys

Awesome. All good for me then :)

sdroege:
Why is this WIP? Good to go in any case with the latest version of the gir PR.

EPashkin:
All regen PR is WIP as it don't have right gir, I just forgot addd WIP to gtk

sdroege:
> All regen PR is WIP as it don't have right gir, I just forgot addd WIP to gtk

gir PR is merged now, so let's get this updated and merged? :)

EPashkin:
I will update all PRs today after https://github.com/gtk-rs/gir/pull/740

EPashkin:
Updated, ready for final check

GuillaumeGomez:
Still 👍

EPashkin:
Fixed tests

sdroege:
Looks good to me

EPashkin:
CI "passed"

GuillaumeGomez:
Thanks!

PR #476, opened 2019-04-26
Port to futures 0.3

sdroege:
While this requires nightly, the API is now stabilized and will become part of Rust 1.36. IMHO we should merge this now so that it can all get wider testing.

### DIFF
--- a/src/main_context_futures.rs
+++ b/src/main_context_futures.rs
@@ -3,16 +3,18 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use futures;
-use futures::executor::{Executor, SpawnError};
+use futures::future::{FutureObj, LocalFutureObj};
 use futures::prelude::*;
-use futures::task::{LocalMap, UnsafeWake, Waker};
-use futures::{Async, Future, Never};
+use futures::task::{
+    Context, LocalSpawn, Poll, RawWaker, RawWakerVTable, Spawn, SpawnError, Waker,
+};
 use get_thread_id;
 use glib_sys;
 use std::mem;
 use std::ptr;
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Looks good to me, thanks! And yes, the sooner the better.

sdroege:
Great, thanks. Waiting for @EPashkin's opinion and the "OK" for all the other PRs and then let's get merging

sdroege:
> and then let's get merging

Note that I have to update all PRs before merging apart from glib and gir once those two are merged.

EPashkin:
@sdroege Big thanks.
As we have some fixes for previous release maybe better do release first?
Then we have 3 month until stable 1.36 for testing.

sdroege:
The idea would be to include it in this release already so people actually start testing. Almost nobody is using git :)

EPashkin:
Ok

sdroege:
Ok, then I'll merge this and update the other ones in a bit.

src/main_context_futures.rs
```diff
@@ -283,17 +314,17 @@ impl MainContext {
...
+                future::ready(())
             });
 
-            let f: *mut Future<Item = (), Error = Never> = Box::into_raw(Box::new(f));
-            // XXX: Transmute to get a 'static lifetime here, super unsafe
-            let f: *mut (Future<Item = (), Error = Never> + 'static) = mem::transmute(f);
-            let f: Box<Future<Item = (), Error = Never> + 'static> = Box::from_raw(f);
+            // Super-unsafe: We transmute here to get rid of the 'static lifetime
+            let f = LocalFutureObj::new(Box::new(f));
+            let f: (LocalFutureObj<'static, ()>) = mem::transmute(f);
```
EPashkin: IMHO comment now has wrong part "get rid" instead just "get"

PR #477, opened 2019-04-26
Run futures code and tests when building with nightly

sdroege:


EPashkin:
Seems this need be after examples

sdroege:
No, this fails for other reasons. Will investigate.

The examples might still fail until all PRs are merged though.

sdroege:
Yeah, silly :) Because it was stabilized now I need to remove the `#![feature]` thingy.

GuillaumeGomez:
It builds but fails because examples don't have `futures` feature.

sdroege:
Yeah, I misread. So this is fine and we need to go through the crates one by one as usual and do the examples as the last one.

sdroege:
Once https://github.com/gtk-rs/examples/pull/240 is in, this one here should also succeed.

sdroege:
@GuillaumeGomez travis is green :)

GuillaumeGomez:
Thanks!

PR #479, opened 2019-05-09
add manual bindings for GRecMutex

gdesmott:
@sdroege : I'll need this for https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/issues/161

Thanks to @GuillaumeGomez for helping me implementing this during the hackfest. :)

### DIFF
--- a/src/rec_mutex.rs
+++ b/src/rec_mutex.rs
@@ -1,0 +1,118 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use glib_sys;
+use std::mem;
+use translate::{from_glib, Stash, ToGlibPtr};
+
+#[derive(Debug)]
+pub enum RecMutex<'a> {
+    Owned(glib_sys::GRecMutex),
+    Borrowed(&'a glib_sys::GRecMutex),
+}
+
+impl<'a> RecMutex<'a> {
+    pub fn new() -> Self {
+        let rec_mutex = unsafe {
+            let mut mutex = mem::zeroed();
+            glib_sys::g_rec_mutex_init(&mut mutex);
+            mutex
+        };
+        RecMutex::Owned(rec_mutex)
+    }
+
+    #[doc(hidden)]
+    pub unsafe fn borrow(rec_mutex: &'a glib_sys::GRecMutex) -> Self {
+        RecMutex::Borrowed(rec_mutex)
+    }
+
+    pub fn lock(&self) -> RecMutexGuard {
+        unsafe {
+            glib_sys::g_rec_mutex_lock(self.as_ptr());
+        }
+        RecMutexGuard { rec_mutex: self }
+    }
+
+    pub fn try_lock(&self) -> Option<RecMutexGuard> {
+        let locked = unsafe { from_glib(glib_sys::g_rec_mutex_trylock(self.as_ptr())) };
+
+        if locked {
+            Some(RecMutexGuard { rec_mutex: self })
+        } else {
+            None
+        }
+    }
+
+    #[doc(hidden)]
+    fn as_ptr(&self) -> *mut glib_sys::GRecMutex {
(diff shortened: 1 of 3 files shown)


src/rec_mutex.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn new() -> Self {
+        let rec_mutex = unsafe {
+            let mut mutex = mem::zeroed();
+            glib_sys::g_rec_mutex_init(&mut mutex);
+            mutex
+        };
+        RecMutex::Owned(rec_mutex)
+    }
+
+    pub fn from_glib_none(rec_mutex: &'a glib_sys::GRecMutex) -> Self {
```
GuillaumeGomez: Maybe add a `#[doc(hidden)]` in here.

src/rec_mutex.rs
```diff
@@ -0,0 +1,117 @@
...
+        let locked = unsafe { from_glib(glib_sys::g_rec_mutex_trylock(self.as_mut_ptr())) };
+
+        if locked {
+            Some(RecMutexGuard { rec_mutex: self })
+        } else {
+            None
+        }
+    }
+
+    fn as_mut_ptr(&self) -> *mut glib_sys::GRecMutex {
```
GuillaumeGomez: Maybe add a `#[doc(hidden)]` in here.

GuillaumeGomez:
The implementation looks good to me. Just a few things we should hide to users but that's it. Good work!

sdroege:
If we decide to add these, I'd recommend first of all adding the normal mutex. I don't think it's a good idea to provide these in the bindings as user-facing API though. If anything it should be used internally for interaction with existing C code. The mutexes/etc in the standard library are better.

@gdesmott see also how I used `GMutex` and `GRecMutex` in `gstreamer-rs` internally in a few places already.

src/rec_mutex.rs
```diff
@@ -0,0 +1,119 @@
...
+        let rec_mutex = unsafe {
+            let mut mutex = mem::zeroed();
+            glib_sys::g_rec_mutex_init(&mut mutex);
+            mutex
+        };
+        RecMutex::Owned(rec_mutex)
+    }
+
+    #[doc(hidden)]
+    pub fn from_glib_none(rec_mutex: &'a glib_sys::GRecMutex) -> Self {
```
sdroege: This is not really a `from_glib_none()`. It does not give us an owned version of the thing, like all other `from_glib_none` do.

Maybe instead let's call it just `borrow`? Also unclear if this shouldn't be `unsafe`. If you initialize ` GRecMutex` with random data then this will probably cause crashes, and you can safely initialize a `GRecMutex` struct with random data.

In `gstreamer-rs` I have such things as `unsafe fn from_ptr<'a>(ptr: *const X) -> &'a Y`.

src/rec_mutex.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        if locked {
+            Some(RecMutexGuard { rec_mutex: self })
+        } else {
+            None
+        }
+    }
+
+    #[doc(hidden)]
+    fn as_mut_ptr(&self) -> *mut glib_sys::GRecMutex {
```
sdroege: Also `as_ptr()` for consistency?

sdroege:
> The mutexes/etc in the standard library are better.

To extend on this, the Rust model is to protect data and not code. The Rust mutexes are containers around data, while the GLib ones are only protecting code.

gdesmott:
I no longer need this so feel free to WONTFIX if you think that's not worst adding to glib-rs.
If not I'll be happy to write similar bindings for `GMutex`.

gdesmott:
If we actually do want this, @GuillaumeGomez suggested another `lock` API taking a `FnOnce` which will be called right away after locking the mutex. This would match better the "mutex protecting code" semantic you described.

sdroege:
I don't think it would be a good idea to merge this as its behaviour is less safe and different to how Rust mutexes work.

gdesmott:
It's up to you and @GuillaumeGomez, I don't really care as a I no longer need it for my bindings. Feel free to close.

gdesmott:
Let's close this for now, we can easily resurrect it later if needed.

PR #480, opened 2019-05-11
Add ByteArray binding

elmarco:
Not exactly the most useful thing, but an interesting exercise during
Rust+GNOME hackfest :)

(The gir definitions were a bit too weird to auto-generate a satisfying binding)

Signed-off-by: Marc-André Lureau <marcandre.lureau@redhat.com>

### DIFF
--- a/src/byte_array.rs
+++ b/src/byte_array.rs
@@ -1,0 +1,265 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+//! # Examples
+//!
+//! ```
+//! use glib::prelude::*; // or `use gtk::prelude::*;`
+//! use glib::ByteArray;
+//!
+//! let ba = ByteArray::from(b"def");
+//! ba.append(b"ghi").prepend(b"abc");
+//! ba.remove_range(3, 3);
+//! assert_eq!(ba, "abcghi".as_bytes());
+//! ```
+
+use glib_sys;
+use std::borrow::Borrow;
+use std::cmp::Ordering;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+use std::mem;
+use std::ops::Deref;
+use std::slice;
+use translate::*;
+use std::ptr::NonNull;
+
+use Bytes;
+
+glib_wrapper! {
+    pub struct ByteArray(Shared<glib_sys::GByteArray>);
+
+    match fn {
+        ref => |ptr| glib_sys::g_byte_array_ref(ptr),
+        unref => |ptr| glib_sys::g_byte_array_unref(ptr),
+        get_type => || glib_sys::g_byte_array_get_type(),
+    }
+}
+
+impl ByteArray {
+    pub fn new() -> ByteArray {
+        unsafe { from_glib_full(glib_sys::g_byte_array_new()) }
+    }
+
+    pub fn with_capacity(size: usize) -> ByteArray {
+        unsafe { from_glib_full(glib_sys::g_byte_array_sized_new(size as u32)) }
+    }
+
+    pub fn into_gbytes(self) -> Bytes {
+        unsafe {
+            let ret = from_glib_full(glib_sys::g_byte_array_free_to_bytes(mut_override(
+                self.to_glib_none().0,
+            )));
+            mem::forget(self);
+            ret
+        }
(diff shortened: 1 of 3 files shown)


src/byte_array.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+impl Default for ByteArray {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+unsafe impl Send for ByteArray {}
+unsafe impl Sync for ByteArray {}
```
GuillaumeGomez: `ByteArray` is a thread-safe type?

elmarco: as much as GString, which also impl Send/Sync.

but only refcounting seems safe

sdroege: It would only be Send/Sync if it used normal Rust mutability rules and no reference counting

src/byte_array.rs
```diff
@@ -0,0 +1,264 @@
...
+        unsafe { from_glib_full(glib_sys::g_byte_array_new()) }
+    }
+
+    pub fn with_capacity(size: usize) -> ByteArray {
+        unsafe { from_glib_full(glib_sys::g_byte_array_sized_new(size as u32)) }
+    }
+
+    pub fn into_gbytes(self) -> Bytes {
+        unsafe {
+            let ret = from_glib_full(glib_sys::g_byte_array_free_to_bytes(mut_override(
```
sdroege: Mixing `free()` and refcounting seems... suboptimal? Is `free` just calling `unref()`?

src/byte_array.rs
```diff
@@ -0,0 +1,264 @@
...
+            glib_sys::g_byte_array_append(
+                self.to_glib_none().0,
+                bytes.as_ptr() as *const _,
+                bytes.len() as u32,
+            );
+        }
+        self
+    }
+
+    pub fn prepend<T: AsRef<[u8]>>(&self, data: T) -> &Self {
```
sdroege: You could here (and above) take the data by reference instead. There's no need to take ownership of it here as the byte array will copy the data

src/byte_array.rs
elmarco: I should drop it then?

src/byte_array.rs
```diff
@@ -0,0 +1,264 @@
...
+
+impl Deref for ByteArray {
+    type Target = [u8];
+
+    fn deref(&self) -> &[u8] {
+        unsafe {
+            let ptr = (*self.to_glib_none().0).data;
+            let len = (*self.to_glib_none().0).len as usize;
+            debug_assert!(!ptr.is_null() || len == 0);
+            slice::from_raw_parts(ptr as *const u8, len)
```
sdroege: This is UB, the pointer must never be null. If the length is 0, pass in mem::align_of::<u8>() or so :)

src/byte_array.rs
sdroege: Yes, it's wrong currently

src/byte_array.rs
elmarco: g_byte_array_free*() is doing somewhat unconventional things. If there are > 1 refs, it frees the internal buffer, not the GArray "wrapper" (see PRESERVE_WRAPPER, https://gitlab.gnome.org/GNOME/glib/blob/master/glib/garray.c#L357)

src/byte_array.rs
elmarco: ok, removed

src/byte_array.rs
sdroege: And then what happens if you use the wrapper? It would start with a new internal buffer?

elmarco: Yes, it does g_realloc() internally which alloc when given NULL.

src/byte_array.rs
elmarco: no idea what you mean here :) from_raw_parts(NULL, 0) wouldn't work as expected?

src/byte_array.rs
sdroege: Seems fine then I guess :)

src/byte_array.rs
sdroege: See https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html
```
data must be non-null and aligned, even for zero-length slices.
```

elmarco: Something like that?
```
+            if ptr.is_null() {
+                ptr = NonNull::dangling().as_ptr();
+            }
```

sdroege: Yeah

PR #482, opened 2019-05-12
Add Array<T> & ArrayPtr

elmarco:


src/translate.rs
```diff
@@ -921,9 +921,19 @@ where T: GlibPtrDefault + ToGlibPtr<'a, <T as GlibPtrDefault>::GlibType> {
 
 pub struct Array(*mut glib_sys::GArray);
 
+impl Array {
+    fn new(zero_terminated: bool, clear: bool, element_size: usize) -> Self {
```
GuillaumeGomez: If you store `T` in `Array` directly (through `PhantomData`), `element_size` becomes useless as argument.

src/translate.rs
```diff
@@ -921,9 +921,19 @@ where T: GlibPtrDefault + ToGlibPtr<'a, <T as GlibPtrDefault>::GlibType> {
...
+                                        clear.to_glib(),
+                                        element_size as u32))
+        }
+    }
+}
+
 impl Drop for Array {
     fn drop(&mut self) {
-        unsafe { glib_sys::g_array_free(self.0, false.to_glib()); }
+        unsafe { glib_sys::g_array_free(self.0, true.to_glib()); }
```
sdroege: Nice one

sdroege:
Thanks, looks good to me but we could do the element size nicer indeed

elmarco:
I am not quite happy with the current `impl<'a, T> ToGlibContainerFromSlice<'a, *mut glib_sys::GArray> for T`

I think it should not build a `GArray<T*>`, but a `GArray<T>`.

So that [T] -> `GArray<T>`
And [&T] -> `GArray<T*>`

Does that make sense?

elmarco:
I think this is getting a bit too convoluted. GList/GSList are simpler, GPtrArray is too, as they all use pointers to elements.

Maybe we shouldn't implement ToGlibContainerFromSlice for GArray after all. And only let the caller correctly allocate/create a GArray, with more wrapped functions (set_size/append/index...)

sdroege:
> And [&T] -> `GArray<T*>`

That would be unsafe as there's nothing that would enforce the lifetime rules, especially with arrays being refcounted.

> Maybe we shouldn't implement ToGlibContainerFromSlice for GArray after all. And only let the caller correctly allocate/create a GArray, with more wrapped functions (set_size/append/index...)

Yes, similar to the `ByteArray` API you added in another PR. That would make sense and allow for implementing special cases that we can't automatically handle manually.

Nonetheless we should be able to handle this somehow. It's not really different from the `Vec<_>` / `&[_]` situation. Based on the type, either pointers or flat values are stored. It should be possible to handle `GArray` the same way.

sdroege:
https://github.com/gtk-rs/gir/issues/596 is also related to this

sdroege:
@elmarco What's the status of this?

elmarco:
@sdroege it could use some help & reviews. The travis failure looks unrelated.

sdroege:
Those failures should be gone when retrying with the latest version now (can you rebase?). What other help do you need, what is missing here? :)

majorz:
The NetworkManager client library makes heavy usage of GPtrArray, but it seems that it needs to be parametrized as well - PtrArray<T>.

An example here:
```
<array name="GLib.PtrArray" c:type="GPtrArray*">
  <type name="Connection"/>
</array>
```

I've been experimenting with it and hopefully I will PR it once I have it working (here and the generation part on the gir repo).

sdroege:
> I've been experimenting with it and hopefully I will PR it once I have it working (here and the generation part on the gir repo).

Instead of proper bindings for it you could also go with the same approach that is currently done for `GList`, etc. and have it convert between slices/`Vec` at the FFI boundary.

majorz:
Thanks for the advice. That is closer to the manual solution I initially did. I will start looking into the `GList` approach.

sdroege:
Check `translate.rs` for `GList`. It's mostly a matter of implementing some traits, and then ideally also teaching `gir` about autogenerating that code.

majorz:
That worked really well! I already looked at the gir sources for a few days last week and now I got quick results just by redoing the same steps as for GList and GSList. Since I did it all by copy/pasting, I will need more time to figure out some of the details. Also I need to verify things are actually working properly after compilation, but at least the code looks right and compiles. I hope to be ready with the PRs in a few days.

majorz:
Finally did all the testing with the NetworkManager client library - works like a charm. I extracted the PtrArray part from this PR and added it in #559.

sdroege:
Let's close this one for now then.

Issue #483, opened 2019-05-12
Object::new() bug


antoyo:
Hi.
The following code:
```rust
let window: gtk::Window = glib::Object::new(gtk::Window::static_type(), &[("type", &WindowType::Toplevel)])
    .unwrap().downcast().unwrap();
```
triggers the following errors at runtime:
```
(.:26952): GLib-GObject-CRITICAL **: 10:05:57.685: g_object_get_qdata: assertion 'G_IS_OBJECT (object)' failed

(.:26952): GLib-GObject-WARNING **: 10:05:57.685: invalid unclassed pointer in cast to 'GtkWindow'

(.:26952): Gtk-CRITICAL **: 10:05:57.685: gtk_window_get_type_hint: assertion 'GTK_IS_WINDOW (window)' failed

(.:26952): GLib-GObject-CRITICAL **: 10:05:57.685: g_object_get_qdata: assertion 'G_IS_OBJECT (object)' failed

(.:26952): GLib-GObject-WARNING **: 10:05:57.686: invalid unclassed pointer in cast to 'GtkWindow'

(.:26952): Gtk-CRITICAL **: 10:05:57.686: gtk_window_get_decorated: assertion 'GTK_IS_WINDOW (window)' failed

(.:26952): GLib-GObject-WARNING **: 10:05:57.686: invalid unclassed pointer in cast to 'GtkWidget'

(.:26952): Gtk-CRITICAL **: 10:05:57.686: gtk_widget_get_toplevel: assertion 'GTK_IS_WIDGET (widget)' failed

(.:26952): GLib-GObject-CRITICAL **: 10:05:57.686: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```
while the following code works:
```rust
use std::ffi::CString;
use glib::translate::*;
let type_property = CString::new("type").unwrap();
let typ = WindowType::Toplevel.to_glib();
let null: *mut gobject_sys::GObject = std::ptr::null_mut();
let window: gtk::Window =
    unsafe {
        gtk::Widget::from_glib_none(gobject_sys::g_object_new(
            gtk::Window::static_type().to_glib(),
            type_property.as_ptr(), typ,
            null) as *mut _
        ).downcast().expect("downcast")
    };
```
Is the first code wrong or is it a bug in `Object::new()`?
Thanks.

sdroege:
see https://github.com/gtk-rs/glib/pull/484

PR #484, opened 2019-05-12
Always ref-sink after g_object_new() if the type inherits from GIniti…

sdroege:
…allyUnowned

Independent of it still being actually floating or not.

Fixes https://github.com/gtk-rs/glib/issues/483

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -882,7 +882,7 @@ impl Object {
             let ptr = gobject_sys::g_object_newv(type_.to_glib(), params_c.len() as u32, mut_override(params_c.as_ptr()));
             if ptr.is_null() {
                 Err(glib_bool_error!("Can't instantiate object"))
-            } else if gobject_sys::g_object_is_floating(ptr) != glib_sys::GFALSE {
+            } else if type_.is_a(&InitiallyUnowned::static_type()) {
                 Ok(from_glib_none(ptr))
             } else {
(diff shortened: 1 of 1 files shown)


src/object.rs
```diff
@@ -882,7 +882,7 @@ impl Object {
             let ptr = gobject_sys::g_object_newv(type_.to_glib(), params_c.len() as u32, mut_override(params_c.as_ptr()));
             if ptr.is_null() {
                 Err(glib_bool_error!("Can't instantiate object"))
-            } else if gobject_sys::g_object_is_floating(ptr) != glib_sys::GFALSE {
+            } else if gobject_sys::g_type_is_a(type_.to_glib(), gobject_sys::g_initially_unowned_get_type()) != glib_sys::GFALSE {
```
antoyo: Could we do something like:
```rust
type::is_a::<glib::object::InitiallyUnowned>()
```
?

sdroege: Yes, does that work?

antoyo: I don't know. I can try it if you want.

antoyo: The following works:
```rust
type_.is_a(&glib::object::InitiallyUnowned::static_type())
```

sdroege: Thanks, updated :)

sdroege:
Will merge once the CI is through

PR #485, opened 2019-05-18
Adds Copy and Clone traits to glib::SignalHandlerId.

xfbs:
I think SignalHandlerId should derive Copy and Clone, since it's just a scalar.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -12,7 +12,7 @@ use std::mem;
 use translate::{from_glib, FromGlib, ToGlib, ToGlibPtr};
 
 /// The id of a signal that is returned by `connect`.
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
From my point of view, we don't want to allow users to manipulate a same signal handler id multiple times at once. But maybe I'm wrong?

cc @sdroege @EPashkin

EPashkin:
Yes, it designed without Copy/Clone specially for that reason

sdroege:
Yes, there should be no Clone/Copy impl for that reason. It's intentionally done like this.

What problems does it cause you right now and in which context?

sdroege:
Closing as it's as intended and there was no response after a month

PR #488, opened 2019-06-03
subclass: panic with error message when registering existing type

gdesmott:
The assertion wasn't very user friendly, it deserves some explanations.

### DIFF
--- a/src/subclass/boxed.rs
+++ b/src/subclass/boxed.rs
@@ -45,10 +45,9 @@ pub fn register_boxed_type<T: BoxedType>() -> ::Type {
         use std::ffi::CString;
 
         let type_name = CString::new(T::NAME).unwrap();
-        assert_eq!(
-            gobject_sys::g_type_from_name(type_name.as_ptr()),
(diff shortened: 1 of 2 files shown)


src/subclass/boxed.rs
```diff
@@ -45,10 +45,9 @@ pub fn register_boxed_type<T: BoxedType>() -> ::Type {
         use std::ffi::CString;
 
         let type_name = CString::new(T::NAME).unwrap();
-        assert_eq!(
-            gobject_sys::g_type_from_name(type_name.as_ptr()),
-            gobject_sys::G_TYPE_INVALID
-        );
+        if gobject_sys::g_type_from_name(type_name.as_ptr()) != gobject_sys::G_TYPE_INVALID {
+            panic!("Type {} has already be registered", type_name.to_str().unwrap());
```
sdroege: s/be/been/ :) Also in the other place below

gdesmott: That's what you get with the first patch in the morning ;)

Issue #469, opened 2019-03-05
Boxed AnyBox::Native potentially causes memory leaks


sdroege:
See [here](https://github.com/gtk-rs/glib/blob/92b365b51f84d9d32ce36f6087a53ebcd261ecea/src/boxed.rs#L268) what I'm talking about.

This variant is used whenever `T::uninitialized()` is called, which gir does automatically for boxed types if there's an out parameter that is `caller allocates`.

If such a boxed struct is then dropped, nothing will happen according to [the code](https://github.com/gtk-rs/glib/blob/92b365b51f84d9d32ce36f6087a53ebcd261ecea/src/boxed.rs#L406-L408). So if the FFI struct contains pointers that have to be freed, for example, then we would be leaking them.

I didn't check any code yet if we already have such a case, but it seems likely and is obviously wrong.

I'm not sure yet how to implement the same thing correctly.

sdroege:
Yes, `gst_video::VideoTimeCode` if it was not manually implemented for other reasons anyway, `glib::String`, `glib::Error`, `glib::ValueArray`.

EPashkin:
@sdroege You sure that any boxed type contains pointer? `gst_video::VideoTimeCode` IMHO not

EPashkin:
As I know `free` only freed allocated memory, out parameters usually filled in stack and I don't see way in glib to "free" it values

sdroege:
> @sdroege You sure that any boxed type contains pointer? `gst_video::VideoTimeCode` IMHO not

See [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs-sys/blob/e274b5495bb1698cd9fe48de40ab5f9f2b0e1faa/gstreamer-video-sys/src/lib.rs#L1413), which is defined [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs-sys/blob/e274b5495bb1698cd9fe48de40ab5f9f2b0e1faa/gstreamer-video-sys/src/lib.rs#L1434-1441) and has a pointer in there. That pointer is owned and needs to be freed later.

Usually for such types there's an `init()` and `clear()` or `reset()` or `unset()` function.

> As I know `free` only freed allocated memory, out parameters usually filled in stack

Search for API that has an out parameter that is annotated as `caller allocates`. `glib::Error` is not such a case (`callee allocates`), but `glib::Value` is for example (exactly same problem btw but we don't have leaks because we manually implement it and the correct `Drop` impl).

[...]

PR #490, opened 2019-06-05
Remove uninitialized() function for Boxed types

sdroege:
These would currently allocate a Box<_> for the FFI struct and would be
used for boxed type out function parameters, where the caller is
supposed to allocate the memory.

The problem with these is that if the FFI struct requires a function to
free memory stored inside it, we have no way of freeing it currently and
would leak the memory.

As such remove the support for this until it is actually needed, and
then we can figure out a solution based on the concrete case.

Fixes https://github.com/gtk-rs/glib/issues/469

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -73,14 +73,6 @@ macro_rules! glib_boxed_wrapper {
         }
 
         #[doc(hidden)]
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
We have a lot of failures coming from the generated code. A gir update will be needed.

PR #493, opened 2019-06-09
Move trampolines

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/src/byte_array.rs
+++ b/src/byte_array.rs
@@ -105,6 +105,22 @@ impl ByteArray {
     }
 
     pub fn sort<F: FnMut(&u8, &u8) -> Ordering>(&self, compare_func: F) {
+        unsafe extern "C" fn compare_func_trampoline(
+            a: glib_sys::gconstpointer,
+            b: glib_sys::gconstpointer,
+            func: glib_sys::gpointer,
+        ) -> i32 {
+            let func = func as *mut &mut (dyn FnMut(&u8, &u8) -> Ordering);
+
+            let a = &*(a as *const u8);
+            let b = &*(b as *const u8);
+
+            match (*func)(&a, &b) {
+                Ordering::Less => -1,
+                Ordering::Equal => 0,
+                Ordering::Greater => 1,
+            }
+        }
         unsafe {
             let mut func = compare_func;
             let func_obj: &mut (dyn FnMut(&u8, &u8) -> Ordering) = &mut func;
@@ -117,23 +133,6 @@ impl ByteArray {
                 func_ptr,
             );
         }
-    }
-}
-
-unsafe extern "C" fn compare_func_trampoline(
-    a: glib_sys::gconstpointer,
-    b: glib_sys::gconstpointer,
-    func: glib_sys::gpointer,
-) -> i32 {
-    let func = func as *mut &mut (dyn FnMut(&u8, &u8) -> Ordering);
-
-    let a = &*(a as *const u8);
-    let b = &*(b as *const u8);
-
-    match (*func)(&a, &b) {
-        Ordering::Less => -1,
-        Ordering::Equal => 0,
-        Ordering::Greater => 1,
     }
 }
 
--- a/src/main_context.rs
+++ b/src/main_context.rs
@@ -88,7 +88,9 @@ impl MainContext {
     {
         unsafe extern "C" fn trampoline<F: FnOnce() + 'static>(func: gpointer) -> gboolean {
             let func: &mut Option<F> = &mut *(func as *mut Option<F>);
-            let func = func.take().expect("MainContext::invoke() closure called multiple times");
+            let func = func
+                .take()
+                .expect("MainContext::invoke() closure called multiple times");
             func();
             glib_sys::G_SOURCE_REMOVE
         }
--- a/src/main_context.rs
+++ b/src/main_context.rs
@@ -86,6 +86,15 @@ impl MainContext {
     where
         F: FnOnce() + 'static,
     {
+        unsafe extern "C" fn trampoline<F: FnOnce() + 'static>(func: gpointer) -> gboolean {
+            let func: &mut Option<F> = &mut *(func as *mut Option<F>);
+            let func = func.take().expect("MainContext::invoke() closure called multiple times");
+            func();
+            glib_sys::G_SOURCE_REMOVE
+        }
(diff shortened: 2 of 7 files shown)


EPashkin:
Gir submodule updated, but version.txt is not

GuillaumeGomez:
Oups, I'll remove the commit.

EPashkin:
Thanks, 👍 
But I still not sure that moving functions in is good for main_context_channel.rs, src/subclass/object.rs` and 
 src/subclass/types.rs, because it is too long functions, and now it is difficult to find where the actual function begins

GuillaumeGomez:
I'm not sure it's better either. Before I update the code, let's see what @sdroege thinks about it.

src/byte_array.rs
```diff
@@ -105,6 +105,22 @@ impl ByteArray {
     }
 
     pub fn sort<F: FnMut(&u8, &u8) -> Ordering>(&self, compare_func: F) {
+        unsafe extern "C" fn compare_func_trampoline(
```
sdroege: This seems good

src/main_context.rs
```diff
@@ -69,7 +69,17 @@ impl MainContext {
     }
 
     unsafe fn invoke_unsafe<F>(&self, priority: Priority, func: F)
-    where F: FnOnce() + 'static {
+        where F: FnOnce() + 'static
+    {
+        unsafe extern "C" fn trampoline<F: FnOnce() + 'static>(func: gpointer) -> gboolean {
```
sdroege: And this

src/main_context_channel.rs
```diff
@@ -224,96 +224,6 @@ struct ChannelSource<T, F: FnMut(T) -> Continue + 'static> {
     callback: Option<RefCell<F>>,
 }
 
-unsafe extern "C" fn prepare<T>(
```
sdroege: These should stay outside. They're not trampolines like the others ones, but the actual implementation of these functions. It's not just something translating between C and Rust.

src/signal.rs
```diff
@@ -48,8 +48,16 @@ impl ToGlib for Inhibit {
...
 
-pub unsafe fn connect_raw<F>(receiver: *mut gobject_sys::GObject, signal_name: *const c_char, trampoline: GCallback,
-                      closure: *mut F) -> SignalHandlerId {
+pub unsafe fn connect_raw<F>(
+    receiver: *mut gobject_sys::GObject,
+    signal_name: *const c_char,
+    trampoline: GCallback,
+    closure: *mut F,
+) -> SignalHandlerId {
+    unsafe extern "C" fn destroy_closure<F>(ptr: *mut c_void, _: *mut gobject_sys::GClosure) {
```
sdroege: This is good

src/subclass/boxed.rs
```diff
@@ -41,6 +41,16 @@ pub trait BoxedType: Clone + Sized + 'static {
 ///
 /// [`glib_boxed_type!`]: ../../macro.glib_boxed_type.html
 pub fn register_boxed_type<T: BoxedType>() -> ::Type {
+    unsafe extern "C" fn boxed_copy<T: BoxedType>(v: glib_sys::gpointer) -> glib_sys::gpointer {
```
sdroege: This also good

src/subclass/boxed.rs
```diff
@@ -41,6 +41,16 @@ pub trait BoxedType: Clone + Sized + 'static {
...
+        let v = &*(v as *mut T);
+        let copy = Box::new(v.clone());
+
+        Box::into_raw(copy) as glib_sys::gpointer
+    }
+    unsafe extern "C" fn boxed_free<T: BoxedType>(v: glib_sys::gpointer) {
```
sdroege: And this of course good too

src/subclass/interface.rs
```diff
@@ -249,14 +249,6 @@ pub trait ObjectInterfaceExt: ObjectInterface {
 
 impl<T: ObjectInterface> ObjectInterfaceExt for T {}
 
-unsafe extern "C" fn interface_init<T: ObjectInterface>(
```
sdroege: This not. While it is close to a trampoline in principle, it's even closer to an actual function implementation. I'd keep this outside.

src/subclass/object.rs
```diff
@@ -65,58 +65,6 @@ pub trait ObjectImpl: ObjectImplExt + 'static {
     }
 }
 
-unsafe extern "C" fn get_property<T: ObjectSubclass>(
```
sdroege: Same for the virtual method ones below, those are nicer outside as they contain quite some code and are not simple trampolines.

sdroege: Especially, for other classes where you have dozens of virtual methods you'd otherwise have a huge block of functions inline inside the `class_init` function. Here we only have three and it's quite hard to read already if not handled as separate functions like it was before.

src/subclass/types.rs
```diff
@@ -363,95 +363,95 @@ pub trait ObjectSubclass: ObjectImpl + Sized + 'static {
     }
 }
 
-unsafe extern "C" fn class_init<T: ObjectSubclass>(klass: glib_sys::gpointer, _klass_data: glib_sys::gpointer)
```
sdroege: These should also all stay as separate functions. Too much code inline otherwise, they're basically like virtual methods above.

src/translate.rs
```diff
@@ -1745,6 +1737,16 @@ impl FromGlibContainer<*const c_char, *mut glib_sys::GHashTable> for HashMap<Str
 #[allow(clippy::implicit_hasher)]
 impl FromGlibPtrContainer<*const c_char, *mut glib_sys::GHashTable> for HashMap<String, String> {
     unsafe fn from_glib_none(ptr: *mut glib_sys::GHashTable) -> Self {
+        unsafe extern "C" fn read_string_hash_table(
```
sdroege: Good :)

src/value_array.rs
```diff
@@ -62,6 +62,22 @@ impl ValueArray {
     }
 
     pub fn sort_with_data<F: FnMut(&Value, &Value) -> Ordering>(&mut self, compare_func: F) {
+        unsafe extern "C" fn compare_func_trampoline(
```
sdroege: Good!

sdroege:
> I'm not sure it's better either. Before I update the code, let's see what @sdroege thinks about it.

I added a comment to each function / group of functions :)

GuillaumeGomez:
Updated!

PR #495, opened 2019-06-10
Manually implement Clone for Sender/SyncSender

sdroege:
The automatically derived implementation will only work if the channel's
item type is also implementing Clone, which is not required here.

Fixes https://github.com/gtk-rs/glib/issues/494

### DIFF
--- a/src/main_context_channel.rs
+++ b/src/main_context_channel.rs
@@ -321,8 +321,14 @@ unsafe extern "C" fn finalize<T, F: FnMut(T) -> Continue + 'static>(
 /// See [`MainContext::channel()`] for how to create such a `Sender`.
 ///
 /// [`MainContext::channel()`]: struct.MainContext.html#method.channel
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct Sender<T>(Option<Channel<T>>);
+
+impl<T> Clone for Sender<T> {
+    fn clone(&self) -> Sender<T> {
+        Sender(self.0.clone())
+    }
+}
 
 impl<T> Sender<T> {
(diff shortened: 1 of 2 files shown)


src/main_context_channel.rs
```diff
@@ -363,9 +369,15 @@ impl<T> Drop for Sender<T> {
 /// See [`MainContext::sync_channel()`] for how to create such a `SyncSender`.
 ///
 /// [`MainContext::sync_channel()`]: struct.MainContext.html#method.sync_channel
-#[derive(Clone, Debug)]
+#[derive(Debug)]
```
GuillaumeGomez: Shouldn't we do the same for `Debug`?

sdroege: No, we can only have a `Debug` impl if `T: Debug`. It is printing the queued items.

GuillaumeGomez: That's what I said: isn't it forcing `T` to implement `Debug`?

sdroege: Yes, so that it can print the currently queued up items.

sdroege: For context, the `std::mpsc` `Sender` / `Receiver` have a manual `Debug` impl that work with any type.

I guess for consistency we should do the same. Means that `Debug` is available always, and that we print fewer details.

What do you think?

sdroege: I'd prefer consistency btw

GuillaumeGomez: I prefer consistency too.

sdroege: Done

PR #496, opened 2019-06-10
Add init/clear function for boxed types

sdroege:
These allow allocated them on our side and releasing all memory of them
    again. It is needed for functions that take an out parameter and require
    us to allocate the underlying object.
    
    Fixes https://github.com/gtk-rs/glib/issues/469

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -17,69 +17,38 @@ use translate::*;
 #[macro_export]
 macro_rules! glib_boxed_wrapper {
     ([$($attr:meta)*] $name:ident, $ffi_name:path, @copy $copy_arg:ident $copy_expr:expr,
-     @free $free_arg:ident $free_expr:expr,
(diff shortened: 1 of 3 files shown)


sdroege:
The build failure requires changes in the `gtk` crate btw, to add these functions to those types there.

GuillaumeGomez:
Looks good to me! :+1:

sdroege:
> Looks good to me! +1

https://github.com/gtk-rs/glib/issues/469#issuecomment-500317950 seems fine to you too?

GuillaumeGomez:
I didn't see that comment. Seems quite complicated...

sdroege:
> I didn't see that comment. Seems quite complicated...

Comment that there please :P And it's complicated indeed, but I don't see an easier solution. Shouldn't matter much though as this will only be needed by very few types, and people will notice if it's missing.

EPashkin:
@sdroege Thanks again, 👍

PR #497, opened 2019-06-10
Format code with rustfmt

sdroege:
@GuillaumeGomez you mentioned a rustfmt config you prefer over the default. Can you provide that one so it can be integrated here?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ script:
     cargo clippy --release;
     fi
   - if [ "$TRAVIS_RUST_VERSION" == "nightly" ] && [ "$GTK" == "3.14" ] && ! [ "$ARM" == "1" ]; then
+    rustup component add rustfmt;
     make regen_check;
     fi
   - if ! [ "$ARM" == "1" ]; then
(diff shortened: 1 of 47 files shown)


GuillaumeGomez:
That's something @alatiera might have. :)

alatiera:
> That's something @alatiera might have. :)

What I have in my local checkout is the following:

```sh
➜  gir ✗ cat rustfmt.toml 
indent_style = "Visual"
```

sdroege:
@alatiera said it was

```toml
indent_style = "Visual"
```

Unfortunately configuration in rustfmt is still unstable

```
Warning: can't set `indent_style = Visual`, unstable features are only available in nightly channel.
```

So either we go with the default for now, or more waiting.

sdroege:
@GuillaumeGomez What do you think? Does the formatted code in this PR look fine to you until we can enable that rustfmt configuration, or do you prefer to wait (in which case we should create an issue to ask for stabilization of the configuration we need: you only require `ident_style`?)?

GuillaumeGomez:
I'm not a big fan but that's just minor stuff so let's go.

sdroege:
Ok, thanks! We can add the configuration at a later time, and in the meantime you could ask for it to be stabilized if it's important enough.

If @EPashkin is also fine with this, let's get this merged then. And then I'll prepare MRs for everything else doing the same.

GuillaumeGomez:
I opened an issue for the feature stabilization: https://github.com/rust-lang/rustfmt/issues/3622

EPashkin:
👍

PR #504, opened 2019-07-09
Regen

GuillaumeGomez:
Part of https://github.com/gtk-rs/gir/pull/806.

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez can you change manual files too?
Also .travis.yml need version update

GuillaumeGomez:
Oh indeed! Sorry, completely forgot about manual code. For travis, let's do it in another PR if you don't mind. Want to do it?

EPashkin:
I prefer to fix travis build in this PR, but I do not insist

GuillaumeGomez:
Let's do it in here then!

GuillaumeGomez:
@EPashkin Do you want me to update the version to `v2_58` directly?

EPashkin:
@GuillaumeGomez I meant rust version (to 1.36.0) and don't think that we need do something with glib version.

src/value.rs
```diff
@@ -201,7 +201,7 @@ impl Value {
     #[doc(hidden)]
     pub fn into_raw(mut self) -> gobject_sys::GValue {
         unsafe {
-            let ret = mem::replace(&mut self.0, mem::uninitialized());
+            let ret = mem::replace(&mut self.0, mem::MaybeUninit::uninit().assume_init());
```
EPashkin: IMHO `MaybeUninit::uninit()` as we forget it anyway

EPashkin: > IMHO MaybeUninit::uninit() as we forget it anyway
Ignore this, I was wrong

EPashkin:
Thanks, and 👍

sdroege:
If we want to still support pre-1.36, there's https://crates.io/crates/maybe-uninit. I wouldn't worry about it though.

GuillaumeGomez:
I prefer not start supporting older version.

.travis.yml
```diff
@@ -15,14 +15,14 @@ matrix:
...
+      rust: 1.36.0
       env: GTK=3.14 FEATURES=subclassing
     - os: linux
-      rust: 1.34.0
+      rust: 1.36.0
       env: GTK=3.24 FEATURES=subclassing,v2_48
     - os: osx
       rust: nightly
-      env: GTK=3.14 FEATURES=subclassing,futures
+      env: GTK=3.16 FEATURES=subclassing,futures
```
sdroege: Why? 3.14 is still the minimum version we support or not?

src/object.rs
```diff
@@ -1436,7 +1436,7 @@ impl<T: ObjectType> ObjectExt for T {
 
     fn downgrade(&self) -> WeakRef<T> {
         unsafe {
-            let w = WeakRef(Box::new(mem::uninitialized()), PhantomData);
+            let w = WeakRef(Box::new(mem::MaybeUninit::uninit().assume_init()), PhantomData);
```
sdroege: This is wrong. `assume_init()` can only be called after the `g_weak_ref_init()` really... but `assume_init()` involves moving data around and that would break the weakref.

I think this can simply be `mem::zeroed()`

src/object.rs
```diff
@@ -1530,7 +1530,7 @@ pub struct WeakRef<T: ObjectType>(Box<gobject_sys::GWeakRef>, PhantomData<*const
 impl<T: ObjectType> WeakRef<T> {
     pub fn new() -> WeakRef<T> {
         unsafe {
-            let w = WeakRef(Box::new(mem::uninitialized()), PhantomData);
+            let w = WeakRef(Box::new(mem::MaybeUninit::uninit().assume_init()), PhantomData);
```
sdroege: Same here

.travis.yml
GuillaumeGomez: Doesn't `MaybeUninitialized` gets stabilized in 1.36?

src/object.rs
```diff
@@ -1560,7 +1560,7 @@ impl<T: ObjectType> Drop for WeakRef<T> {
 impl<T: ObjectType> Clone for WeakRef<T> {
     fn clone(&self) -> Self {
         unsafe {
-            let c = WeakRef(Box::new(mem::uninitialized()), PhantomData);
+            let c = WeakRef(Box::new(mem::MaybeUninit::uninit().assume_init()), PhantomData);
```
sdroege: And here

src/time_val.rs
```diff
@@ -27,6 +27,6 @@ impl<'a> ToGlibPtrMut<'a, *mut glib_sys::GTimeVal> for TimeVal {
 
 impl Uninitialized for TimeVal {
     unsafe fn uninitialized() -> TimeVal {
-        mem::uninitialized()
+        mem::MaybeUninit::uninit().assume_init()
```
sdroege: This is wrong. You can't call `assume_init()` without actually ensuring before that it's initialized. `mem::zeroed()` is fine for this one here too.

src/value.rs
sdroege: This is also wrong for the same reason :) This one can simply be

```rust
let ret = ptr::read(&self.0);
mem::forget(self);
ret
```

.travis.yml
sdroege: Yes but here you change the version for GTK, not Rust :)

GuillaumeGomez: Oh damn, good catch!

GuillaumeGomez:
Updated.

sdroege:
:+1:

PR #505, opened 2019-07-11
Zeroed

GuillaumeGomez:
I'll resume the regenerations of all other crates tomorrow.

cc @EPashkin @sdroege

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -366,7 +366,7 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> Uninitialized for Boxed<T, MM> {
     unsafe fn uninitialized() -> Self {
         Boxed {
             inner: {
-                let mut inner = Box::<T>::new(mem::zeroed());
+                let mut inner = Box::<T>::new(mem::MaybeUninit::zeroed().assume_init());
                 MM::init(&mut *inner);
 
                 AnyBox::Native(inner)
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -366,10 +366,10 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> Uninitialized for Boxed<T, MM> {
     unsafe fn uninitialized() -> Self {
         Boxed {
             inner: {
-                let mut inner = Box::<T>::new(mem::MaybeUninit::zeroed().assume_init());
-                MM::init(&mut *inner);
+                let mut inner = mem::MaybeUninit::zeroed();
+                MM::init(inner.as_mut_ptr());
 
-                AnyBox::Native(inner)
+                AnyBox::Native(Box::new(inner.assume_init()))
             },
             _dummy: PhantomData,
         }
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -93,7 +93,7 @@ impl Closure {
         let mut result = unsafe { Value::uninitialized() };
 
         let v_args: Vec<Value>;
-        let mut s_args: [Value; 10] = unsafe { mem::zeroed() };
+        let mut s_args: [Value; 10] = unsafe { mem::MaybeUninit::zeroed().assume_init() };
         let values = if values.len() <= 10 {
             for (i, arg) in values.iter().enumerate() {
                 s_args[i] = arg.to_value();
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -93,7 +93,7 @@ impl Closure {
         let mut result = unsafe { Value::uninitialized() };
 
         let v_args: Vec<Value>;
-        let mut s_args: [Value; 10] = unsafe { mem::MaybeUninit::zeroed().assume_init() };
+        let mut s_args: [Value; 10] = unsafe { mem::zeroed() };
         let values = if values.len() <= 10 {
             for (i, arg) in values.iter().enumerate() {
                 s_args[i] = arg.to_value();
--- a/src/object.rs
+++ b/src/object.rs
@@ -1269,8 +1269,9 @@ impl<T: ObjectType> ObjectExt for T {
             return Err(glib_bool_error!("Signal not found"));
         }
 
-        let mut details = mem::zeroed();
-        gobject_sys::g_signal_query(signal_id, &mut details);
+        let mut details = mem::MaybeUninit::zeroed();
+        gobject_sys::g_signal_query(signal_id, details.as_mut_ptr());
+        let details = details.assume_init();
         if details.signal_id != signal_id {
             return Err(glib_bool_error!("Signal not found"));
         }
@@ -1370,8 +1371,9 @@ impl<T: ObjectType> ObjectExt for T {
                 return Err(glib_bool_error!("Signal not found"));
             }
 
-            let mut details = mem::zeroed();
-            gobject_sys::g_signal_query(signal_id, &mut details);
+            let mut details = mem::MaybeUninit::zeroed();
+            gobject_sys::g_signal_query(signal_id, details.as_mut_ptr());
+            let details = details.assume_init();
             if details.signal_id != signal_id {
                 return Err(glib_bool_error!("Signal not found"));
             }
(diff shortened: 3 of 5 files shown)


EPashkin:
Thanks, also note that this is seems done internally too https://github.com/rust-lang/rust/pull/62150

src/boxed.rs
```diff
@@ -366,7 +366,7 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> Uninitialized for Boxed<T, MM> {
     unsafe fn uninitialized() -> Self {
         Boxed {
             inner: {
-                let mut inner = Box::<T>::new(mem::zeroed());
+                let mut inner = Box::<T>::new(mem::MaybeUninit::zeroed().assume_init());
```
sdroege: This is wrong again and defeats the purpose of `MaybeUninit`. This code here has to be

```rust
let mut inner = Box::<T>::new(mem::MaybeUninit::zeroed();
MM::init(inner.as_mut_ptr());

AnyBox::Native(inner.assume_init());
```

src/closure.rs
```diff
@@ -93,7 +93,7 @@ impl Closure {
         let mut result = unsafe { Value::uninitialized() };
 
         let v_args: Vec<Value>;
-        let mut s_args: [Value; 10] = unsafe { mem::zeroed() };
+        let mut s_args: [Value; 10] = unsafe { mem::MaybeUninit::zeroed().assume_init() };
```
sdroege: Same. See the example about arrays in the `MaybeUninit` documentation. Or keep this as `mem::zeroed()` if that's not applicable here.

But `assume_init()` right after creating the `MaybeUninit` is an anti-pattern and wrong in 99% of the cases.

src/object.rs
```diff
@@ -1389,7 +1391,7 @@ impl<T: ObjectType> ObjectExt for T {
             }
 
             let mut v_args: Vec<Value>;
-            let mut s_args: [Value; 10] = mem::zeroed();
+            let mut s_args: [Value; 10] = mem::MaybeUninit::zeroed().assume_init();
```
sdroege: Same

src/object.rs
```diff
@@ -1436,7 +1438,7 @@ impl<T: ObjectType> ObjectExt for T {
 
     fn downgrade(&self) -> WeakRef<T> {
         unsafe {
-            let w = WeakRef(Box::new(mem::zeroed()), PhantomData);
+            let w = WeakRef(Box::new(mem::MaybeUninit::zeroed().assume_init()), PhantomData);
```
sdroege: Same

```rust
let w = Box::new(mem::MaybeUninit::zeroed());
gobject_sys::g_weak_ref_init(
    mut_override((&mut *w).as_mut_ptr()),
    self.as_object_ref().to_glib_none().0,
);

// Or maybe this line can be *w = (&mut *w).assume_init()?
ptr::write(&mut *w, ptr::read(&*w).assume_init());

WeakRef(w)
```

src/object.rs
```diff
@@ -1530,7 +1532,7 @@ pub struct WeakRef<T: ObjectType>(Box<gobject_sys::GWeakRef>, PhantomData<*const
 impl<T: ObjectType> WeakRef<T> {
     pub fn new() -> WeakRef<T> {
         unsafe {
-            let w = WeakRef(Box::new(mem::zeroed()), PhantomData);
+            let w = WeakRef(Box::new(mem::MaybeUninit::zeroed().assume_init()), PhantomData);
```
sdroege: See above

src/object.rs
```diff
@@ -1560,7 +1562,7 @@ impl<T: ObjectType> Drop for WeakRef<T> {
 impl<T: ObjectType> Clone for WeakRef<T> {
     fn clone(&self) -> Self {
         unsafe {
-            let c = WeakRef(Box::new(mem::zeroed()), PhantomData);
+            let c = WeakRef(Box::new(mem::MaybeUninit::zeroed().assume_init()), PhantomData);
```
sdroege: See above

src/time_val.rs
```diff
@@ -27,6 +27,6 @@ impl<'a> ToGlibPtrMut<'a, *mut glib_sys::GTimeVal> for TimeVal {
 
 impl Uninitialized for TimeVal {
     unsafe fn uninitialized() -> TimeVal {
-        mem::zeroed()
+        mem::MaybeUninit::zeroed().assume_init()
```
sdroege: As we want to get rid of this trait anyway, just leave it as `mem::zeroed()`. What you write is exactly equivalent to it apart from being more keypresses.

src/value.rs
```diff
@@ -273,7 +273,7 @@ impl From<SendValue> for Value {
 
 impl Uninitialized for Value {
     unsafe fn uninitialized() -> Value {
-        Value(mem::zeroed(), PhantomData)
+        Value(mem::MaybeUninit::zeroed().assume_init(), PhantomData)
```
sdroege: See above

GuillaumeGomez:
Updated. However, I'm **absolutely** not sure that this code is correct. I find this new rust API very bad...

sdroege:
On 12 July 2019 23:56:13 EEST, Guillaume Gomez <notifications@github.com> wrote:
>Updated. However, I'm **absolutely** not sure that this code is
>correct. I find this new rust API very bad...

I'll review this and the other ones tomorrow evening.

Personally I think the new API is much clearer. You can't accidentally use something uninitialized and it's absolutely clear at which point things are actually initialized. It's much safer compared to before, which was basically like in C and not very Rust-like.


-- 

src/closure.rs
sdroege: Just keep this `mem::zeroed()` if you can't nicely use `MaybeUninit` here :)

src/boxed.rs
```diff
@@ -366,10 +366,10 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> Uninitialized for Boxed<T, MM> {
     unsafe fn uninitialized() -> Self {
         Boxed {
             inner: {
-                let mut inner = Box::<T>::new(mem::zeroed());
-                MM::init(&mut *inner);
+                let mut inner = Box::new(mem::MaybeUninit::zeroed());
+                MM::init(inner.as_mut_ptr());
 
-                AnyBox::Native(inner)
+                AnyBox::Native(Box::new(inner.assume_init()))
```
sdroege: You're creating two boxes here now instead of one. Maybe just

```rust
let mut inner = mem::MaybeUninit::zeroed();
MM::init(inner.as_mut_ptr());
AnyBox::Native(Box::new(inner.assume_init()))
```
?

src/object.rs
```diff
@@ -1436,12 +1438,13 @@ impl<T: ObjectType> ObjectExt for T {
...
-            let w = WeakRef(Box::new(mem::zeroed()), PhantomData);
+            let mut w = Box::new(mem::MaybeUninit::zeroed());
             gobject_sys::g_weak_ref_init(
-                mut_override(&*w.0),
+                mut_override((&mut *w).as_mut_ptr()),
                 self.as_object_ref().to_glib_none().0,
             );
-            w
+
+            WeakRef(Box::new((&mut *w).assume_init()), PhantomData)
```
sdroege: This creates two boxes and also changes the memory location of the weak reference, so it will break at runtime. It circumvents the whole reason why we box here at all.

The way how it was before is the best we can do here.

src/object.rs
```diff
@@ -1530,9 +1533,13 @@ pub struct WeakRef<T: ObjectType>(Box<gobject_sys::GWeakRef>, PhantomData<*const
...
-            let w = WeakRef(Box::new(mem::zeroed()), PhantomData);
-            gobject_sys::g_weak_ref_init(mut_override(&*w.0), ptr::null_mut());
-            w
+            let mut w = Box::new(mem::MaybeUninit::zeroed());
+            gobject_sys::g_weak_ref_init(
+                (&mut *w).as_mut_ptr(),
+                ptr::null_mut(),
+            );
+
+            WeakRef(Box::new((&mut *w).assume_init()), PhantomData)
```
sdroege: Same

src/object.rs
sdroege: `mem:zeroed()` is fine here, fewer keystrokes and same effect.

GuillaumeGomez:
Updated.

PR #508, opened 2019-07-20
Implement FromGlib and ToGlib traits on Pid type

GuillaumeGomez:
cc @sdroege @EPashkin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,12 @@ matrix:
     - os: linux
       rust: 1.36.0
       env: GTK=3.24 FEATURES=subclassing,v2_48
+    - os: linux
+      rust: stable
+      env: GTK=3.14 FEATURES=subclassing
+    - os: linux
+      rust: stable
+      env: GTK=3.24 FEATURES=subclassing,v2_48
     - os: osx
       rust: nightly
       env: GTK=3.14 FEATURES=subclassing,futures
@@ -31,6 +37,12 @@ matrix:
       env: GTK=3.14 FEATURES=subclassing
     # - os: osx
     #   rust: beta
+    #   env: GTK=3.24 FEATURES=subclassing,v2_48
+    - os: osx
+      rust: 1.36.0
+      env: GTK=3.14 FEATURES=subclassing
+    # - os: osx
+    #   rust: 1.36.0
     #   env: GTK=3.24 FEATURES=subclassing,v2_48
     - os: osx
       rust: stable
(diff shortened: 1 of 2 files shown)


EPashkin:
Windows is again against us.
Maybe we need use `glib_sys::GPid` instead `i32` ?
```
error[E0308]: mismatched types
  --> src\source.rs:57:9
   |
56 |     fn to_glib(&self) -> i32 {
   |                          --- expected `i32` because of return type
57 |         self.0
   |         ^^^^^^ expected i32, found *-ptr
   |
   = note: expected type `i32`
              found type `*mut libc::c_void`
error[E0308]: mismatched types
  --> src\source.rs:65:13
   |
65 |         Pid(val)
   |             ^^^ expected *-ptr, found i32
   |
   = note: expected type `*mut libc::c_void`
              found type `i32`
```

GuillaumeGomez:
@EPashkin I think too. I'll update the PR.

sdroege:
Wants a `cargo fmt` apparently, otherwise looks good to me.

Maybe not a good idea to run the `cargo fmt` CI test with nightly. That should be done with stable or otherwise we get spurious issues like this one whenever there's a rustfmt regression in nightly.

.travis.yml
```diff
@@ -59,7 +59,7 @@ script:
     rustup component add clippy;
     cargo clippy --release;
     fi
-  - if [ "$TRAVIS_RUST_VERSION" == "nightly" ] && [ "$GTK" == "3.14" ] && ! [ "$ARM" == "1" ]; then
+  - if ! [ "$TRAVIS_RUST_VERSION" == "nightly" ] && ! [ "$TRAVIS_RUST_VERSION" == "beta" ] && [ "$GTK" == "3.14" ] && ! [ "$ARM" == "1" ]; then
```
sdroege: Hmm why not `== stable`? Hrm, and actually add a stable job to travis in addition to the lower-bound 1.36 job :)

GuillaumeGomez: Sounded simpler that way. I don't want yet another special case.

sdroege: We should have a stable job too though, that is always running with latest stable instead of fixed 1.36

GuillaumeGomez: This is a good argument. Should I run it on both OSX and linux or just linux is fine?

sdroege: all

EPashkin:
👍 @GuillaumeGomez Thanks

PR #512, opened 2019-08-05
Mark ByteArray::set_size() as unsafe

sdroege:
Just like the version of Vec, this does not actually initialize the
memory if the array has to grow and allows to get uninitialized memory.

The unit test actually caused valgrind warnings because of that.

Also extend unit test a bit.

----

CC @elmarco @GuillaumeGomez


EPashkin:
@sdroege Thanks

PR #513, opened 2019-08-07
 Value::get: return a Result to account for type mismatch...

### LINKED ISSUE #510: Impossible to distinguish Value::get::<glib::Object>() returning None because type mismatch or no value contained
We should probably change the signature so that it returns a `Result`, and in case of nullable types (like objects) the success case of the `Result` would be an `Option`. The error case of the `Result` would happen on type mismatches.

Opinions?

fengalin:
... and introduce `get_some` for types implementing `FromValue`.

First commit is for non-regression validation.

I'll update the other crates (gtk-rs & gstreamer-rs) when this is merged.

Fixes #510

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -135,10 +135,10 @@ mod tests {
 
     fn closure_fn(values: &[Value]) -> Option<Value> {
         assert_eq!(values.len(), 2);
-        let string_arg: Option<String> = values[0].get().unwrap();
-        assert_eq!(string_arg, Some("test".to_string()));
-        let int_arg: i32 = values[1].get_some().unwrap();
-        assert_eq!(int_arg, 42);
+        let string_arg = values[0].get::<String>();
+        assert_eq!(string_arg, Ok(Some("test".to_string())));
+        let int_arg = values[1].get_some::<i32>();
+        assert_eq!(int_arg, Ok(42));
         Some(24.to_value())
     }
 
@@ -150,10 +150,10 @@ mod tests {
         let closure = Closure::new(move |values| {
             count.fetch_add(1, Ordering::Relaxed);
             assert_eq!(values.len(), 2);
-            let string_arg: Option<String> = values[0].get().unwrap();
-            assert_eq!(string_arg, Some("test".to_string()));
-            let int_arg: i32 = values[1].get_some().unwrap();
-            assert_eq!(int_arg, 42);
+            let string_arg = values[0].get::<String>();
+            assert_eq!(string_arg, Ok(Some("test".to_string())));
+            let int_arg = values[1].get_some::<i32>();
+            assert_eq!(int_arg, Ok(42));
             None
         });
         let result = closure.invoke(&[&"test".to_string(), &42]);
@@ -166,7 +166,7 @@ mod tests {
 
         let closure = Closure::new(closure_fn);
         let result = closure.invoke(&[&"test".to_string(), &42]);
-        let int: Option<i32> = result.and_then(|result| result.get().unwrap());
-        assert_eq!(int, Some(24));
+        let int_res = result.map(|result| result.get_some::<i32>());
+        assert_eq!(int_res, Some(Ok(24)));
     }
 }
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -135,10 +135,10 @@ mod tests {
 
     fn closure_fn(values: &[Value]) -> Option<Value> {
         assert_eq!(values.len(), 2);
-        let string: Option<String> = values[0].get();
-        assert_eq!(string, Some("test".to_string()));
-        let int: Option<i32> = values[1].get();
-        assert_eq!(int, Some(42));
+        let string_arg: String = values[0].get_some().unwrap();
+        assert_eq!(string_arg, "test".to_string());
+        let int_arg: i32 = values[1].get_some().unwrap();
+        assert_eq!(int_arg, 42);
         Some(24.to_value())
     }
 
@@ -150,10 +150,10 @@ mod tests {
         let closure = Closure::new(move |values| {
             count.fetch_add(1, Ordering::Relaxed);
             assert_eq!(values.len(), 2);
-            let string: Option<String> = values[0].get();
-            assert_eq!(string, Some("test".to_string()));
-            let int: Option<i32> = values[1].get();
-            assert_eq!(int, Some(42));
+            let string_arg: String = values[0].get_some().unwrap();
+            assert_eq!(string_arg, "test".to_string());
+            let int_arg: i32 = values[1].get_some().unwrap();
+            assert_eq!(int_arg, 42);
             None
         });
         let result = closure.invoke(&[&"test".to_string(), &42]);
@@ -166,7 +166,7 @@ mod tests {
 
         let closure = Closure::new(closure_fn);
         let result = closure.invoke(&[&"test".to_string(), &42]);
-        let int: Option<i32> = result.and_then(|result| result.get());
+        let int: Option<i32> = result.and_then(|result| result.get().unwrap());
         assert_eq!(int, Some(24));
     }
 }
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -135,8 +135,8 @@ mod tests {
 
     fn closure_fn(values: &[Value]) -> Option<Value> {
         assert_eq!(values.len(), 2);
-        let string_arg: String = values[0].get_some().unwrap();
-        assert_eq!(string_arg, "test".to_string());
+        let string_arg: Option<String> = values[0].get().unwrap();
+        assert_eq!(string_arg, Some("test".to_string()));
         let int_arg: i32 = values[1].get_some().unwrap();
         assert_eq!(int_arg, 42);
         Some(24.to_value())
@@ -150,8 +150,8 @@ mod tests {
         let closure = Closure::new(move |values| {
             count.fetch_add(1, Ordering::Relaxed);
             assert_eq!(values.len(), 2);
-            let string_arg: String = values[0].get_some().unwrap();
-            assert_eq!(string_arg, "test".to_string());
+            let string_arg: Option<String> = values[0].get().unwrap();
+            assert_eq!(string_arg, Some("test".to_string()));
             let int_arg: i32 = values[1].get_some().unwrap();
             assert_eq!(int_arg, 42);
             None
--- a/src/value.rs
+++ b/src/value.rs
@@ -23,7 +23,7 @@
 //!
 //! ```
 //! use glib::prelude::*; // or `use gtk::prelude::*;`
-//! use glib::{TypedValue, Value};
+//! use glib::{TypedValue, Value, ValueTypeMismatch};
 //!
 //! // Value and TypedValue implement From<&i32>, From<&str>
 //! // and From<Option<&str>>. Another option is the `ToValue` trait.
@@ -40,7 +40,7 @@
 //! // `get` tries to get an optional value of the specified type
 //! // and returns an `Err` if the type doesn't match.
 //! assert_eq!(num.get().unwrap(), Some(10));
-//! assert!(num.get::<String>().is_err());
+//! assert_eq!(num.get::<String>(), Err(ValueTypeMismatch));
 //! assert_eq!(hello.get().unwrap(), Some(String::from("Hello!")));
 //! assert_eq!(hello.get::<String>().unwrap(), Some(String::from("Hello!")));
 //! assert_eq!(str_none.get::<String>().unwrap(), None);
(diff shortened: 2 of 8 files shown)


src/value.rs
```diff
@@ -156,26 +163,36 @@ impl Value {
...
-    /// Tries to get a value of type `T`.
-    ///
-    /// Returns `Some` if the type is correct and the value is not `None`.
+    /// Tries to get a possibly optional value of type `T`.
     ///
-    /// This function doesn't distinguish between type mismatches and correctly
-    /// typed `None` values. Use `downcast` or `is` for that.
-    pub fn get<'a, T: FromValueOptional<'a>>(&'a self) -> Option<T> {
+    /// Returns `Ok` if the type is correct.
+    pub fn get<'a, T: FromValueOptional<'a>>(&'a self) -> Result<Option<T>, BoolError> {
```
sdroege: This should probably be a new error type, some kind of
```rust
struct ValueTypeMismatchError;
```

src/value.rs
```diff
@@ -156,26 +163,36 @@ impl Value {
...
+                )))
             }
         }
     }
 
+    /// Tries to get a value of type `T`.
+    ///
+    /// Returns `Ok` if the type is correct and `Some` value is available.
+    /// (check the `BoolError` description for details in case of an error).
+    pub fn get_some<'a, T: FromValueOptional<'a>>(&'a self) -> Result<T, BoolError> {
```
sdroege: This is a weird one :) I think you want the trait bound to be simply `FromValue`, just like `TypeValue::get_some()` works. For types implementing that trait there can't be a `None`

src/value.rs
fengalin: Yes I agree, but on the other hand, `BoolError` is used elsewhere, like in `{g,s}et_property`.

src/value.rs
fengalin: I thought it would also be useful when you expect a value to be defined, even though the type is "option-able".

sdroege: That seems wrong. In that case you'd explicitly unwrap/expect because *you* know that the value exists :) Deferring that to the function here hides this assumption in the code

src/value.rs
sdroege: No strong opinion, what you think is best here :) Or maybe @GuillaumeGomez has an opinion

I'm not a fan of `BoolError` for things where we explicitly handle different error cases (other than just a C function returning `false`), and I'd like to change things to explicit error types. Just didn't get to it, or even to create an issue for discussion :)

src/closure.rs
```diff
@@ -135,10 +135,10 @@ mod tests {
 
     fn closure_fn(values: &[Value]) -> Option<Value> {
         assert_eq!(values.len(), 2);
-        let string: Option<String> = values[0].get();
-        assert_eq!(string, Some("test".to_string()));
-        let int: Option<i32> = values[1].get();
-        assert_eq!(int, Some(42));
+        let string_arg: String = values[0].get_some().unwrap();
```
sdroege: With `get_some()` changed, this would also then still have `Option<String>` as type here, or would need a second explicit `unwrap()` for the "None-ability" of the string.

I think it also makes various other code changes in this PR clearer

src/value.rs
fengalin: It doesn't seem hidden to me since it is explicit in the function name :)

If we were expected to use it solely for `FromValue` types, we should name it something like `get_non_optional` (I agree this is ugly) :)

sdroege: I'd keep `get_some()` consistent with what we have right now in `TypedValue`. So either keep the name and change this to `FromValue` here, or rename both and have them both work with `FromValue` :)

fengalin: I can do that!

@GuillaumeGomez & @EPashkin, what do you think?

src/value.rs
GuillaumeGomez: I agree with @sdroege on this one. I prefer to have things as explicit as possible.

src/value.rs
GuillaumeGomez: I'm not sure we want to have something "option-able" in such a generic context. `FromValue` seems good enough for me.

sdroege: But keep it called `get_some()` (which is weird), or name it differently (how?)?

GuillaumeGomez: I don't have opinion on this point. Good luck! :)

sdroege: `get_non_optional()` seems most descriptive to me, just a bit long to type. I don't have any better suggestions :)

fengalin: `get_mandatory`? But that not accurate IMO as the types are not optional anyway.

Another option would be to rename current `get` as `get_optional` and to implement `FromValueOptional` only on types that can host `None`. The trait would be renamed as `FromOptionalValue` and we could rename current `get_some` methods as `get`.

sdroege: > Another option would be to rename current `get` as `get_optional`

That seems fine

> The trait would be renamed as `FromOptionalValue` and we could rename current `get_some` methods as `get`.

That too but why rename?

> Implement `FromValueOptional` only on types that can host `None`

And this would break things :) The optional types are a full superset of the non-optional ones. You can handle each non-optional type (e.g. the integers) as a optional one, you'd just get a `Some` around it, always.

Why would you change this?

----

My main worry with changing `get()` to only work on non-optional types is all the user complaints. Most of the time you work with optional types and `get()` is much more visible than some other `get_XXX()`, and people will probably get confused. And they'll need to update more of their code than otherwise.

fengalin: I was proposing to rename `FromValueOptional` to reflect that it would only apply to optional values (as opposed to yielding an `Option` from a `Value`). I noticed this trait was a superset of `FromValue` type, but it seemed confusing IMHO. Why would we want to get an `Option<T>` when we explicitly indicate a type that is not an option-able? Then I though that `FromValueOptional` was implemented for these types for convenience when used in `get` methods. Using `get` / `get_optional` would make it no longer necessary to be a superset. But I must admit I didn't engage in a large scale survey, so I might be missing other use cases.

I understand your concerns about users being bothered by these changes. If we decide to proceed with these changes, I could add a comment in the documentation of the new `get` methods and link to the `get_optional` versions to clarify. I think it is convenient to being able to `get` embedded numerics, booleans and types without the extra `unwrap()`.

sdroege: Having the `get()` work for all types allows you to write generic code over all types, instead of having to special case both traits everywhere.

fengalin: Ok, so I'll keep the `get` methods as-is and will try to figure out a better name for `get_some`, possibly `get_non_optional`.

sdroege: Ok! Not great either, I agree. But all alternatives I can see currently seem worse :)

src/value.rs
```diff
@@ -89,6 +93,9 @@ use gstring::GString;
 use translate::*;
 use types::{StaticType, Type};
 
+#[derive(Debug, PartialEq)]
+pub struct ValueTypeMismatch;
```
sdroege: Should impl `std::error::Error`

src/value.rs
sdroege: Looks good to me now otherwise, only needs renaming I guess

src/value.rs
sdroege: Maybe should also carry the two `glib::Type`s, the expected and the one we found? That would allow for a more useful `Display` impl at least.

fengalin: Yes! I was doing exactlty so :) Thanks for the heads-up

fengalin:
Not ready yet! I still hope to find a good name for `get_some` methods :)

src/value.rs
```diff
@@ -94,7 +96,26 @@ use translate::*;
 use types::{StaticType, Type};
 
 #[derive(Debug, PartialEq)]
-pub struct ValueTypeMismatch;
+pub struct ValueTypeMismatch {
```
sdroege: `ValueTypeMismatchError`. At least in the standard library all the errors end on `Error` IIRC... or not?

fengalin: I can change that, but I though it was already explicitly referring to an error, especially when used in expressions such as `Err(ValueTypeMismatch)`. `Err(ValueTypeMismatchError)` seems to state 3x there is an error :)

A [brief search in the `std` doc](https://doc.rust-lang.org/std/index.html?search=Error) shows that error types are more generic ex:

- [RecvError](https://doc.rust-lang.org/std/sync/mpsc/struct.RecvError.html)
- [Utf8Error](https://doc.rust-lang.org/std/str/struct.Utf8Error.html)
- [ParseError](https://doc.rust-lang.org/std/string/type.ParseError.html)

My first implementation was:

``` rust
pub enum ValueError {
    TypeMismatch,
}
```

but then I though it was a little too much since there was no sign of any other error types in `Value` yet, but that could be a good compromise. What do you think (of course with the `actual` and `requested` fields)? And of course, I'll change the `struct` to `ValueTypeMismatchError` if you think it is more Rust-compliant.

sdroege: Yeah I agree. I also would've done an enum first, but there's literally no other possible error with anything related to `GValue` :)

For other things there will be enum errors (like for getting properties, emitting signals, etc) as there are multiple error cases.

`TryFromIntError` would be an example from std that is a plain struct but has the `Error` suffix. That could've also be called `OutOfRange` I guess... so maybe in analogy to that, our error type should actually be called `ValueGetError` or something like that? :)

fengalin: Yes you're right, `Error`s seem to be named after the method that generated them, see in [Receiver](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html) which shows several examples. But they don't use the module's name. I guess this is because they are declared in the module and not re-exported in the crate root. So we could name this one `GetError` and I will no longer export it in `lib.rs`, we'll have to use `value::GetError` from other modules.

While I'm at it, I'll implement `Clone`, `Copy`, `Eq`, `Send` & `Sync` on it.

sdroege: > Edit: `Type` is neither `Send` nor `Sync`. Any reason for this?

It is both, and your error type is also both already. These are automatically implemented in this case.

sdroege: > So we could name this one `GetError` and I will no longer export it in `lib.rs`, we'll have to use `value::GetError` from other modules.

Works for me, or `ValueGetError` and re-exporting is also fine. No preference.

fengalin: > It is both, and your error type is also both already. These are automatically implemented in this case.

Indeed! Thanks

fengalin:
I think we should keep `get_some` eventually. I can't find a better short replacement.

I am worried about the [CI error on Windows 32bits](https://ci.appveyor.com/project/GuillaumeGomez/glib/builds/26555598/job/f90x0sqsfokbj90i#L425): it seems like this platform doesn't like [`#[should_panic]` attributes in tests](https://github.com/fengalin/glib/blob/1dacba6412cf71bf712844f72c91f945a23d5cbe/src/subclass/object.rs#L698). Any of you guys seen this before?

sdroege:
Probably related to using the mingw toolchain. Not sure what to do about that, the Windows CI is generally borderline broken (see also GTK failing).

sdroege:
Maybe blacklist that test on Windows for the time being.

EPashkin:
There also error on installing mingw-w64-i686-glib2, but it was not passed build too:
https://ci.appveyor.com/project/GuillaumeGomez/glib/builds/26555598/job/f90x0sqsfokbj90i#L103

fengalin:
> Maybe blacklist that test on Windows for the time being

I'm not sure if there is a better way to do this: I added a `#[cfg(..)]` clause before the 2 tests that `#[should_panic]`.

> There also error on installing mingw-w64-i686-glib2

Yup! It seems kind of broken ATM.

fengalin:
CI on Windows 32bits is now aligned with the other platforms.

I think the next step would be to merge https://github.com/gtk-rs/gir/pull/825 so that I can regen the depending crates.

sdroege:
This PR here looks good to me in any case, thanks!

If you're interested you could also update all the `BoolError` in our manual code (like in the `set_property()` function) to proper error types :)

fengalin:
> If you're interested you could also update all the BoolError in our manual code (like in the set_property() function) to proper error types :)

Ok, I'll add it to my todo list :) But I had other `glib`-related plans actually. I'll post a proposal when I'm done with the `Value::get` changes.

sdroege:
> Ok, I'll add it to my todo list :) But I had other `glib`-related plans actually. I'll post a proposal when I'm done with the `Value::get` changes.

No problem :)

EPashkin:
As result code looks like this:
```
let s = model
                    .get_value(&iter, 1)
                    .get::<String>()
                    .expect("Value is not string"),
                    .expect("Couldn't get string value"),
```
Maybe we try include `None` to error?
```
pub enum GetError {
    Nothing,
    WrongType{requested: Type, actual: Type},
}
```

fengalin:
> Maybe we try include `None` to error?

The `None` case is not necessarily an error. There are cases for which you will want an optional `String`. See for instance [this test case](https://github.com/fengalin/glib/blob/64c93c54f4e0e43bdb256ad52a0f7195ba5e76a8/src/subclass/object.rs#L574) in `subclass/object.rs`.

For types that don't allow storing a `None` variant, we can use [`get_some`](https://github.com/fengalin/glib/blob/64c93c54f4e0e43bdb256ad52a0f7195ba5e76a8/src/value.rs#L215).

EPashkin:
Yes, with my proposal most of code will be use `.ok()` to get `Option<T>` so type error will be not handled correctly.
👍 for current version.
@fengalin Thanks

fengalin:
https://github.com/gtk-rs/examples/pull/257 all green!

I think we can merge this one and I'll proceed with the direct dependents (`gio`, `atk`, `gdk-pixbuf`), then the others...

sdroege:
All good for me, but needs @GuillaumeGomez to merge all the trivial regen PRs :)

Issue #515, opened 2019-08-13
Make glib::value::GetError fields publics


fengalin:
... so that we can read them outside of the module and we can build an error for use in `assert_eq!` expressions.

sdroege:
Or a constructor? Either way seems fine to me

fengalin:
Yeah, I'm not sure yet. We don't want users to change the fields values, but OTOH they might want to read the fields.

fengalin:
Of course adding a constructor is useful, there's no question about this. One argument in favour of setting the fields as public is that when we declare an `Error` based on an `enum`, the variants fields are public.

PR #517, opened 2019-08-16
value::GetError: add a constructor and make fields public

fengalin:
The constructor allows instantiating the error outside of the module, e.g. for use in an `assert_eq!`.

People might want to access the public fields when handling the error. Besides, if the error was implemented using an `enum`, the fields would be public.

Fixes https://github.com/gtk-rs/glib/issues/515


fengalin:
Failures on `nightly` & `beta` are unreleated. They most probably are a cause by https://github.com/gtk-rs/glib/pull/513 and might explain why it failed on Windows 32bits (they occur in tests `subclass::object::test::{test_signal_return_wrong_type, test_signal_return_wrong_object_type}`).

I'm on it, and will submit a separate PR.

fengalin:
I'll rebase once https://github.com/gtk-rs/glib/pull/519 is merged

PR #521, opened 2019-09-06
Remove subclass feature

GuillaumeGomez:
Part of https://github.com/gtk-rs/gtk/issues/887

The second commit is for a better dox support.

cc @sdroege @EPashkin


sdroege:
Note that this will break all the other crates. Better get all PRs ready first and then we merge them all at once to keep the time as small as possible during which stuff doesn't build :)

GuillaumeGomez:
Indeed. I'll send the other PRs tomorrow.

sdroege:
Forgot to mention: looks good to me otherwise :)

PR #525, opened 2019-10-01
Implement Value::transform()

sdroege:
And some minor fixes.

### DIFF
--- a/src/checksum.rs
+++ b/src/checksum.rs
@@ -68,5 +68,4 @@ mod tests {
         let vec = cs.get_digest();
         assert_eq!(vec, CS_SLICE);
     }
-
 }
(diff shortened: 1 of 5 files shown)


EPashkin:
@sdroege Thanks

sdroege:
Hrm

```
(process:7505): GLib-CRITICAL **: source_remove_from_context: assertion 'source_list != NULL' failed

thread '<unnamed>' panicked at 'Receiver without channel', src/libcore/option.rs:1190:5

test main_context_channel::tests::test_sync_channel ... FAILED
```

Unrelated to this change but should be investigated. Putting that on my list and merging.

sdroege:
reproduced here after running the tests in a loop for about 2 hours... ugly :)

sdroege:
This is https://gitlab.gnome.org/GNOME/glib/merge_requests/675#note_589306 . Needs new API in GLib for fixing it reliable, which I'm going to work on in the next days but of course means that we won't have this working reliable until GLib 2.64 at least.

Nothing we can do about that here.

EPashkin:
@sdroege Thanks for look at this issue

PR #527, opened 2019-10-22
Add clone macro

antoyo:


examples/clone.rs
```diff
@@ -0,0 +1,61 @@
+extern crate gio;
```
sdroege: Maybe move this to the examples repo?

sdroege: And use the new macro in the other examples :)

examples/clone.rs
```diff
@@ -0,0 +1,61 @@
...
+use gio::{ApplicationExt, ApplicationExtManual};
+use gtk::{
+    Application,
+    ApplicationWindow,
+    Button,
+    ButtonExt,
+    ContainerExt,
+    GtkWindowExt,
+    WidgetExt,
+};
```
sdroege: `use gtk::prelude::*` and `use gio::prelude::*`

examples/clone.rs
```diff
@@ -0,0 +1,61 @@
...
+use glib::clone;
+
+struct State {
+    started: bool,
+    count: i32,
+}
+
+impl State {
+    fn new() -> Self {
```
sdroege: Clippy will be sad. `[#derive(Default)]` and use that :)

examples/clone.rs
```diff
@@ -0,0 +1,61 @@
...
+
+fn main() {
+    let application = Application::new(
+        Some("com.github.gtk-rs.examples.basic"),
+        Default::default(),
+    ).expect("failed to initialize GTK application");
+
+    let state = Rc::new(RefCell::new(State::new()));
+
+    application.connect_activate(clone!(state => move |app| {
```
sdroege: An example with weak references would also be good

src/clone.rs
```diff
@@ -0,0 +1,62 @@
...
+    type Target = Rc<T>;
+
+    fn upgrade(&self) -> Option<Self::Target> {
+        self.upgrade()
+    }
+}
+
+#[macro_export]
+macro_rules! clone {
+    ($($n:ident),+ => move |$($p:pat),*| $body:expr) => (
```
sdroege: This needs configuration for weak vs strong references per variable

tests/clone_panic.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn new() -> Self {
+        Self {
+            started: false,
+            count: 0,
+        }
+    }
+}
+
+#[test]
+#[should_panic]
```
sdroege: Why will this panic?

antoyo: Because the state will be dropped before the main loop runs.

tests/clone_panic.rs
```diff
@@ -0,0 +1,65 @@
+extern crate gio;
+extern crate glib;
+extern crate gtk;
```
sdroege: These circular dependencies here are not ideal, even if it's just for the tests :) Can you remove them, or move this to the gtk repo?

src/clone.rs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            $( let $n = $crate::clone::Downgrade::downgrade(&$n); )+
+            move |$($p,)*| {
+                $(let $n = match $crate::clone::Upgrade::upgrade(&$n) {
+                    Some(val) => val,
+                    None => panic!("cannot upgrade weak reference `{}`", stringify!($n)),
```
sdroege: Instead of panicking, this should probably do something like the `upgrade_weak!` macro we have in some of the examples: return with some default value if upgrading fails.

Upgrade failure is not something that should panic, it can be completely normal and expected.

tests/clone_panic.rs
sdroege: Ah I see. See https://github.com/gtk-rs/glib/pull/527/files#r337428391 then :)

Cargo.toml
```diff
@@ -29,6 +29,8 @@ glib-sys = { git = "https://github.com/gtk-rs/sys" }
 gobject-sys = { git = "https://github.com/gtk-rs/sys" }
 
 [dev-dependencies]
+gio = { git = "https://github.com/gtk-rs/gio" }
+gtk = { git = "https://github.com/gtk-rs/gtk" }
```
EPashkin: Better don't do circular dependencies

src/clone.rs
antoyo: Not sure I understand how it can be normal.
If you try to access a variable that was freed, how could it not be a bug?
Do you have an example?

sdroege: Because it was freed already but e.g. your signal handler was called again nonetheless, so you just do nothing instead of exploding. This can happen during shutdown for example, when a signal handler is called right again before you disconnected it, etc.

The whole point of weak references is to be able to handle the non-existence of the underlying object gracefully.

antoyo: Do you want only one default return value even if there are multiple weak references?
I'm not sure how it could work otherwise (i.e. with weak references to widgets).

sdroege: Yes, only one. See the `upgrade_weak!` macro we have in some examples.

antoyo:
@sdroege @EPashkin @GuillaumeGomez I think this is ready for another round of review.

tests/clone.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+#[test]
+fn clone_closure() {
+    let state = Rc::new(RefCell::new(State::new()));
+    assert_eq!(state.borrow().started, false);
+
+    let closure = {
+        clone!(@weak state => move || {
```
sdroege: Maybe also add a test for strong, weak+strong mixed and the weak-failure-return-value :)

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[macro_export]
+macro_rules! to_return_value {
+    () => (());
+    ($value:expr) => ( $value );
+}
+
+#[macro_export]
+macro_rules! clone {
+    ($($(@ $weak:ident)? $variables:ident),+ => move || $body:block $(, $return_value:expr)? ) => (
```
sdroege: It might be less confusing to have the return value *before* the closure block. Additional arguments after a `{}` block are usually hard to read

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl<T> Upgrade for sync::Weak<T> {
+    type Target = Arc<T>;
+
+    fn upgrade(&self) -> Option<Self::Target> {
+        self.upgrade()
+    }
+}
+
+impl<T> Upgrade for rc::Weak<T> {
```
sdroege: These also need to be implemented in the `glib_wrapper` macro still

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            $( $crate::to_type_before!($(@ $weak)? $variables); )*
+            let return_value = $crate::to_return_value!($($return_value)?);
+            move || {
+                $( $crate::to_type_after!($(@ $weak)? $variables, return_value );)*
+                $body
+            }
+        }
+    );
+    ($($(@ $weak:ident)? $variables:ident),+ => move | $($pattern:pat),* | $body:block $(, $return_value:expr)? ) => (
```
sdroege: Does this allow weak and strong references interleaved or do all weak ones have to be at the front?

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[macro_export]
+macro_rules! to_type_after {
+    (_ , $return_value:expr) => ();
+    ($variable:ident , $return_value:expr) => ();
+    (@weak $variable:ident , $return_value:expr) => (
+        let $variable = match $crate::clone::Upgrade::upgrade(&$variable) {
+            Some(val) => val,
+            None => return $return_value,
```
sdroege: Also this still allows for panicking :) You can pass `panic!("...")` as the return value and it will panic instead of returning

GuillaumeGomez: Is it an issue?

GuillaumeGomez: Is it an issue?

sdroege: No, just for your information

tests/clone.rs
antoyo: Here are the current tests:
weak only,
weak and strong (strong is by default, i.e. no keyword),
weak failure with return value.
What is missing?

src/clone.rs
antoyo: You can interleave them. Look at the tests.

src/clone.rs
sdroege: You could require prefixing the default return value with `@default-return` or something to get rid of the ambiguity

tests/clone.rs
sdroege: Nothing, I was reviewing before you added that last commit with the new tests :)

antoyo:
Tests fail because I use `?` in macros which is not supported in edition 2015.

src/clone.rs
antoyo: This is done. Please do another round of review :) .

antoyo:
So, there's a problem when implementing the trait in the other crates:
only traits defined in the current crate can be implemented for arbitrary types.

sdroege:
> So, there's a problem when implementing the trait in the other crates:
> only traits defined in the current crate can be implemented for arbitrary types.

You only have to implement it on the final type, not on all types in the class hierarchy. Otherwise you'd implement it multiple times for the same type anyway.

src/object.rs
```diff
@@ -386,6 +386,30 @@ glib_wrapper! {
     }
 }
 
+#[macro_export]
+macro_rules! glib_weak_impl {
```
sdroege: Maybe want to hide this from the docs? This shouldn't really be called directly :)

sdroege: Or do it like for all the other things and make it part of the main macro with a `@` selector for the case. Probably better to keep this consistent.

src/object.rs
```diff
@@ -386,6 +386,30 @@ glib_wrapper! {
...
+        #[doc(hidden)]
+        impl $crate::clone::Upgrade for $crate::object::WeakRef<$name> {
+            type Target = $name;
+
+            fn upgrade(&self) -> Option<Self::Target> {
+                self.upgrade()
+            }
+        }
+    };
+    ($name:ident, $ty:ty) => ();
```
sdroege: Why this case? Why do you only generate this for types that have an FFI class struct? That seems wrong, you need to generate it for all types and there's not really any relationship between having an FFI class struct and being able to have weak references.

src/clone.rs
```diff
@@ -0,0 +1,99 @@
+use std::rc::{self, Rc};
+use std::sync::{self, Arc};
+
+pub trait Downgrade {
```
sdroege: This whole file should get some docs :)

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+impl<T> Upgrade for rc::Weak<T> {
+    type Target = Rc<T>;
+
+    fn upgrade(&self) -> Option<Self::Target> {
+        self.upgrade()
+    }
+}
+
+#[macro_export]
+macro_rules! to_type_before {
```
sdroege: Hide all macros from the docs should not be called from external code

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[macro_export]
+macro_rules! to_type_after {
+    (_ , $return_value:expr) => ();
+    ($variable:ident , $return_value:expr) => ();
+    (@weak $variable:ident , $return_value:expr) => (
+        let $variable = match $crate::clone::Upgrade::upgrade(&$variable) {
+            Some(val) => val,
+            None => return ($return_value)(),
```
sdroege: Why the `()`?

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn upgrade(&self) -> Option<Self::Target> {
+        self.upgrade()
+    }
+}
+
+#[macro_export]
+macro_rules! to_type_before {
+    (_) => ();
+    ($variable:ident) => (
+        let $variable = $variable.clone();
```
sdroege: If you want to be super-correct, call `clone()` by naming the correct trait from `std` :)

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+    );
+    (@weak $variable:ident) => (
+        let $variable = $crate::clone::Downgrade::downgrade(&$variable);
+    );
+}
+
+#[macro_export]
+macro_rules! to_type_after {
+    (_ , $return_value:expr) => ();
```
sdroege: What's this case for?

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+macro_rules! to_return_value {
+    () => (());
+    ($value:expr) => ( $value );
+}
+
+#[macro_export]
+macro_rules! clone {
+    ($($(@ $weak:ident)? $variables:ident),+ => $(@default-return $return_value:expr,)? move || $body:block ) => (
+        {
+            $( $crate::to_type_before!($(@ $weak)? $variables); )*
```
sdroege: While unpacking the indents, you might want to match on the identifier being `self` and in that case evaluate to the `compiler_error!()` macro to give a more meaningful error message.

src/object.rs
antoyo: I thought I read somewhere that only final types don't have a FFI class.

src/clone.rs
antoyo: Because I create a closure for the return value: this is to make it work with panics (you wouldn't want to panic before calling the closure).

src/clone.rs
antoyo: For the wildcard pattern used in the callback.

src/clone.rs
```diff
@@ -0,0 +1,99 @@
...
+    type Target = Rc<T>;
+
+    fn upgrade(&self) -> Option<Self::Target> {
+        self.upgrade()
+    }
+}
+
+#[macro_export]
+macro_rules! to_type_before {
+    (_) => ();
```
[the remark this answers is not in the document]
antoyo: For the wildcard pattern used in the callback.

src/object.rs
sdroege: That's correct but why do you only want to do this for final types? That would mean that you couldn't use the macro for `gio::Cancellable` for example.

sdroege:
@antoyo You're going to work on the remaining parts?

antoyo:
@sdroege: I won't have time soon.

sdroege:
Should be good to go now

src/object.rs
```diff
@@ -386,6 +386,30 @@ glib_wrapper! {
...
+        impl $crate::clone::Downgrade for $name {
+            type Target = $crate::object::WeakRef<Self>;
+
+            fn downgrade(&self) -> Self::Target {
+                <Self as $crate::object::ObjectExt>::downgrade(&self)
+            }
+        }
+
+        #[doc(hidden)]
+        impl $crate::clone::Upgrade for $crate::object::WeakRef<$name> {
```
sdroege: @antoyo How did this compile for you? We can't implement this outside of the glib trait because `WeakRef` and `Upgrade` are both declared in the glib trait. So compiling e.g. gtk with this change fails :)

I tried moving everything into a single trait but that causes some other problems.

sdroege: Fixed

sdroege:
@GuillaumeGomez Green!

sdroege:
@GuillaumeGomez Also let me know if you want *me* to merge it ;)

src/clone.rs
```diff
@@ -0,0 +1,187 @@
...
+macro_rules! to_return_value {
+    () => {
+        ()
+    };
+    ($value:expr) => {
+        $value
+    };
+}
+
+/// Macro for passing variables  as strong or weak references into a closure.
```
GuillaumeGomez: "variables  as strong" -> there is one extra whitespace after "variables".

src/clone.rs
```diff
@@ -0,0 +1,187 @@
...
+/// This macro can be useful in combination with closures, e.g. signal handlers, to reduce the
+/// boilerplate required for passing strong or weak references into the closure. It will
+/// automatically create the new reference and pass it with the same name into the closure.
+///
+/// If upgrading the weak reference to a strong reference inside the closure is failing, the
+/// closure is immediately returning an optional default return value. If none is provided, `()` is
+/// returned.
+///
+///  ### Passing a strong reference
+/// ```rust
```
GuillaumeGomez: Please add an empty line before the example (makes the code simpler to read).

src/clone.rs
```diff
@@ -0,0 +1,187 @@
...
+/// let v = Rc::new(1);
+/// let closure = clone!(v => move |x| {
+///     println!("v: {}, x: {}", v, x);
+/// });
+///
+/// closure(2);
+/// ```
+///
+///  ### Passing a strong and weak reference
+/// ```rust
```
GuillaumeGomez: Please add an empty line before the example (makes the code simpler to read).

src/clone.rs
```diff
@@ -0,0 +1,187 @@
...
+/// let u = Rc::new(2);
+/// let closure = clone!(v, @weak u => move |x| {
+///     println!("v: {}, u: {}, x: {}", v, u, x);
+/// });
+///
+/// closure(3);
+/// ```
+///
+///  ### Providing a default return value if upgrading a weak reference fails
+/// ```rust
```
GuillaumeGomez: Please add an empty line before the example (makes the code simpler to read).

PR #534, opened 2019-11-14
Extend clone macro

### LINKED ISSUE #532: Default behavior of the clone macro
I just realized something: by default, we still clone the content, not using the downgrade/upgrade mechanisms. From my point of view, we should do the opposite: downgrading/upgrading by default or cloning by specifying `@clone`.

cc @sdroege @EPashkin @antoyo

### LINKED ISSUE #533: panic as clone return value?
In some cases, if the upgrading fails, we'd like to panic. For now it looks like this:

```rust
let listbox = gtk::ListBox::new();
listbox.bind_model(Some(&model),
    clone!(@weak window => @default-return panic!("failed to upgrade"), move |item| {
        // blablabla
    }));
```

What about adding another macro to handle such case? Like `clone_or_panic!` or something along the line?

GuillaumeGomez:
Fixes #532.
Fixes #533.

cc @sdroege @EPashkin @antoyo

### DIFF
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -232,19 +232,6 @@ macro_rules! to_return_value {
 ///         });
 ///     }
 /// }
-/// ```
-///
-/// **Using another keyword than `weak` or `strong`**:
-///
-/// ```compile_fail
-/// # use glib::clone;
-/// # use std::rc::Rc;
-/// let v = Rc::new(1);
-///
-/// let closure = clone!(@hello v => move |x| {
-///     println!("v: {}, x: {}", v, x);
-/// });
-/// ```
 #[macro_export]
 macro_rules! clone {
     ($($(@ $strength:ident)? $variables:ident),+ => @default-panic, move || $body:block ) => (
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -87,6 +87,10 @@ macro_rules! to_type_before {
     (@weak $variable:ident) => (
         let $variable = $crate::clone::Downgrade::downgrade(&$variable);
     );
+    (@ $keyword:ident $variable:ident) => (
+        compile_error!("Unknown keyword `{}`, only `weak` and `strong` are allowed",
+                       stringify!($keyword));
+    );
 }
 
 #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


sdroege:
You might also want to add another branch to the macro after the `$variable:ident` case for `$variable:expr`, which then maps to `compile_error!` for giving a better error message for things like `x.y` as variable.

Did you test that all the other parts are working? The syntax looks fine to me in any case. Maybe we can add some tests for this (and fix the existing tests)? :)

GuillaumeGomez:
`cargo test` is running fine on my computer?

GuillaumeGomez:
Oh you meant the examples maybe? Well, the PR is already open, just not up to date.

GuillaumeGomez:
I also added a check for the keyword (better than having a macro error. :) )

src/clone.rs
```diff
@@ -223,12 +227,24 @@ macro_rules! to_return_value {
...
 ///
-///         let closure = clone!(self => move |x| {
+///         let closure = clone!(@strong self => move |x| {
 ///             println!("v: {}, x: {}", v, x);
 ///         });
 ///     }
 /// }
 /// ```
+///
+/// **Using another keyword than `weak` or `strong`**:
```
sdroege: Why do you put an example in the docs for this? :) Nobody should ever do that and having that in the docs only distracts really.

sdroege:
@GuillaumeGomez Improved the error handling a bit and also fixed a bug. This should be good to go I guess :)

src/clone.rs
```diff
@@ -69,33 +69,40 @@ impl<T> Upgrade for rc::Weak<T> {
...
+        compile_error!("You need to specify if this is a weak or a strong clone.");
+    );
+    (@strong $variable:ident) => (
         let $variable = $variable.clone();
     );
     (@weak $variable:ident) => (
         let $variable = $crate::clone::Downgrade::downgrade(&$variable);
     );
+    (@ $keyword:ident $variable:ident) => (
+        compile_error!("Unknown keyword, only `weak` and `strong` are allowed");
```
GuillaumeGomez: The `stringify!(keyword)` wasn't working?

sdroege:
>The `stringify!(keyword)` wasn't working?

It was but compile_error! does not allow formatting currently, it requires a single string.

Maybe you can get around that with concat! etc, give it a try :)

-- 

GuillaumeGomez:
Na, not worth it. Looks good to me now. Once merged, I'll update my examples PR.

sdroege:
Merged, please fix the examples then :) Thanks!

GuillaumeGomez:
Will do before the end of the day!

PR #535, opened 2019-11-16
Support downgrade on references as well

GuillaumeGomez:


GuillaumeGomez:
Tests passed, need to fix examples now. Merging since @sdroege told me to do so once ok.

PR #536, opened 2019-11-17
Add renaming support

GuillaumeGomez:
I'll add the enforced fields renaming in the same PR. Running a few tests in the meantime. :)

### DIFF
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -77,7 +77,7 @@ impl<T> Upgrade for rc::Weak<T> {
 #[macro_export]
 macro_rules! to_type_before {
     (_) => ();
-    ($variable:ident $(as $rename:ident)?) => (
+    ($($variable:ident).+ $(as $rename:ident)?) => (
         compile_error!("You need to specify if this is a weak or a strong clone.");
     };
     (@strong $variable:ident) => {
@@ -86,13 +86,13 @@ macro_rules! to_type_before {
     (@weak $variable:ident) => {
         let $variable = $crate::clone::Downgrade::downgrade(&$variable);
     );
-    (@strong $variable:ident as $rename:ident) => (
-        let $rename = $variable.clone();
+    (@strong $($variable:ident).+ as $rename:ident) => (
+        let $rename = $($variable).+.clone();
     );
-    (@weak $variable:ident as $rename:ident) => (
-        let $rename = $crate::clone::Downgrade::downgrade(&$variable);
+    (@weak $($variable:ident).+ as $rename:ident) => (
+        let $rename = $crate::clone::Downgrade::downgrade(&$($variable).+);
     );
-    (@ $keyword:ident $variable:ident $(as $rename:ident)?) => (
+    (@ $keyword:ident $($variable:ident).+ $(as $rename:ident)?) => (
         compile_error!("Unknown keyword, only `weak` and `strong` are allowed");
     };
 }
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Should be complete now. Just wondering about the formatting stuff...

src/clone.rs
```diff
@@ -163,6 +179,21 @@ macro_rules! to_return_value {
...
 ///
+/// ### Renaming variables
+///
+/// ```
+/// use glib::clone;
+/// use std::rc::Rc;
+///
+/// let v = Rc::new(1);
+/// let u = Rc::new(2);
+/// let closure = clone!(@strong v as y, @weak u => move |x| {
```
EPashkin: @GuillaumeGomez  can you add test with two @strong renamed variable too?

GuillaumeGomez: What would it change?

EPashkin: Just to show what syntax used in this case

src/clone.rs
```diff
@@ -246,63 +244,131 @@ macro_rules! to_return_value {
...
+/// # use std::rc::Rc;
+/// #[derive(Debug)]
+/// struct Foo {
+///     v: Rc<usize>,
+/// }
+///
+/// impl Foo {
+///     fn foo(&self) {
+///         let closure = clone!(@strong self.v => move |x| {
+///             println!("v: {:?}", v);
```
sdroege: I don't think we should allow that. It allows annoying bugs to sneak in

```rust
clone!(@strong self.v, self.bla.v => move |x| {
    // what's `v` here now? Might even be the same type so the closure would silently misbehave
});
```

I believe we should require renaming for `self` and field accesses and not come up with any magic names ourselves.

tests/clone.rs
```diff
@@ -36,6 +36,27 @@ fn clone_and_references() {
...
+fn subfields_renaming() {
+    struct Foo {
+        v: Rc<usize>,
+    }
+
+    impl Foo {
+        fn foo(&self) {
+            let state = Rc::new(RefCell::new(State::new()));
+
+            let closure = clone!(@strong self.v as v, @weak state as hello => move |_| {
```
sdroege: Can you also add a test for `self` renaming?

src/clone.rs
GuillaumeGomez: We don't. The block is marked as "compile_fail" you know. :p

tests/clone.rs
GuillaumeGomez: Already done in the docs.

src/clone.rs
GuillaumeGomez: I'm not sure to follow...

EPashkin: I was just not sure if this case work or need two `@strong`, now you can ignore this comment
```
clone!(@strong a as b, c as d => ...)
```

GuillaumeGomez: The example you wrote doesn't work either: you have to add `@strong` or `@weak` before each variable.

src/clone.rs
sdroege: Ah please remove this from the docs then. People like me assume that examples in the docs actually work :) I didn't see that annotation

tests/clone.rs
sdroege: Ok!

src/clone.rs
GuillaumeGomez: People reading the docs will see the "this code doesn't compile" element. rustdoc stuff you know. :p

sdroege: How does that look like in the docs, can you show an example in e.g. the std docs? It seems strange to me to include broken code examples in the docs.

sdroege: Ok you convinced me

src/clone.rs
sdroege: @EPashkin anything missing here?

EPashkin: IMHO all good

EPashkin: except for new merge conflict 😢

GuillaumeGomez:
Urg. Where is this conflict coming from?! ><

fengalin:
@GuillaumeGomez: sorry, my fault, I [ran `‘fmt`](https://github.com/gtk-rs/glib/commit/410c4ae5dedc747e8898c3dbf28241646ee5ec62).

GuillaumeGomez:
Fixed the conflict. Let's see if the CI is still happy.

GuillaumeGomez:
@fengalin `fmt` is definitely my personal nemesis...

PR #541, opened 2019-11-18
API additions for connecting non-Send closures and thread-safety fixes to the main context channel and futures

sdroege:
After checking that the callback of the main context channel was always dropped from the correct thread, this check was actually failing often. I refactored the code now so that this can't happen anymore and we always ensure that non-Send values are only ever accessed from the correct thread.

Similar things happened with the GSource in the futures executor, which is also fixed and refactored now.

Overall the code should be much simpler now, and more correct!

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -10,6 +10,7 @@ use std::slice;
 
 use libc::{c_uint, c_void};
 
+use get_thread_id;
 use gobject_sys;
 use translate::{from_glib_none, mut_override, ToGlibPtr, ToGlibPtrMut, Uninitialized};
 use types::Type;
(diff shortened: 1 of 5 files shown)


sdroege:
CC @EPashkin @GuillaumeGomez

GuillaumeGomez:
Code looks good, thanks a lot! Now we need tests. :)

sdroege:
> Code looks good, thanks a lot! Now we need tests. :)

We already have tests. The bugs were caught by the existing tests after adding the checks for doing things from the correct thread :)

GuillaumeGomez:
If the CI was ok before, then it means it's not checked correctly. Please either update the travis build or add tests. :D

sdroege:
On 19 November 2019 00:31:33 EET, Guillaume Gomez <notifications@github.com> wrote:
>If the CI was ok before, then it means it's not checked correctly.
>Please either update the travis build or add tests. :D

Run the tests without the last commits and you'll see them failing if you run them often enough.

They fail with the addition of the first commits because that's adding the checks for usage and dropping on the correct thread. Previously that was not checked and would've caused safety issues instead.

-- 

EPashkin:
@sdroege Thanks.
There was one job failed due network, I restarted it

PR #542, opened 2019-11-20
Remove Send bound from SourceFuture/SourceStream

sdroege:
The resulting Future/Stream will simply not be Send if the closure to
create it was not Send


sdroege:
@GuillaumeGomez @EPashkin Ok? Needed for https://github.com/gtk-rs/gio/pull/258

sdroege:
@GuillaumeGomez CI failure comes from your rename gtk commit, you need to fix the examples too now :)

Issue #543, opened 2019-11-23
KeyFile::get_string() leaks memory on invalid UTF8


sdroege:
Code in question is https://github.com/gtk-rs/glib/blob/547887ff8e1ea1672b9d5c20bf6972012a503a5e/src/auto/key_file.rs#L248-L263

If there's invalid UTF8, `error` will be set but we will also get the invalid string as return value and are supposed to free it.

See for example the tests in GLib: https://gitlab.gnome.org/GNOME/glib/blob/7b774398414f2bf35115603da4e3badd45ffb025/glib/tests/keyfile.c#L558-564

sdroege:
The same probably also applies to `get_locale_string()`, `get_string_list()` and `get_locale_string_list()`.

sdroege:
Also `get_value()` / `set_value()` must not work with strings at all. The whole point of these functions is to work with plain bytes, potentially invalid UTF8. Have to work on `&[u8]` / `Vec<u8>` or similar.

EPashkin:
good catch

sdroege:
Actually `set_value()` / `get_value()` are still required to have valid UTF-8, so that part at least is fine.

PR #544, opened 2019-11-24
KeyFile::get_string() can return a string in error case that still ha…

sdroege:
…s to be freed

If it's an invalid UTF-8 string for example we still get it returned but
should simply free it. We can't return it as a String in any way.

Fixes https://github.com/gtk-rs/glib/issues/543


EPashkin:
@sdroege Thanks, 👍

Issue #545, opened 2019-12-04
Calls into object subclasses that are still floating can cause reference issues


sdroege:
For virtual methods we ensure that the object is still floating once the virtual method returns and not floating during the scope of the virtual method. However the same problem would happen if other callbacks with the floating object would happen, any kind of callback. The first use in the bindings in a `transfer none` context would sink the floating reference and cause memory leaks or worse.

Unclear how to handle this other than using the floating reference guard in all callbacks, signal handler trampolines, etc.

sdroege:
As this now actually causes crashes in practice and I have a plan that improves the situation at least, let's block the release on this too.

PR #547, opened 2019-12-07
Manually implement FFI code for GObject instead of using glib_shared_wrapper!

sdroege:
This way we can implement special cases more correctly:
    - g_value_get_boxed() is invalid for GObjects
    - Debug impl can print the actual type
    - Only from_glib_none() will ever steal floating references and e.g.
      cloning does not, see https://github.com/gtk-rs/glib/issues/545
    
    After this the only places where we would steal floating references from
    C code would be when explicitly calling from_glib_none() somewhere. That
    is, when we call a function and get a (transfer none) reference to an
    object back. (transfer floating) is an alias for this.
    
    Especially, cloning and from_glib_borrow() would never steal floating
    references, and because of this it is very important to only ever use
    from_glib_borrow() inside callbacks, signal handlers and virtual method
    trampolines.

----

@EPashkin @GuillaumeGomez @federicomenaquintero @MathieuDuponchelle @ebassi Does the above make sense to you?

This is basically the workaround for C code passing us floating references that we should not actually own. We try now very hard to not steal them, i.e. the only place where we would steal them now is if a function is returning a `(transfer none)` object to use (return value or out parameter), or more specifically if `from_glib_none()` is used. We won't anymore steal floating references passed into: callbacks, signal handlers, virtual methods and passed to us via `GValue`s but keep the floating flag intact to have C worry about this mess instead.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -6,6 +6,7 @@
 
 use glib_sys;
 use gobject_sys;
+use std::cmp;
 use std::fmt;
 use std::hash;
 use std::marker::PhantomData;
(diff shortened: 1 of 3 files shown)


EPashkin:
@sdroege Sound reasonable, thanks

sdroege:
> WIP: Make FloatingReferenceGuard a no-op

This commit I'd like to get rid of in a separate PR. That way the CI can stay green all the time, @GuillaumeGomez

GuillaumeGomez:
Sounds good for me as well, thanks!

MathieuDuponchelle:
Addressed the issue I had in my C test suite (uses `GstHarness`, which takes a simple ref and doesn't actually take ownership of the floating ref). Well done :)

PR #550, opened 2019-12-11
Regen

GuillaumeGomez:
Part of https://github.com/gtk-rs/atk/pull/45


Gir.toml
```diff
@@ -339,6 +339,10 @@ status = "generate"
     pattern = "ref_string_.+"
     #need manual implementation
     ignore = true
+    [[object.function]]
+    name = "spawn_async_with_fds"
```
sdroege: Also `spawn_async_with_pipes` for the return value integers

src/functions.rs
```diff
@@ -0,0 +1,115 @@
...
+use Error;
+use Pid;
+use SpawnFlags;
+#[cfg(not(windows))]
+use std::os::unix::io::AsRawFd;
+#[cfg(windows)]
+use ::std::os::windows::io::AsRawHandle;
+
+#[cfg(not(windows))]
+pub fn spawn_async_with_fds<P: AsRef<std::path::Path>, T: AsRawFd, U: AsRawFd, V: AsRawFd>(
```
sdroege: This and the Windows version looks good to me

sdroege:
Windows CI is unhappy in a confusing way. I don't understand what's wrong there for `translate.rs`, that didn't change.

However the `as_raw_handle()` part in your code needs some adjustments.

EPashkin:
Error because new file ` src\functions.rs` now resolved as `::functions::`,
strange that travis passed.
IMHO test can use just `::path_get_basename` etc.

EPashkin:
Seems resolve broken because `src\functions.rs` not compiles only in Windows,
so rust registered "namespace" `::functions` but not filled it with functions.

EPashkin:
Windows still broken and test needed formating

src/functions.rs
```diff
@@ -0,0 +1,120 @@
...
+            working_directory.as_ref().to_glib_none().0,
+            argv.to_glib_none().0,
+            envp.to_glib_none().0,
+            flags.to_glib(),
+            child_setup,
+            Box_::into_raw(super_callback0) as *mut _,
+            child_pid.as_mut_ptr(),
+            stdin_fd.as_raw_handle() as usize,
+            stdout_fd.as_raw_handle() as usize,
+            stderr_fd.as_raw_handle() as usize,
```
sdroege: This actually wants an `i32`, but that would lose information. Not sure how this is supposed to work on Windows.

src/functions.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+
+pub fn spawn_async_with_pipes<P: AsRef<std::path::Path>>(
+    working_directory: P,
+    argv: &[&std::path::Path],
+    envp: &[&std::path::Path],
+    flags: SpawnFlags,
+    child_setup: Option<Box_<dyn FnOnce() + 'static>>,
+) -> Result<(Pid, File, File, File), Error> {
```
sdroege: These are not files :) Use `FromRawFd` here, I guess

src/functions.rs
```diff
@@ -0,0 +1,198 @@
...
+                ))
+            }
+            #[cfg(windows)]
+            {
+                use std::os::windows::io::{FromRawHandle, RawHandle};
+                Ok((
+                    child_pid,
+                    File::from_raw_handle(standard_input as usize as RawHandle),
+                    File::from_raw_handle(standard_output as usize as RawHandle),
+                    File::from_raw_handle(standard_error as usize as RawHandle),
```
sdroege: Is this correct? Is every file descriptor on Windows the same as a HANDLE?

src/functions.rs
```diff
@@ -0,0 +1,198 @@
...
+            working_directory.as_ref().to_glib_none().0,
+            argv.to_glib_none().0,
+            envp.to_glib_none().0,
+            flags.to_glib(),
+            child_setup,
+            Box_::into_raw(super_callback0) as *mut _,
+            child_pid.as_mut_ptr(),
+            stdin_fd.as_raw_handle() as usize as _,
+            stdout_fd.as_raw_handle() as usize as _,
+            stderr_fd.as_raw_handle() as usize as _,
```
sdroege: This loses precision: usize to i32 cast. This is probably not correct and needs investigation by someone on Windows.

How do we do this elsewhere?

sdroege: We don't do that anywhere else and don't seem to have any Windows-specific fd code. Someone needs to investigate how this is supposed to work on Windows.

There's this code in GStreamer for Windows: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/blob/90854eb0471e04a11c37fd5a5beaf7ab0cf29dfe/gstreamer/src/bus_windows.rs#L41
But I think this is equally suspicious. It's casting an `i32` to a `HANDLE` (i.e. a pointer).

I also found some code inside C GStreamer though, which does similar things and that code is actually known to work on Windows. So maybe this is all fine? Needs some investigation from someone who knows Windows :) @EPashkin ?

sdroege: Ah the part in GStreamer works because of https://github.com/gtk-rs/sys/blob/94a8fcf7119c380e27510f1bef4c546964804ef8/glib-sys/src/manual.rs#L18 . So that works all fine.

Unclear how it's possible to use these functions here on Windows though.

EPashkin: I will try to look at the weekend

sdroege: Let's defer it to next release then. We said three weeks ago to not add any new features for this release anyway, and here we are nonetheless :) Let's just ignore these functions on Windows or completely for this release, and then figure it out without rushing it afterwards.

GuillaumeGomez: Then let's ignore for windows. It works on unix so that'd be a shame not keep them. And yes, didn't see them when I made the regen, sorry...

EPashkin: @GuillaumeGomez Then please remove windows part or add todo with comment about double checking.
Also strange than `cargo +stable fmt` passed on Appveyor but fail on travis

GuillaumeGomez: I'll comment the code. About rustfmt, it's my local run. So I don't know what travis is doing...

src/functions.rs
sdroege: Still valid

src/functions.rs
```diff
@@ -0,0 +1,198 @@
...
+//         let child_pid = from_glib(child_pid.assume_init());
+//         if error.is_null() {
+//             Ok(child_pid)
+//         } else {
+//             Err(from_glib_full(error))
+//         }
+//     }
+// }
+
+pub fn spawn_async_with_pipes<P: AsRef<std::path::Path>>(
```
sdroege: Probably also not on Windows

sdroege: Not just "probably", same problem. So definitely not on Windows for now :)

GuillaumeGomez: I disabled the whole file for windows. ;)

src/functions.rs
GuillaumeGomez: Ok!

src/functions.rs
sdroege: Ah that also works, but makes the `#[cfg(not(windows))]` further up not required anymore :)

src/lib.rs
```diff
@@ -150,6 +150,10 @@ mod checksum;
 pub mod closure;
 mod enums;
 mod file_error;
+#[cfg(not(windows))]
+mod functions;
+#[cfg(not(windows))]
+pub use functions::*;
```
sdroege: This is a potential footgun. Someone who might want to add a new function there later might miss this and then it's only available for non-Windows :)

GuillaumeGomez: Ok.

sdroege:
:+1:

sdroege:
On Windows:

```
warning: unused import: `std::os::windows::io::AsRawHandle`
  --> src\functions.rs:17:5
   |
17 | use std::os::windows::io::AsRawHandle;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
warning: unused import: `std::os::windows::io::AsRawHandle`
  --> src\functions.rs:17:5
   |
17 | use std::os::windows::io::AsRawHandle;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
    Finished dev [unoptimized + debuginfo] target(s) in 37.29s
```

PR #556, opened 2019-12-18
Implement Send/Sync for more types where it's correct

sdroege:
CC @GuillaumeGomez @EPashkin


GuillaumeGomez:
For such implementations, I'd expect some tests to be sure nothing weird happens.

sdroege:
> For such implementations, I'd expect some tests to be sure nothing weird happens.

There's nothing that can be tested here, that's why it's `unsafe` :)

GuillaumeGomez:
Hum. :-/

EPashkin:
@sdroege Thanks, 👍

PR #558, opened 2019-12-20
More compilation errors for clone macro

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -330,6 +330,21 @@ macro_rules! to_return_value {
 /// ```
 #[macro_export]
 macro_rules! clone {
+    ( => $($_:tt)*) => (
+        // In case we have:
+        // clone!( => move || {});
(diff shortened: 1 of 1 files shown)


src/clone.rs
```diff
@@ -330,6 +330,21 @@ macro_rules! to_return_value {
...
+    );
+    ($(move)? || $($_:tt)*) => (
+        // In case we have:
+        // clone!(|| {});
+        compile_error!("If you have nothing to clone, no need to use this macro!");
+    );
+    ($(move)? | $($pattern:pat),* | $($_:tt)*) => (
+        // In case we have:
+        // clone!(|a, b| {});
+        compile_error!("If you have nothing to clone, no need to use this macro!")
```
sdroege: Or maybe we can allow these 3 cases and just do nothing? That could help during development when trying things

PR #559, opened 2019-12-27
Add PtrArray and related trait implementations

majorz:
Needed for creating bindings of the NetworkManager client library which heavily relies on GPtrArray.

This is the PtrArray part of #482 and was taken directly from there. On top of it I run `cargo fmt`.

Tested with a modified gir that includes PtrArray. Going to PR that part as well once PtrArray is included in glib.

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -1003,6 +1003,88 @@ where
     }
 }
 
+pub struct PtrArray(*mut glib_sys::GPtrArray);
+
+impl Drop for PtrArray {
+    fn drop(&mut self) {
+        unsafe {
+            glib_sys::g_ptr_array_unref(self.0);
+        }
+    }
+}
+
+impl<'a, T> ToGlibContainerFromSlice<'a, *mut glib_sys::GPtrArray> for T
+where
+    T: GlibPtrDefault + ToGlibPtr<'a, <T as GlibPtrDefault>::GlibType>,
+{
+    type Storage = (
+        Option<PtrArray>,
+        Vec<Stash<'a, <T as GlibPtrDefault>::GlibType, T>>,
+    );
+
+    #[inline]
+    fn to_glib_none_from_slice(t: &'a [T]) -> (*mut glib_sys::GPtrArray, Self::Storage) {
+        let stash_vec: Vec<_> = t.iter().map(ToGlibPtr::to_glib_none).collect();
+        let arr = unsafe { glib_sys::g_ptr_array_sized_new(t.len() as _) };
+        unsafe {
+            for stash in &stash_vec {
+                glib_sys::g_ptr_array_add(arr, Ptr::to(stash.0));
+            }
+        }
+        (arr, (Some(PtrArray(arr)), stash_vec))
+    }
+
+    #[inline]
+    fn to_glib_container_from_slice(t: &'a [T]) -> (*mut glib_sys::GPtrArray, Self::Storage) {
+        let stash_vec: Vec<_> = t.iter().map(ToGlibPtr::to_glib_none).collect();
+        let arr = unsafe { glib_sys::g_ptr_array_sized_new(t.len() as _) };
+        unsafe {
+            for stash in &stash_vec {
+                glib_sys::g_ptr_array_add(arr, Ptr::to(stash.0));
+            }
+        }
+        (arr, (None, stash_vec))
+    }
+
+    #[inline]
+    fn to_glib_full_from_slice(t: &[T]) -> *mut glib_sys::GPtrArray {
+        let arr = unsafe { glib_sys::g_ptr_array_sized_new(t.len() as _) };
+        unsafe {
+            for ptr in t.iter().map(ToGlibPtr::to_glib_full) {
+                glib_sys::g_ptr_array_add(arr, Ptr::to(ptr));
+            }
+        }
+        arr
+    }
+}
+
+impl<'a, T> ToGlibContainerFromSlice<'a, *const glib_sys::GPtrArray> for T
+where
+    T: GlibPtrDefault + ToGlibPtr<'a, <T as GlibPtrDefault>::GlibType>,
+{
+    type Storage = (
+        Option<PtrArray>,
+        Vec<Stash<'a, <T as GlibPtrDefault>::GlibType, T>>,
+    );
+
+    #[inline]
+    fn to_glib_none_from_slice(t: &'a [T]) -> (*const glib_sys::GPtrArray, Self::Storage) {
+        let (arr, stash) =
+            ToGlibContainerFromSlice::<*mut glib_sys::GPtrArray>::to_glib_none_from_slice(t);
+        (arr as *const glib_sys::GPtrArray, stash)
+    }
+
+    #[inline]
+    fn to_glib_container_from_slice(_t: &'a [T]) -> (*const glib_sys::GPtrArray, Self::Storage) {
+        unimplemented!()
+    }
+
+    #[inline]
+    fn to_glib_full_from_slice(_t: &[T]) -> *const glib_sys::GPtrArray {
+        unimplemented!()
+    }
+}
+
 /// Translate a simple type.
 pub trait FromGlib<T>: Sized {
     fn from_glib(val: T) -> Self;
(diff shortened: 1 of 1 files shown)


src/ptr_array.rs
```diff
@@ -0,0 +1,218 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use glib_sys;
+use std::fmt;
+use translate::*;
+
+glib_wrapper! {
+    pub struct PtrArray(Shared<glib_sys::GPtrArray>);
```
sdroege: If we want to have proper `GPtrArray` bindings then this needs to generic over the contained pointer type and have the corresponding memory management handling for that type.

Unless you want to go that route, which is quite some effort, I'd recommend not having this part here but only the trait impls below for converting between pointer array and `Vec<T>` / `&[T]`. You can also move those trait impls into `translate.rs` next to the ones for `GList` and friends then.

src/ptr_array.rs
```diff
@@ -0,0 +1,218 @@
...
+                glib_sys::g_ptr_array_add(arr, ptr as *mut _);
+            }
+
+            (arr, (None, stash_vec))
+        }
+    }
+
+    #[inline]
+    fn to_glib_full_from_slice(_t: &[T]) -> *mut glib_sys::GPtrArray {
+        unimplemented!()
```
sdroege: This seems like the one that is actually useful. The others are all potential footguns

src/ptr_array.rs
```diff
@@ -0,0 +1,218 @@
...
+    T: GlibPtrDefault
+        + FromGlibPtrNone<<T as GlibPtrDefault>::GlibType>
+        + FromGlibPtrFull<<T as GlibPtrDefault>::GlibType>,
+{
+    unsafe fn from_glib_none_num_as_vec(ptr: *mut glib_sys::GPtrArray, num: usize) -> Vec<T> {
+        if num == 0 || ptr.is_null() {
+            return Vec::new();
+        }
+        let pdata = (*ptr).pdata;
+        let mut res = Vec::with_capacity(num);
```
sdroege: Should first of all check here that the pointer array has `num` items available at all. Should probably become an assertion

src/ptr_array.rs
majorz: Now everything is moved to translate.rs

src/ptr_array.rs
majorz: This is now implemented together with the rest of the methods.

src/ptr_array.rs
majorz: Added asserts at the two places I am using `with_capacity`.

majorz:
@sdroege I covered all the previous review items. Additionally I tested some of my NetworkManager examples with Valgrind and so far so good. Please check again whenever possible and let me know if I need to do more adjustments. Thanks for reviewing that!

src/translate.rs
```diff
@@ -1003,6 +1003,88 @@ where
     }
 }
 
+pub struct PtrArray(*mut glib_sys::GPtrArray);
```
sdroege: This type is not used anywhere, is it? Should be removed then :)

src/translate.rs
```diff
@@ -1003,6 +1003,88 @@ where
...
+
+impl Drop for PtrArray {
+    fn drop(&mut self) {
+        unsafe {
+            glib_sys::g_ptr_array_free(self.0, false.to_glib());
```
sdroege: This should be with `true`, otherwise you leak memory. But more important, `gst_ptr_array_unref()`

src/translate.rs
sdroege: I mean, make it not public if it's only used internally here

src/translate.rs
```diff
@@ -1855,6 +1937,120 @@ impl FromGlibPtrContainer<*const c_char, *mut glib_sys::GHashTable> for HashMap<
     }
 }
 
+impl<T> FromGlibContainerAsVec<<T as GlibPtrDefault>::GlibType, *mut glib_sys::GPtrArray> for T
+where
+    T: GlibPtrDefault
+        + FromGlibPtrNone<<T as GlibPtrDefault>::GlibType>
+        + FromGlibPtrFull<<T as GlibPtrDefault>::GlibType>,
```
sdroege: You duplicated this trait bound

GuillaumeGomez: Na, just enforced it even more!

src/translate.rs
```diff
@@ -1855,6 +1937,120 @@ impl FromGlibPtrContainer<*const c_char, *mut glib_sys::GHashTable> for HashMap<
...
+                res.push(from_glib_none(item_ptr));
+            }
+        }
+        res
+    }
+
+    unsafe fn from_glib_container_num_as_vec(ptr: *mut glib_sys::GPtrArray, num: usize) -> Vec<T> {
+        let res = FromGlibContainer::from_glib_none_num(ptr, num);
+        if !ptr.is_null() {
+            glib_sys::g_ptr_array_free(ptr, false.to_glib());
```
sdroege: `g_ptr_array_unref()`

src/translate.rs
```diff
@@ -1855,6 +1937,120 @@ impl FromGlibPtrContainer<*const c_char, *mut glib_sys::GHashTable> for HashMap<
...
+        let pdata = (*ptr).pdata;
+        assert!((*ptr).len as usize >= num);
+        let mut res = Vec::with_capacity(num);
+        for i in 0..num {
+            let item_ptr: <T as GlibPtrDefault>::GlibType = Ptr::from(ptr::read(pdata.add(i)));
+            if !item_ptr.is_null() {
+                res.push(from_glib_full(item_ptr));
+            }
+        }
+        glib_sys::g_ptr_array_free(ptr, false.to_glib());
```
sdroege: `g_ptr_array_unref()`

src/translate.rs
```diff
@@ -1855,6 +1937,120 @@ impl FromGlibPtrContainer<*const c_char, *mut glib_sys::GHashTable> for HashMap<
...
+        if num == 0 || ptr.is_null() {
+            return Vec::new();
+        }
+        let pdata = (*ptr).pdata;
+        assert!((*ptr).len as usize >= num);
+        let mut res = Vec::with_capacity(num);
+        for i in 0..num {
+            let item_ptr: <T as GlibPtrDefault>::GlibType = Ptr::from(ptr::read(pdata.add(i)));
+            if !item_ptr.is_null() {
+                res.push(from_glib_full(item_ptr));
```
sdroege: This is a bit problematic as we free the array, and it might then free its items with the free function configured on it.

I think for the time being we have to use `from_glib_none()` here until we have high-level bindings for the pointer array instead of converting to a vec.

sdroege:
Looks good to me apart from those little things. Thanks for updating :)

src/translate.rs
majorz: Looks like it needs to be public:
```
error[E0446]: private type `translate::PtrArray` in public interface
    --> /home/majorz/resin/libnm-rs-build/glib/src/translate.rs:1020:5
     |
1006 |   struct PtrArray(*mut glib_sys::GPtrArray);
     |   - `translate::PtrArray` declared as private
...
1020 | /     type Storage = (
1021 | |         Option<PtrArray>,
1022 | |         Vec<Stash<'a, <T as GlibPtrDefault>::GlibType, T>>,
1023 | |     );
     | |______^ can't leak private type
```
It is the same for the structs `SList` and `List`.

src/translate.rs
majorz: You are indeed correct. I looked into it more closely and found a case where an array was passed as reference and it was leaking. It is changed now to `true` and does not leak anymore.

src/translate.rs
majorz: Right, I copied from the same trait implementations for `SList` and `List`.

src/translate.rs
majorz: This is done now.

src/translate.rs
majorz: Also done.

src/translate.rs
majorz: This is now done as well.

src/translate.rs
sdroege: Oh, can you also fix those then? Thanks :)

src/translate.rs
sdroege: Indeed. That sucks but what can we do :)

src/translate.rs
```diff
@@ -1003,6 +1003,88 @@ where
     }
 }
 
+pub struct PtrArray(*mut glib_sys::GPtrArray);
+
+impl Drop for PtrArray {
+    fn drop(&mut self) {
+        unsafe {
+            glib_sys::g_ptr_array_free(self.0, true.to_glib());
```
sdroege: `g_ptr_array_unref()` still

majorz: I misunderstood the previous comment, sorry. Not it is using `g_ptr_array_unref()` instead of `g_ptr_array_free()`.

sdroege: Sorry for being not more clear!

sdroege:
Thanks, looks good to me now! @GuillaumeGomez might be nice to get this also into the 0.9.1 bugfix release. Doesn't change any API but adds some, but will be useful for people doing new bindings :)

GuillaumeGomez:
Sure! Just add it into the PR as well once merged.

sdroege:
Ah the macos issue here too. Worst platform ever

PR #560, opened 2020-01-01
Switch from lazy_static to once_cell

sdroege:
@GuillaumeGomez @EPashkin


src/auto/versions.txt
```diff
@@ -1,2 +1,2 @@
-Generated by gir (https://github.com/gtk-rs/gir @ 5836573)
+Generated by gir (https://github.com/gtk-rs/gir @ 83cc9add)
```
EPashkin: Last 'd' unneeded
```
diff --git b/src/auto/versions.txt a/src/auto/versions.txt

index 02762ba..90528f6 100644

--- b/src/auto/versions.txt

+++ a/src/auto/versions.txt

@@ -1,2 +1,2 @@

-Generated by gir (https://github.com/gtk-rs/gir @ 83cc9ad)

+Generated by gir (https://github.com/gtk-rs/gir @ 83cc9add)
```

EPashkin:
Regen check fails and seems need same fix as in https://github.com/gtk-rs/gdk/pull/326

src/auto/versions.txt
sdroege: Hmm where does that additional d come from? I didn't add it there :)

Issue #554, opened 2019-12-16
Improve clone macro error messages


GuillaumeGomez:
In those cases:

```rust
clone!(foo => move || => {});
clone!(@weak foo => default-return false || {}); // missing comma after "false"
```

sdroege:
Are you going to work on that?

GuillaumeGomez:
Yes, unless someone wants to?

sdroege:
Anything else you want to do here?

GuillaumeGomez:
I'm pretty sure there is. Just didn't make the mistake yet. :D

ricvelozo:
Also, the following code:

```rust
action.connect_activate(clone!( @weak application => move |_, _| application.quit() ));
```
Return this error:

```console
error: Variables need to be valid identifiers, e.g. field accesses are not allowed
```

Adding curly brackets lets to compile.

GuillaumeGomez:
Nicely found @ricvelozo ! Fixing it.

PR #563, opened 2020-01-02
Check clone

GuillaumeGomez:
This adds checks for the clone macro compilation errors and also fixes https://github.com/gtk-rs/glib/issues/554#issuecomment-569218290


sdroege:
I think you want https://crates.io/crates/trybuild :)

GuillaumeGomez:
Didn't know this one. But considering we are in a simple case, not sure if it's worth it. If we had proc-macros with more advanced errors, it'd make sense though.

sdroege:
It makes sense because you don't have to add a 150 line Python script for doing what that crate already does :)

GuillaumeGomez:
150 lines seem very little to me but I see your point. What do you think @EPashkin ?

ricvelozo:
@GuillaumeGomez Would it be possible to support closures without curly braces, like common Rust? Specially for simpler cases.

GuillaumeGomez:
We could. But not sure if we want it.

@EPashkin @sdroege: opinion?

sdroege:
Sure, why not? More consistent with how closures work otherwise.

GuillaumeGomez:
Updatd. Tried trybuild, needs to be improved before we can use it here.

sdroege:
> Updatd. Tried trybuild, needs to be improved before we can use it here.

Can you create an issue against trybuild with your problems so it is not forgotten? :)

GuillaumeGomez:
I did better: I created a calendar event so I get spammed every day until I've done it.

sdroege:
ok :) Anything missing here? Do you also want this in the 0.9.1 bugfix release?

src/clone.rs
```diff
@@ -389,9 +410,24 @@ macro_rules! clone {
...
+                let return_value = || $crate::to_return_value!($($return_value)?);
+                $( $crate::to_type_after!($(as $rename)? $(@ $strength)? $($variables).+, return_value);)*
+                $body
+            }
+        }
+    );
+    ($($(@ $strength:ident)? $($variables:ident).+ $(as $rename:ident)?),+ => $(@default-return $return_value:expr,)? move | $($pattern:pat),* | $($body:tt)* ) => (
+        // In case we have:
+        // clone!(@weak foo => move |bla| bla);
+        compile_error!("Closure body needs to be surrounded with curly braces");
```
sdroege: I wonder if rustfmt would remove the curly braces if the closure body is too small. What does it do, can you check? :)

If rustfmt removes them then we need to support that case unfortunately.

GuillaumeGomez: Normally it's supposed to be removed. Forgot to remove this one I guess...

sdroege: I mean a test that succeeds to compile and run :)

GuillaumeGomez: I updated one in the doc. :)

src/clone.rs
```diff
@@ -360,6 +362,15 @@ macro_rules! clone {
             }
         }
     );
+    ($($(@ $strength:ident)? $($variables:ident).+ $(as $rename:ident)?),+ => @default-panic, move || $body:expr ) => (
```
sdroege: Couldn't this just call itself and put `{}` around `$body`? Less code duplication then

src/clone.rs
```diff
@@ -370,6 +381,16 @@ macro_rules! clone {
             }
         }
     );
+    ($($(@ $strength:ident)? $($variables:ident).+ $(as $rename:ident)?),+ => $(@default-return $return_value:expr,)? move || $body:expr ) => (
```
sdroege: Same here

src/clone.rs
```diff
@@ -389,9 +410,19 @@ macro_rules! clone {
             }
         }
     );
+    ($($(@ $strength:ident)? $($variables:ident).+ $(as $rename:ident)?),+ => @default-panic, move | $($pattern:pat),* | $body:expr ) => (
```
sdroege: And here

GuillaumeGomez:
Added more tests for the macro.

src/clone.rs
```diff
@@ -360,16 +356,22 @@ macro_rules! clone {
...
+    ($($(@ $strength:ident)? $($variables:ident).+ $(as $rename:ident)?),+ => @default-panic, move || $body:expr ) => (
+        clone!($($(@ $strength)? $($variables).+ $(as $rename)?),* => @default-panic, move || { $body })
+    );
     ($($(@ $strength:ident)? $($variables:ident).+ $(as $rename:ident)?),+ => $(@default-return $return_value:expr,)? move || $body:block ) => (
         {
             $( $crate::to_type_before!($(@ $strength)? $($variables).+ $(as $rename)?); )*
             move || {
-                let return_value = || $crate::to_return_value!($($return_value)?);
-                $( $crate::to_type_after!($(as $rename)? $(@ $strength)? $($variables).+, return_value );)*
+                let _return_value = || $crate::to_return_value!($($return_value)?);
```
sdroege: Why the underscore?

sdroege:
Looks good to me otherwise

src/clone.rs
GuillaumeGomez: In case you don't use the returned value, it avoids getting a warning about it (discovered it when adding the tests).

sdroege: Ah, makes sense. Merge when the CI is green please, or I'll merge it later once I notice that the CI has passed :)

I'll also do a backport for 0.9.1 then and then we can release, or what do you think?

GuillaumeGomez: Sounds good to me!

GuillaumeGomez:
Fixed typos and forgot to update the tests. :)

GuillaumeGomez:
And now formatting...

sdroege:
Hmm that formatting looks like rustfmt can't figure out our macro correctly. I think you'll want to format this like following, or similar. Why do we have the arrow in the macro there btw instead of just a comma?

```rust
        let _ = clone!(
            @strong v,
            @strong w as x =>
            @default-panic,
            move || println!("v: {}, w: {}", v, x)
        );
```

GuillaumeGomez:
> Why do we have the arrow in the macro there btw instead of just a comma?

To make the reading easier by splitting the arguments and the rest visually.

sdroege:
@GuillaumeGomez You do the 0.9.1 backport and then the release?

GuillaumeGomez:
Sure.

PR #568, opened 2020-01-06
add genum-derive crate

gdesmott:
This crate provides a proc derive macro implementing all the
boilerplate needed to define GEnum properties.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ futures-util = "0.3"
 futures-channel = "0.3"
 glib-sys = { git = "https://github.com/gtk-rs/sys" }
 gobject-sys = { git = "https://github.com/gtk-rs/sys" }
+glib-macros = { path = "glib-macros" }
 
 [dev-dependencies]
 tempfile = "3"
--- a/src/subclass/mod.rs
+++ b/src/subclass/mod.rs
@@ -16,24 +16,57 @@
 //! use glib::subclass;
 //! use glib::subclass::prelude::*;
 //!
-//! use std::cell::RefCell;
+//! use std::cell::{Cell, RefCell};
+//!
+//! #[derive(Debug, Eq, PartialEq, Clone, Copy, GEnum)]
+//! #[repr(u32)]
+//! // type_name: GType name of the GEnum (mandatory)
+//! #[genum(type_name = "SimpleObjectAnimal")]
+//! enum Animal {
+//!   Goat = 0,
+//!   #[genum(name = "The Dog")]
+//!   Dog = 1,
+//!   // name: the name of the GEnumValue (optional, default to the enum name in CamelCase
+//!   // nick: the nick of the GEnumValue (optional, default to the enum name in kebab-case
+//!   #[genum(name = "The Cat", nick = "chat")]
+//!   Cat = 2,
+//! }
+//!
+//! impl Default for Animal {
+//!     fn default() -> Self {
+//!         Animal::Goat
+//!     }
+//! }
 //!
 //! // Static array for defining the properties of the new type.
-//! static PROPERTIES: [subclass::Property; 1] = [subclass::Property("name", |name| {
-//!     glib::ParamSpec::string(
-//!         name,
-//!         "Name",
-//!         "Name of this object",
-//!         None,
-//!         glib::ParamFlags::READWRITE,
-//!     )
-//! })];
+//! static PROPERTIES: [subclass::Property; 2] = [
+//!     subclass::Property("name", |name| {
+//!         glib::ParamSpec::string(
+//!             name,
+//!             "Name",
+//!             "Name of this object",
+//!             None,
+//!             glib::ParamFlags::READWRITE,
+//!         )
+//!     }),
+//!     subclass::Property("animal", |name| {
+//!         glib::ParamSpec::enum_(
+//!             name,
+//!             "Animal",
+//!             "Animal",
+//!             Animal::static_type(),
+//!             Animal::Goat as i32,
+//!             glib::ParamFlags::READWRITE,
+//!         )
+//!     }),
+//! ];
 //!
 //! // This is the struct containing all state carried with
 //! // the new type. Generally this has to make use of
 //! // interior mutability.
 //! pub struct SimpleObject {
 //!     name: RefCell<Option<String>>,
+//!     animal: Cell<Animal>,
 //! }
 //!
 //! // ObjectSubclass is the trait that defines the new type and
@@ -69,6 +102,7 @@
 //!     fn new() -> Self {
 //!         Self {
 //!             name: RefCell::new(None),
+//!             animal: Cell::new(Animal::default()),
 //!         }
 //!     }
 //! }
@@ -89,7 +123,13 @@
 //!                     .get()
 //!                     .expect("type conformity checked by `Object::set_property`");
 //!                 self.name.replace(name);
-//!             }
+//!             },
+//!             subclass::Property("animal", ..) => {
+//!                 let animal = value
+//!                     .get()
+//!                     .expect("type conformity checked by `Object::set_property`");
+//!                 self.animal.replace(animal.unwrap());
+//!             },
 //!             _ => unimplemented!(),
 //!         }
 //!     }
@@ -101,6 +141,7 @@
 //!
 //!         match *prop {
 //!             subclass::Property("name", ..) => Ok(self.name.borrow().to_value()),
+//!             subclass::Property("animal", ..) => Ok(self.animal.get().to_value()),
 //!             _ => unimplemented!(),
 //!         }
 //!     }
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -31,6 +31,7 @@ futures-util = "0.3"
 futures-channel = "0.3"
 glib-sys = { git = "https://github.com/gtk-rs/sys" }
 gobject-sys = { git = "https://github.com/gtk-rs/sys" }
+genum-derive = { path = "genum-derive" }
```
sdroege: I'd name it `glib-macros` :) I'm sure over time we'll add more things in there

genum-derive/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "genum-derive"
+version = "0.1.0"
+authors = ["Guillaume Desmottes <guillaume.desmottes@collabora.com>"]
+edition = "2018"
```
sdroege: License, description, etc? :)

genum-derive/src/lib.rs
```diff
@@ -0,0 +1,295 @@
+// Copyright 2017-2018, The Gtk-rs Project Developers.
```
sdroege:
```suggestion
// Copyright 2019, The Gtk-rs Project Developers.
```

genum-derive/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+use proc_macro_error::{abort_call_site, proc_macro_error};
+use quote::{format_ident, quote, quote_spanned};
+use syn::{
+    parse_macro_input, punctuated::Punctuated, spanned::Spanned, token::Comma, Attribute, Data,
+    DeriveInput, Ident, Lit, Meta, MetaList, NestedMeta, Variant,
+};
+
+#[proc_macro_derive(GEnum, attributes(genum))]
+#[proc_macro_error]
+pub fn genum_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
sdroege: Please put this macro into a separate module instead of directly in `lib.rs`

genum-derive/tests/test.rs
```diff
@@ -0,0 +1,55 @@
+use ::genum_derive::GEnum;
+use glib::prelude::*;
+use glib::translate::{FromGlib, ToGlib};
+
+#[test]
+fn derive_genum() {
+    #[derive(Debug, Eq, PartialEq, Clone, Copy, GEnum)]
+    #[repr(u32)]
```
sdroege: Should the `repr` attribute maybe be added by the derive macro directly?

genum-derive/src/lib.rs
gdesmott: Aren't proc macros supposed to be put directly in `lib.rs`?

genum-derive/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+
+        attrs.into_iter().for_each(|attr|
+            match attr {
+                ItemAttribute::Name(n) => value_name = n,
+                ItemAttribute::Nick(n) => value_nick = n,
+            }
+        );
+
+        let value_name = format!("{}\0", value_name);
+        let value_nick = format!("{}\0", value_nick);
```
sdroege: above you write `\0` but in the comment you write `\u{0}` :)

genum-derive/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+//     },
+fn gen_genum_values(
+    enum_name: &Ident,
+    enum_variants: &Punctuated<Variant, Comma>,
+) -> (TokenStream, usize) {
+    // start at one as GEnumValue array is null-terminated
+    let mut n = 1;
+    let recurse = enum_variants.iter().map(|v| {
+        let name = &v.ident;
+        let mut value_name = name.to_string();
```
sdroege: Isn't this usually upper-snake-case?

genum-derive/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+            unsafe fn from_value(value: &glib::Value) -> Self {
+                use glib::translate::ToGlibPtr;
+
+                glib::translate::from_glib(gobject_sys::g_value_get_enum(value.to_glib_none().0))
+            }
+        }
+
+        impl glib::value::SetValue for #name {
+            unsafe fn set_value(value: &mut glib::Value, this: &Self) {
+                use glib::translate::{ToGlib, ToGlibPtrMut};
```
sdroege: `use` in macro always makes me a bit nervous. It could cause errors if the same type is already imported outside the macro, for example. Or in case of traits if the used type also implements a function with the given name then not the one of the trait would be used.

Generally it's nicer to directly call the trait methods `glib::translate::ToGlib::to_glib(this)`, `glib::translate::ToGlibPtrMut(value)`

genum-derive/src/lib.rs
gdesmott: Aren't we in 2020 now? ;)

genum-derive/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        impl StaticType for #name {
+            fn static_type() -> glib::Type {
+                #get_type()
+            }
+        }
+
+        fn #get_type() -> glib::Type {
+            use std::sync::Once;
```
sdroege: Also here

src/subclass/mod.rs
```diff
@@ -12,28 +12,60 @@
 //! ```rust
 //! #[macro_use]
 //! extern crate glib;
+//! #[macro_use]
+//! extern crate genum_derive;
```
sdroege: Maybe we should re-export that crate from the `glib` crate so it's easier available?

src/subclass/mod.rs
```diff
@@ -12,28 +12,60 @@
...
+//! use std::cell::{Cell, RefCell};
+//!
+//! #[derive(Debug, Eq, PartialEq, Clone, Copy, GEnum)]
+//! #[repr(u32)]
+//! #[genum(type_name = "SimpleObjectAnimal")]
+//! enum Animal {
+//!   Goat = 0,
+//!   #[genum(name = "The Dog")]
+//!   Dog = 1,
+//!   #[genum(name = "The Cat", nick = "chat")]
```
sdroege: Maybe should get some docs somewhere that also mention what the default values for name/nick would be and that type_name is mandatory

src/subclass/mod.rs
```diff
@@ -12,28 +12,60 @@
...
+//!   Cat = 2,
+//! }
+//!
+//! impl Default for Animal {
+//!     fn default() -> Self {
+//!         Animal::Goat
```
sdroege: This is btw what `#[derive(Default)]` would do already :)

genum-derive/src/lib.rs
sdroege: Not sure, but even if so you could put at least the implementation (what you call in here) into a separate module :)

I'm mostly worried about having `lib.rs` becoming huge once we add more macros

genum-derive/src/lib.rs
sdroege: Nah, it's just 2019 part 2 🤦‍♂️

src/subclass/mod.rs
gdesmott: Would you use it in such example? Wondering if it isn't more exemplary to keep it explicit.

src/subclass/mod.rs
gdesmott: What's the proper way to re-export such macro?
I tried this:
```rust
#[macro_use]
extern crate glib_macros;
pub use glib_macros::genum_derive;
```

but got:
```
error[E0432]: unresolved import `glib_macros::genum_derive`
  --> src/lib.rs:98:9
   |
98 | pub use glib_macros::genum_derive;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^ no `genum_derive` in the root
```

sdroege: `pub use glib_macros::GEnum` i think

genum-derive/tests/test.rs
gdesmott: Do we really need it actually? It seems to work fine without it.

sdroege: This way you make sure that it becomes a compiler error if the enum actually contains items that can't be represented as `u32`

gdesmott: I don't know how to tweak the macro to generate this before the `enum {}` block and the rest after it. Any idea?

gdesmott:
@sdroege : new version fixing all your comments but the last 2 still open.

src/subclass/mod.rs
sdroege: I would derive here. It's an example of GLib thingies, not general Rust :) But I don't mind either way

genum-derive/tests/test.rs
sdroege: I don't, so maybe let's keep that for a later time. Your `FromGlib` / `ToGlib` impls would fail to compile if the enum is of an incompatible shape so it should be fine.

src/subclass/mod.rs
gdesmott: Looks like it's not supported with enum:

```
error[E0665]: `Default` cannot be derived for enums, only structs
  --> src/subclass/mod.rs:21:45
   |
10 | #[derive(Debug, Eq, PartialEq, Clone, Copy, Default, GEnum)]
   |                                             ^^^^^^^
```

sdroege: Curious, which language was it that allowed this and had this behaviour then? Oh well :)

EPashkin:
@gdesmott Thanks.
👍

sdroege:
Let's get this in then once the CI is happy :)

PR #578, opened 2020-01-24
Add glib::ThreadPool bindings

sdroege:
Including a futures-enabled push function that allows to await on the
result of the closure passed to the thread pool:

    let pool = glib::ThreadPool::new_shared(None)?;
    let res = pool.push_future(|| calculate_things())?.await;
    do_something_with_things(res);


EPashkin:
@sdroege Thanks

PR #579, opened 2020-01-24
Implements FromGlibPtrFull<*const T> for boxed types.

luukvanderduim:
The aim is to solve issue: 865 of the gir repository [865](https://github.com/gtk-rs/gir/issues/865)

AnyBox::Native(Box<T>) looks like the suitable variant for a fully owned *const T
Is checking non_null sufficient here before converting the raw pointer to a box?
I'm assuming here MM::copy() assures other type requirements are met.
Is this correct? 
While this solution makes my code compile and most of this impl is copied and adapted from other impls in the same file, there may well be things I have missed or misunderstood. Therefore a close look at what I've done is much appreciated.

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -471,7 +471,7 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> FromGlibPtrFull<*const T> for Boxed<
     unsafe fn from_glib_full(ptr: *const T) -> Self {
         assert!(!ptr.is_null());
         Boxed {
-            inner: AnyBox::ForeignOwned(ptr::NonNull::new_unchecked(ptr as *mut _)),
+            inner: AnyBox::ForeignOwned(ptr::NonNull::new_unchecked(ptr as *mut T)),
             _dummy: PhantomData,
         }
     }
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -473,7 +473,7 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> FromGlibPtrFull<*const T> for Boxed<
         let ptr = MM::copy(ptr);
         Boxed {
             inner: AnyBox::Native(Box::from_raw(ptr)),
-            _dummy: PhantomData
+            _dummy: PhantomData,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/boxed.rs
```diff
@@ -457,6 +466,18 @@ impl<T: 'static, MM: BoxedMemoryManager<T>> FromGlibPtrFull<*mut T> for Boxed<T,
     }
 }
 
+impl<T: 'static, MM: BoxedMemoryManager<T>> FromGlibPtrFull<*const T> for Boxed<T, MM> {
+    #[inline]
+    unsafe fn from_glib_full(ptr: *const T) -> Self {
+        assert!(!ptr.is_null());
+        let ptr = MM::copy(ptr);
+        Boxed {
+            inner: AnyBox::Native(Box::from_raw(ptr)),
```
sdroege: This is only allowed if the pointer was originally a `Box` and allocated from Rust. That's not the case here :)

Also simply copying the memory behind the pointer is not enough here. You'd leak memory.

What you want here is to call the `*mut _` version of `from_glib_full()` by simply casting the pointer.

luukvanderduim: Thanks! Will try to get that in place.

luukvanderduim: Simplest proposal:
`unsafe fn from_glib_full(ptr: *const T) -> Self {
        assert!(!ptr.is_null());
        let ptr = ptr as *mut _;
        Boxed {
            inner: AnyBox::ForeignOwned(ptr::NonNull::new_unchecked(ptr)),
            _dummy: PhantomData,
        }
    }`
This is essentially the same as the FromGlibPtrFull on *mut but with cast to *mut. Would this suffice?

sdroege: You can do it even easier, but basically yes. You could do `from_glib_full(ptr as *mut T)`

luukvanderduim: Could we even generalize over *mut/*const pointers by using a suitable pointer trait in the signature? I quickly tried Ptr but run into trouble.

sdroege: I don't think so

luukvanderduim:
Travis'CI has an osx-nightly run that fails. I've seen this yesterday  as well. 
```
Caused by:
  Unable to update https://github.com/gtk-rs/gio 
```
Harmless Travis quirk?

EPashkin:
Restarted job

PR #580, opened 2020-01-30
Minor version update [ci skip]

EPashkin:
cc @GuillaumeGomez, @sdroege


sdroege:
What's the reason for this PR together with https://github.com/gtk-rs/glib/pull/581 ? This one is part of the other one or not? :)

EPashkin:
@sdroege This one is on master, while other is on "crate" based "0.9",
we need both to make version synchronized

sdroege:
Oh, I see :)

PR #585, opened 2020-02-07
WIP:re-generate with GLib 2.63.5

bilelmoussaoui:
this is mostly to make the new g_get_os_info available https://developer.gnome.org/glib/unstable/glib-Miscellaneous-Utility-Functions.html#g-get-os-info. This requires a newer glib version, the stable release isn't out yet but it would allow me to handle this https://gitlab.gnome.org/GNOME/gnome-tour/issues/10 properly.

The gir-files repo needs to be updated separately


src/auto/constants.rs
```diff
@@ -17,12 +17,6 @@ pub static CSET_a_2_z: once_cell::sync::Lazy<&'static str> =
     once_cell::sync::Lazy::new(|| unsafe {
         CStr::from_ptr(glib_sys::G_CSET_a_2_z).to_str().unwrap()
     });
-pub static KEY_FILE_DESKTOP_ACTION_GROUP_PREFIX: once_cell::sync::Lazy<&'static str> =
```
EPashkin: Strange, why it was removed?

EPashkin:
@bilelmoussaoui gir-files repo need to be updated before this PR

src/auto/constants.rs
bilelmoussaoui: I wondered the same. Most of them seems to not be that useful or not used anymore

EPashkin:
also https://github.com/gtk-rs/sys need regen before this too

EPashkin:
I canceling most travis jobs

bilelmoussaoui:
@EPashkin I was looking for where the glib-sys parts are. I will wait until https://github.com/gtk-rs/gir-files/pull/49 is merged and re-update this & -sys. Thanks for the fast review :)

sdroege:
There's also the problem that we can't do a stable release with this until the stable GLib release is out. Any new API might still change until then.

Not necessarily a blocker for merging this  but something to keep in mind.
-- 

bilelmoussaoui:
I'm using GTK4 so I don't need a stable release for now.

GuillaumeGomez:
Well, I'd prefer if we only merged released API. I don't want us to maintain API that aren't stable.

bilelmoussaoui:
> Well, I'd prefer if we only merged released API. I don't want us to maintain API that aren't stable.

Makes sense. I will keep this open until there's a stable glib release :)

src/auto/constants.rs
sdroege: It's a SuSE/Debian/Ubuntu-specific patch that added these in the first place :man_facepalming: 
https://sources.debian.org/src/glib2.0/2.62.4-2/debian/patches/01_gettext-desktopfiles.patch/

sdroege:
@bilelmoussaoui 2.64 was released now :)

Issue #586, opened 2020-02-08
Confusing FromGlibPtrNone docs


dcz-purism:
When trying to create a custom GVariant using `ToVariant::to_variant`, I tried to use `from_glib_ptr_full`, only to see glib criticals due to double-free. As I investigated, I came across the implementation for basic types in https://github.com/gtk-rs/glib/blob/master/src/variant.rs#L260

This implementation, confusingly, returns an owned `Variant`, using `from_glib_ptr_none`.

> FromGlibPtrNone: Translate from a pointer type without taking ownership, transfer: none.

It seems this function is only useful for references here. I looked for something explaining this discrepancy, and I found https://github.com/gtk-rs/glib/blob/master/src/object.rs#L551

> impl FromGlibPtrNone {
> unsafe fn from_glib_none {
> // Attention: This takes ownership of floating references!

This makes it clear that the `Variant` instance returned is actually owned, but on the other hand, it contradicts the documentation for `FromGlibPtrNone`.

I have an impression that `FromGlibPtrNone` actually does more/something else than its documentation says, but I'm not quite sure what it is. I believe the documentation for the translate traits calls for some more details, so that the use cases are clear. It might be useful to explain the "floating references", or make it clear how they are being used in the bindings.

EPashkin:
As Variant use Shared, you better look to it, not to Object.
In  https://github.com/gtk-rs/glib/blob/master/src/shared.rs#L393
[g_variang_ref_sing](https://developer.gnome.org/glib/stable/glib-GVariant.html#g-variant-ref-sink)  called, and it claim floating reference.

EPashkin:
Docs of FromGlibPtrNone show generic case, but floating references add complexity.
And shared types like Variant add its own complexity too.

EPashkin:
Also https://developer.gnome.org/glib/stable/glib-GVariant.html#g-variant-new-boolean etc. says `[transfer none]` too.

dcz-purism:
Thanks for answering.

Does `[transfer none]` relate to `FromGlibPtrNone`? They seem to be completely different:

> don't free data after the code is done

this sounds a lot like "obtain and take ownership", which is what `FromGlibPtrFull` says it does.

If they are the same, then the docs for `FromGlibPtrNone` are wrong, because ownership is actually transferred.

EPashkin:
"take ownership" meant "you must free data after you finished to use or you get memory leak"

EPashkin:
In case referenced types:
[...]

PR #587, opened 2020-02-08
translate: Clarify the purpose and the effect of translation traits

dcz-purism:
Attempts to solve https://github.com/gtk-rs/glib/issues/586

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -21,13 +21,31 @@
 //! ```
 //!
 //! `ToGlibPtr`, `FromGlibPtrNone`, `FromGlibPtrFull` and `FromGlibPtrBorrow` work on `gpointer`s
-//! and support different modes of ownership transfer.
+//! and ensure correct ownership of values
+//! according to [Glib ownership transfer rules](https://gi.readthedocs.io/en/latest/annotations/giannotations.html).
+//!
+//! `FromGlibPtrNone` and `FromGlibPtrFull`
+//! must be called on values obtained from C,
+//! according to their `transfer` annotations.
+//! They acquire non-gobject types,
+//! as well as turning floating references to strong ones,
+//! which are the only ones properly handled by the Rust bindings.
+//!
+//! For more information about floating references, please refer to the "Floating references" section
+//! of [the gobject reference](https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html).
 //!
 //! ```ignore
 //!     fn get_title(&self) -> Option<String> {
 //!         unsafe {
 //!             let title = gtk_sys::gtk_window_get_title(self.pointer);
 //!             from_glib_none(title)
+//!         }
+//!     }
+//!     fn create_bool(value: gboolean) -> Variant {
+//!         unsafe {
+//!             let variant = glib_sys::g_variant_new_boolean(value);
+//!             // g_variant_new_boolean has `transfer none`
+//!             from_glib_none(variant)
 //!         }
 //!     }
 //! ```
@@ -1162,17 +1180,60 @@ impl FromGlib<i32> for Option<u64> {
     }
 }
 
-/// Translate from a pointer type without taking ownership, transfer: none.
+/// Translate from a pointer type which is annotated with `transfer none`.
+/// The resulting value is referenced at least once, by the bindings.
+///
+/// This is suitable for floating references, which become strong references.
+/// It is also suitable for acquiring non-gobject values, like `gchar*`.
+///
+/// The implementation of this trait should acquire a reference to the value
+/// in a way appropriate to the type,
+/// e.g. by increasing the reference count or copying.
+/// Values obtained using this trait must be properly released on `drop()`
+/// by the implementing type.
+///
+/// For more information, refer to module level documentation.
 pub trait FromGlibPtrNone<P: Ptr>: Sized {
     unsafe fn from_glib_none(ptr: P) -> Self;
 }
 
-/// Translate from a pointer type taking ownership, transfer: full.
+/// Translate from a pointer type which is annotated with `transfer full`.
+/// This transfers the ownership of the value to the Rust side.
+///
+/// Because ownership can only be transferred if something is already referenced,
+/// this is unsuitable for floating references.
+///
+/// The implementation of this trait should not alter the reference count
+/// or make copies of the underlying value.
+/// Values obtained using this trait must be properly released on `drop()`
+/// by the implementing type.
+///
+/// For more information, refer to module level documentation.
 pub trait FromGlibPtrFull<P: Ptr>: Sized {
     unsafe fn from_glib_full(ptr: P) -> Self;
 }
 
-/// Translate from a pointer type by borrowing. Don't increase the refcount
+/// Translate from a pointer type by borrowing, without affecting the refcount.
+///
+/// The purpose of this trait is to access values inside callbacks
+/// without changing their reference status.
+/// The obtained borrow must not be accessed outside of the scope of the callback,
+/// and called procedures must not store any references to the underlying data.
+/// Safe Rust code must never obtain a mutable Rust reference.
+///
+/// The implementation of this trait as well as the returned type
+/// must satisfy the same constraints together.
+/// They must not take ownership of the underlying value, copy it,
+/// and should not change its rerefence count.
+/// If it does, it must properly release obtained references.
+///
+/// The returned value, when dropped,
+/// must leave the underlying value in the same state
+/// as before from_glib_borrow was called:
+/// - it must not be dropped,
+/// - it must be the same type of reference, e.g. still floating.
+///
+/// For more information, refer to module level documentation.
 pub trait FromGlibPtrBorrow<P: Ptr>: Sized {
     unsafe fn from_glib_borrow(_ptr: P) -> Self {
         unimplemented!();
@@ -1180,6 +1241,8 @@ pub trait FromGlibPtrBorrow<P: Ptr>: Sized {
 }
 
 /// Translate from a pointer type, transfer: none.
+///
+/// See [`FromGlibPtrNone`](trait.FromGlibPtrNone.html).
 #[inline]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe fn from_glib_none<P: Ptr, T: FromGlibPtrNone<P>>(ptr: P) -> T {
@@ -1187,6 +1250,8 @@ pub unsafe fn from_glib_none<P: Ptr, T: FromGlibPtrNone<P>>(ptr: P) -> T {
 }
 
 /// Translate from a pointer type, transfer: full (assume ownership).
+///
+/// See [`FromGlibPtrFull`](trait.FromGlibPtrFull.html).
 #[inline]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe fn from_glib_full<P: Ptr, T: FromGlibPtrFull<P>>(ptr: P) -> T {
@@ -1194,6 +1259,8 @@ pub unsafe fn from_glib_full<P: Ptr, T: FromGlibPtrFull<P>>(ptr: P) -> T {
 }
 
 /// Translate from a pointer type, borrowing the pointer.
+///
+/// See [`FromGlibPtrBorrow`](trait.FromGlibPtrBorrow.html).
 #[inline]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe fn from_glib_borrow<P: Ptr, T: FromGlibPtrBorrow<P>>(ptr: P) -> T {


src/translate.rs
```diff
@@ -21,7 +21,14 @@
 //! ```
 //!
 //! `ToGlibPtr`, `FromGlibPtrNone`, `FromGlibPtrFull` and `FromGlibPtrBorrow` work on `gpointer`s
-//! and support different modes of ownership transfer.
+//! and ensure correct ownership of references on the Rust side
```
EPashkin: Maybe s/references/data/

src/translate.rs
```diff
@@ -30,6 +37,13 @@
 //!             from_glib_none(title)
 //!         }
 //!     }
+//!     fn create_bool(value: gboolean) -> Variant {
+//!         unsafe {
+//!             let variant = glib_sys::g_variant_new_boolean(value);
+//!             // g_variant_new_boolean has `transfer none`
```
EPashkin: This comment can be added to previous example: // gtk_window_get_title has `transfer none` for return type

src/translate.rs
```diff
@@ -1162,38 +1176,64 @@ impl FromGlib<i32> for Option<u64> {
     }
 }
 
-/// Translate from a pointer type without taking ownership, transfer: none.
+/// Translate from a pointer type which is annotated with `transfer: none`.
+/// The resulting value is referenced at least once, by the bindings.
+/// The value must be unreferenced on `drop()`.
```
EPashkin: s/must/will/ ?

src/translate.rs
```diff
@@ -1162,38 +1176,64 @@ impl FromGlib<i32> for Option<u64> {
...
+///
+/// This is suitable for floating references, which become full and owned.
```
EPashkin: "As soon as g_object_ref_sink() is called on the object, the floating reference is converted to a strong reference, and the calling code assumes ownership of the object. "
"become strong referenced" ?

src/translate.rs
dcz-purism: I used "value" later. Is "value" or "data" better?

src/translate.rs
dcz-purism: I wanted to do that, but the docs have no annotation other than `[nullable]`: https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-get-title

and it's a string and not a GObject, so I thought it's good to have a string example too.

src/translate.rs
EPashkin: IMHO "value" is better.

src/translate.rs
dcz-purism: The trait can be defined on any type, so it can't guarantee that things will get dropped correctly. In other words, the trait cannot say "this trait will do this for you". The implementor is responsible for and must drop things, so I wrote "must" for "to implement this trait correctly, you must do this".

src/translate.rs
EPashkin: GVariant also not GObject.
Ok, I vote up for current version.

src/translate.rs
dcz-purism: Thanks.

src/translate.rs
EPashkin: You right

dcz-purism: I tried to make it a little more where the responsibility for `drop()` lies by spelling it out.

src/translate.rs
```diff
@@ -1162,38 +1176,64 @@ impl FromGlib<i32> for Option<u64> {
...
+/// this is unsuitable for floating references.
+///
+/// For more information, refer to module level documentation.
 pub trait FromGlibPtrFull<P: Ptr>: Sized {
     unsafe fn from_glib_full(ptr: P) -> Self;
 }
 
-/// Translate from a pointer type by borrowing. Don't increase the refcount
+/// Translate from a pointer type by borrowing, without affecting the refcount.
+/// Types implementing this trait must *not* unreference the obtained value on `drop()`.
```
EPashkin: One type can implement all 3 traits, so maybe "value, obtained through this trait," or so

EPashkin:
Note: these traits can be used not only for "ref/unref" (Shared, Object in glib-rs term> types but for "copy/free" (Boxed) too

src/translate.rs
```diff
@@ -1162,38 +1176,67 @@ impl FromGlib<i32> for Option<u64> {
     }
 }
 
-/// Translate from a pointer type without taking ownership, transfer: none.
+/// Translate from a pointer type which is annotated with `transfer: none`.
+/// The resulting value is referenced at least once, by the bindings.
```
EPashkin: Maybe we need separate description from 2 sides: from users (this phrase), and for implementers (next phrase) for all aspects.
Analog for this phrase for impls: "For referenced types implementation need to use `ref_sink` instead `ref` to correctly handle floating reference,
for copy types implementation need to make `copy`"

dcz-purism:
I don't know much about Boxed, but looking at how I wrote it, it seems it still applies to full copies. Should I still add a note?

src/translate.rs
EPashkin: Or not mention "implementers" side at all

EPashkin: "`ref_sink` (if it present for type) instead .." or so

EPashkin:
At minimum all "unreferenced on `drop()`" IMHO need changed to "unreferenced or freed on `drop()`"

dcz-purism:
What about "properly released" instead of "freed" and "unreferenced"?

EPashkin:
> What about "properly released" instead of "freed" and "unreferenced"?

👍

src/translate.rs
```diff
@@ -21,7 +21,14 @@
 //! ```
 //!
 //! `ToGlibPtr`, `FromGlibPtrNone`, `FromGlibPtrFull` and `FromGlibPtrBorrow` work on `gpointer`s
-//! and support different modes of ownership transfer.
+//! and ensure correct ownership of values on the Rust side
```
sdroege: Maybe "ensure correct ownership of values when passing them between Rust and C or the other way around"? It's not just about the Rust side but also the C side

src/translate.rs
```diff
@@ -21,7 +21,14 @@
...
+//! according to [Glib ownership transfer rules](https://gi.readthedocs.io/en/latest/annotations/giannotations.html).
+//!
+//! `FromGlibPtrNone` and `FromGlibPtrFull`
+//! must be called on values obtained from C,
+//! according to their `transfer` annotations.
+//! In particular, they turn floating references to strong ones,
```
sdroege: I'd move the mention of floating references into a separate paragraph with some more explanation what that is (maybe with link to the GObject docs) and why/when it matters. Also only `from_glib_none()` does this (on references that are still floating).

I wouldn't put floating references right at the top as for most users of the API this is just going to confuse and not actually relevant.

src/translate.rs
```diff
@@ -21,7 +21,14 @@
 //! ```
 //!
 //! `ToGlibPtr`, `FromGlibPtrNone`, `FromGlibPtrFull` and `FromGlibPtrBorrow` work on `gpointer`s
-//! and support different modes of ownership transfer.
+//! and ensure correct ownership of values on the Rust side
+//! according to [Glib ownership transfer rules](https://gi.readthedocs.io/en/latest/annotations/giannotations.html).
+//!
+//! `FromGlibPtrNone` and `FromGlibPtrFull`
```
sdroege: There's also `from_glib_borrow()` which might be worth mentioning somewhere else here :) Especially with that one it's important that it should only be used from C callbacks, and that the return value must never be passed outside the scope of the callback, and that only immutable references to it must be passed to safe Rust code.

src/translate.rs
sdroege: There are some defaulting rules for cases where there are no annotations, that's why `gtk_window_get_title()` has no explicit annotations. It's somewhere on the GObject-Introspection wiki.

src/translate.rs
```diff
@@ -1162,38 +1176,67 @@ impl FromGlib<i32> for Option<u64> {
     }
 }
 
-/// Translate from a pointer type without taking ownership, transfer: none.
+/// Translate from a pointer type which is annotated with `transfer: none`.
+/// The resulting value is referenced at least once, by the bindings.
+/// Values obtained using this trait must be unreferenced on `drop()`
+/// by the implementing type.
+///
+/// This is suitable for floating references, which become strong references.
```
sdroege: Maybe with a link to some wider explanation about floating references, see above

src/translate.rs
```diff
@@ -1162,38 +1176,67 @@ impl FromGlib<i32> for Option<u64> {
...
 }
 
-/// Translate from a pointer type by borrowing. Don't increase the refcount
+/// Translate from a pointer type by borrowing, without affecting the refcount.
+/// Values obtained using this trait must *not* be unreferenced on `drop()`
+/// by the implementing type.
+///
+/// It is not recommended to use this on floating references,
+/// which may get transferred to a callee and dropped,
+/// leaving the pointer stale.
```
sdroege: Actually because of one of the many special cases it is required to use this for floating references instead of `from_glib_none()`. Generally this one is only used in callbacks from C (signal handlers, virtual methods, proper callbacks) for getting access on the Rust side to a value that we only "borrow" during the scope of the callback. (see also what I wrote above about some rules for safely using it)

In case of floating references this does not sink them or anything but just keeps them as is. This is especially important for virtual method trampolines: otherwise when e.g. implementing a `GtkWidget` in Rust and overriding the `GObject::constructed` virtual method what would happen is that we would pass it to `from_glib_none()` on the Rust side. That would sink and take ownership of that floating reference then, and at the end of the virtual method we would drop it. Which then means that the caller (who just created a new instance!) would get an already freed pointer back from the constructor.

Now this is all detailed background information and should probably not go verbatim into the docs here :)

src/translate.rs
```diff
@@ -1162,38 +1176,67 @@ impl FromGlib<i32> for Option<u64> {
...
+/// See `FromGlibPtrFull`.
 #[inline]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe fn from_glib_full<P: Ptr, T: FromGlibPtrFull<P>>(ptr: P) -> T {
     FromGlibPtrFull::from_glib_full(ptr)
 }
 
 /// Translate from a pointer type, borrowing the pointer.
+///
+/// See `FromGlibPtrBorrow`.
```
sdroege: Make these things actual links in the docs please :)

src/translate.rs
dcz-purism: > Also only from_glib_none() does this (on references that are still floating).

I wrote that because I understood that the contract on the Rust side ensured by the traits together is "no floating references on the Rust side". This section is less about what each function does, but more about what guarantees the user should expect and uphold.

I would also keep a mention of floating references here at the module level if possible, as this is the reason I dug myself into this hole :)

src/translate.rs
dcz-purism: > only immutable references to it must be passed to safe Rust code

Where does that come from? To cover all cases where the refcount changes?

src/translate.rs
sdroege: Maybe a link to a section elsewhere that explains in more detail what that actually means?

> I wrote that because I understood that the contract on the Rust side ensured by the traits together is "no floating references on the Rust side".

Except for callbacks (see below), where we must not touch them. Whenever we get a floating reference in bindings that we can potentially take ownership of, we do so. To side-step all the problems that otherwise result from floating references (e.g. did `gtk_container_add()` steal my reference or not?)

src/translate.rs
sdroege: It would otherwise be unsafe. We only own the reference during the scope of the callback so if Rust code takes the object and stores it away somewhere then it will have a dangling pointer once the callback has returned. For that reason we must only pass a reference away, lend it, to Rust code. The Rust code does not own it but only borrows it.

dcz-purism:
I think I covered all:

- a link to what floating references are (no special section though)
- the purpose of this is to deal with ownership, not just Rust ownership
- implementor's notes in each trait hint at acquiring refs
- realized that `gchar*` and such were not covered, added notes

src/translate.rs
```diff
@@ -1162,38 +1180,77 @@ impl FromGlib<i32> for Option<u64> {
...
     unsafe fn from_glib_full(ptr: P) -> Self;
 }
 
-/// Translate from a pointer type by borrowing. Don't increase the refcount
+/// Translate from a pointer type by borrowing, without affecting the refcount.
+///
+/// The implementation of this trait must not acquire any references to the value.
+/// Values obtained using this trait must *not* be unreferenced on `drop()` of the returned type.
+///
+/// The purpose of this trait is to access values inside callbacks
```
EPashkin: IMHO better swap this paragraph and previous here and in other traits:
firstly we talk about usage (this parts is separated now), and then talk about implementation.

sdroege:
> realized that `gchar*` and such were not covered, added notes

What do you mean with them not being covered? They are covered so that string/path conversion works correctly.

dcz-purism:
They were not covered as in, the docs didn't mention them outside of the example. So I listed them explicitly.

EPashkin:
Actually this descriptions always use term "reference" while all trait works with refcounted (ref/unref) and "normal" (copy/drop) types int same way for getting its "value" from "pointer".
Not sure how to rephrase first line, but for me old text with ownership in term of
https://developer.gnome.org/programming-guidelines/stable/memory-management.html.en was better.

dcz-purism:
> this descriptions always use term "reference" while all trait works with refcounted (ref/unref) and "normal" (copy/drop) types int same way for getting its "value" from "pointer".

Yes, "reference" is the most generic term because the actual values are behind references in both cases, i.e. the actual value is not directly accessible but only via the pointer.

I'm not sure how this relates to the link you posted.

src/translate.rs
dcz-purism: You're right, I changed the order.

src/translate.rs
```diff
@@ -1162,38 +1180,77 @@ impl FromGlib<i32> for Option<u64> {
...
-/// Translate from a pointer type without taking ownership, transfer: none.
+/// Translate from a pointer type which is annotated with `transfer none`.
+/// The resulting value is referenced at least once, by the bindings.
+///
+/// This is suitable for floating references, which become strong references.
+/// It is also suitable for acquiring non-gobject values, like `gchar*`.
+///
+/// The implementation of this trait should acquire a reference to the value
+/// in a way appropriate to the type, e.g. by referencing or copying.
+/// Values obtained using this trait must be properly unreferenced on `drop()`
```
EPashkin: "unreferenced" here, "released" in "full",  and again "unreferenced" in "borrow"

EPashkin:
@dcz-purism You can ignore my previous nagging, I just like "transfer ownership" more.
Thanks.

src/translate.rs
dcz-purism: Thank you, fixed.

dcz-purism:
The travis build errors out on some network error.

GuillaumeGomez:
Restarted the job.

sdroege:
Is anything missing here?

dcz-purism:
I don't have anything to add.

src/translate.rs
```diff
@@ -1162,38 +1180,77 @@ impl FromGlib<i32> for Option<u64> {
...
     unsafe fn from_glib_full(ptr: P) -> Self;
 }
 
-/// Translate from a pointer type by borrowing. Don't increase the refcount
+/// Translate from a pointer type by borrowing, without affecting the refcount.
+///
+/// The purpose of this trait is to access values inside callbacks
+/// without changing their reference status.
+/// The obtained borrow must not be accessed outside of the scope of the callback.
+/// Care must be taken not to change the reference count of the borrowed value,
```
sdroege: Why? It's perfectly fine (and often required) to call `clone()` on the value to then store it away somewhere. The only thing that must not be done is to take ownership of the value returned by `from_glib_borrow()`

src/translate.rs
```diff
@@ -1162,38 +1180,77 @@ impl FromGlib<i32> for Option<u64> {
...
+/// and not to let any called procedure to store a reference.
+/// In particular, safe Rust code should never obtain a mutable borrow.
```
sdroege: Rust code must never have access to the value itself, only to immutable/shared references to it

src/translate.rs
```diff
@@ -1162,38 +1180,77 @@ impl FromGlib<i32> for Option<u64> {
...
+///
+/// The implementation of this trait must not acquire any references to the value.
```
sdroege:
```suggestion
/// The implementation of this trait must not acquire any references to the value or copy it.
```

src/translate.rs
```diff
@@ -1162,38 +1180,77 @@ impl FromGlib<i32> for Option<u64> {
...
 
-/// Translate from a pointer type without taking ownership, transfer: none.
+/// Translate from a pointer type which is annotated with `transfer none`.
+/// The resulting value is referenced at least once, by the bindings.
+///
+/// This is suitable for floating references, which become strong references.
+/// It is also suitable for acquiring non-gobject values, like `gchar*`.
+///
+/// The implementation of this trait should acquire a reference to the value
+/// in a way appropriate to the type, e.g. by referencing or copying.
```
sdroege: `by referencing` -> `by increasing its reference count` here and also in various other place. While the function is called `ref()`, the correct verb is `increase the reference count`. `referencing` would be ambiguous (e.g. the `&` operator is also referencing something).

src/translate.rs
```diff
@@ -1162,38 +1180,77 @@ impl FromGlib<i32> for Option<u64> {
...
 }
 
-/// Translate from a pointer type taking ownership, transfer: full.
+/// Translate from a pointer type which is annotated with `transfer full`.
+/// This transfers the ownership of the value to the Rust side.
+///
+/// Because ownership can only be transferred if something is already referenced,
+/// this is unsuitable for floating references.
+///
+/// The implementation of this trait should not acquire any references to the value.
```
sdroege: `acquire references` -> `increase reference count`, and maybe also mention that it should not copy

sdroege:
> I don't have anything to add.

Ok, thanks :) I went through it another time. Please just write a comment here if the previous review comments are all addressed, I missed that last time and that's why I didn't go over it again earlier.

src/translate.rs
dcz-purism: Doesn't `clone()`, as you mentioned above, acquire a reference?

sdroege: It increases the reference count, yes

src/translate.rs
dcz-purism: What is meant by "copy" here? Is it about the `Copy` trait or something else?

sdroege: Copy as in creating a new version of the value. Allocating new memory and creating the same thing again. E.g. what the `Clone` trait does on `String`, but not what it does on `Arc`

dcz-purism: I used "acquire reference" throughout the text as a compromise between "increase reference count" and "copy", because the same trait applies to either kind of data if I understood this right.

src/translate.rs
dcz-purism: Then I'm seeing a contradiction:

> It's perfectly fine (and often required) to call clone() on the value

but

> /// The implementation of this trait must not acquire any references to the value

How do those two statements work together?

src/translate.rs
sdroege: Yes that's correct. `acquire reference` is not very descriptive though, I would simply mention both possibilities there.

src/translate.rs
sdroege: The first statement is for safe Rust code that gets a reference (`&T`) to the value passed. It can safely `clone()` if it wants.

The second is about the implementation of the trait. That should not copy or increase the reference count, but just "borrow" the value.

dcz-purism: Okay, the first sentence is about what can be done on the resulting value, and the second is about the transformation done by the trait. I'm not sure why the two cases are separate though.

sdroege: One is the implementation of the trait and how it should behave. The other is about what is allowed and not allowed with the resulting value (you must not pass it by value or mutable reference to safe code).

dcz-purism: From the position of the contract of this trait, I think deep-copying the value is always allowed, because that doesn't affect what can be done with the produced reference-thing. Would you agree?

sdroege: Yes, that's the case for all 3 traits. But in case of full/borrow it should not do that unless there's no other way. It would be unexpected to have the overhead of a copy.

src/translate.rs
dcz-purism: What I (think I) understand:
- no procedure may obtain the value via move, because the lifetime of the underlying pointer is constrained to the scope of the call (but not enforced by the type signature)
- no procedure may be allowed to store a Rust `&` reference for the same reasons, with the same solution

What I don't understand is what's wrong with mutable references being passed around. They can't be stored anyway due to the above, and mutating callback params in C code is the usual thing. It's also not about Glib references, because clone() creates them and is allowed.

sdroege: If you have a mutable reference then you can move (in safe code!) the value out of it via `mem::replace()`. This would then allow to have the value outside the scope of the call.

src/translate.rs
dcz-purism: Addressed by writing a paragraph on the invariants that the implementation must keep.

src/translate.rs
dcz-purism: Addressed as "Safe Rust code must never obtain a mutable Rust reference."

src/translate.rs
dcz-purism: Addressed by spelling out behaviour for references and copies.

dcz-purism:
This should be closer to the ideal now.

sdroege:
Thanks!

dcz-purism:
Glad I could help, and thanks for taking the time to explain and review!

Issue #113, opened 2016-03-07
Add GVariant bindings for non-basic types, nested types: arrays, tuples, dicts, ...


gkoz:
We don't have 'em yet.

sdroege:
While general GVariant support is there since a while, what is missing is support for non-basic, nested types: arrays, tuples, dicts, etc.

wmanley:
I've been working on a pure rust implementation of `GVariant`: [source](https://github.com/wmanley/gvariant-rs), [docs](https://docs.rs/gvariant/0.2.0/gvariant/), [crate](https://crates.io/crates/gvariant).

The design is a little different to GLib's in that the type signatures must be known at compile time.  The behaviour of the crate matches GLib's, at least for all data in "normal form".

I'm preparing to release version 1.0: https://github.com/wmanley/gvariant-rs/issues/1 but before I do I'd like to get some feedback.  Specifically:

@sdroege: could you comment on what it would take for the glib-rs docs to recommend using the gvariant crate instead of the bindings to the GLib GVariant implementation?

If the answer is "not going to happen" that's fine too, I'll just stop worrying about it.

sdroege:
> @sdroege: could you comment on what it would take for the glib-rs docs to recommend using the gvariant crate instead of the bindings to the GLib GVariant implementation?

@wmanley  There's also [zvariant](https://gitlab.freedesktop.org/zeenix/zbus/-/tree/master/zvariant), which unlike yours also provides serialization support and has serde integration. We could probably list native Rust alternatives in the docs (also zbus instead of GDbus once bindings for that are there), do you want to prepare a PR for that?

We should still provide proper GVariant bindings here, independent of all that.

wmanley:
> @wmanley There's also [zvariant](https://gitlab.freedesktop.org/zeenix/zbus/-/tree/master/zvariant), which unlike yours also provides serialization support and has serde integration.

Note: ZVariant is an implementation of the DBus wire format, not GVariant.  The [docs](https://docs.rs/zvariant/2.0.0/zvariant/) say "GVariant is also very common and will be supported by a future version of this crate.".  I discuss other implementations in the docs of the gvariant crate: https://docs.rs/gvariant/0.2.0/gvariant/#comparison-to-and-relationship-with-other-projects

>We could probably list native Rust alternatives in the docs (also zbus instead of GDbus once bindings for that are there), do you want to prepare a PR for that?

Will do.

zeenix:
[...]

PR #588, opened 2020-02-11
Bind API to convert from/to bytes to/from Variant

zeenix:
Bind Variant::new_from_bytes & Variant::get_data_as_bytes().

Related: #113.

### DIFF
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -38,6 +38,7 @@
 //! assert_eq!(num.get_str(), None);
 //! ```
 
+use bytes::Bytes;
 use glib_sys;
 use gobject_sys;
 use gstring::GString;
@@ -140,6 +141,32 @@ impl Variant {
                 _ => None,
             }
         }
+    }
+
+    /// Constructs a new serialised-mode GVariant instance.
+    pub fn new_from_bytes<T: StaticVariantType>(bytes: &Bytes) -> Self {
+        unsafe {
+            from_glib_none(glib_sys::g_variant_new_from_bytes(
+                T::static_variant_type().as_ptr() as *const _,
+                bytes.to_glib_none().0,
+                false.to_glib(),
+            ))
+        }
+    }
+
+    /// Same as `new_from_bytes`, except checks on the passed data are skipped, which makes it a
+    /// potentially unsafe function. You should not use this function on data from external sources.
+    pub unsafe fn new_from_bytes_trusted<T: StaticVariantType>(bytes: &Bytes) -> Self {
+        from_glib_none(glib_sys::g_variant_new_from_bytes(
+            T::static_variant_type().as_ptr() as *const _,
+            bytes.to_glib_none().0,
+            true.to_glib(),
+        ))
+    }
+
+    /// Returns the serialised form of a GVariant instance.
+    pub fn get_data_as_bytes(&self) -> Bytes {
+        unsafe { from_glib_full(glib_sys::g_variant_get_data_as_bytes(self.to_glib_none().0)) }
     }
 }
 


src/variant.rs
```diff
@@ -141,6 +142,19 @@ impl Variant {
             }
         }
     }
+
+    /// Constructs a new serialised-mode GVariant instance.
+    pub fn new_from_bytes<T: StaticVariantType>(bytes: &Bytes, trusted: bool) -> Self {
```
sdroege: If you pass random data and `trusted=true`, will it crash? If so this needs to be an `unsafe fn` (you probably want a safe variant with `trusted=false` and an unsafe with `trusted=true` then).

sdroege:
Maybe also add the `to_bytes()` binding while you're at it? Then we keep it symmetric :)

src/variant.rs
zeenix: I'll check.

zeenix:
> Maybe also add the `to_bytes()` binding while you're at it? Then we keep it symmetric :)

Yeah, i intend to. I just wanted to get early feedback for learning purposes. :)

src/variant.rs
zeenix: i tried a few things (mainly tried to give it less bytes and some curropted bytes) with `trusted=true` but i failed to make it crash. Also looking briefly through the code and docs, it seems unlikely that it skips critical checks when decoding the bytes so I think it's good like this.

sdroege: What is it actually doing then? I feel like this should get an issue reported for improving the docs against GLib :)

zeenix:
> Maybe also add the `to_bytes()` binding while you're at it? Then we keep it symmetric :)

Done. I've not tested that one yet. I'll report it here once done.

sdroege:
> Done. I've not tested that one yet. I'll report it here once done.

Looks good to me as long as that can never return NULL :)

src/variant.rs
zeenix: Actually ebassi says it's UB so I'll go with your approach.

zeenix:
> Looks good to me as long as that can never return NULL :)

Actually it can. :(

> If value is a fixed-sized value that was deserialised from a corrupted serialised container then NULL may be returned. In this case, the proper thing to do is typically to use the appropriate number of nul bytes in place of value . If value is not fixed-sized then NULL is never returned.

This is from docs of `g_variant_get_data` but `g_variant_get_data_as_bytes` docs say the semantics are identical.

sdroege:
> Actually it can. :(

`Option` it is then :)

zeenix:
All done.

EPashkin:
From https://github.com/GNOME/glib/blob/mainline/glib/gvariant-core.c#L938 `g_variant_get_data_as_bytes` always return Bytes,
even in case NULL from `g_variant_get_data`
it doing something strange, but returns Bytes anyway.

EPashkin:
@zeenix Sorry, I was wrong, `g_bytes_new_from_bytes` return NULL if it get wrong parameters, so `Option` is needed.
Thanks, 👍

sdroege:
> @zeenix Sorry, I was wrong, `g_bytes_new_from_bytes` return NULL if it get wrong parameters, so `Option` is needed.

When?

EPashkin:
If bytes_data==NULL, then `data - bytes_data=>data` so `g_bytes_new_from_bytes` get pointer as `offset` parameter and it not pass check in https://github.com/GNOME/glib/blob/mainline/glib/gbytes.c#L227

sdroege:
That's a `g_return_val_if_fail()`, those are basically like panics. Nothing we need to take into account (other than trying to not make it possible to call C API in ways that end up in there).

zeenix:
I removed the `Option`.

sdroege:
Nice

>   process didn't exit successfully: `/home/travis/build/gtk-rs/glib/target/debug/deps/glib-b0a38ca36f7dfe57` (signal: 11, SIGSEGV: invalid memory reference)

On all nightly builds :see_no_evil:

sdroege:
I can confirm, it's the `subclass::object::test::test_create` test that explodes. Unrelated to this PR though.

sdroege:
Regression caused by https://github.com/rust-lang/rust/commit/fc23a81831d5b41510d3261c20c34dd8d32f0f31, hm!

sdroege:
https://github.com/rust-lang/rust/issues/69102

GuillaumeGomez:
Daily breaking changes time is back!

sdroege:
Don't be so dramatic :) It's just a bug, and might even be one on our side that just happened to not trigger for a long long time.

GuillaumeGomez:
My bad... I have memories from the past where every day was a whole new language haha.

PR #589, opened 2020-02-12
Don't store subclass impl/private data in an Option<T>

sdroege:
We don't know the exact memory layout of the Option<T> in relation to T
so can't easily go from a *const T to the surrounding *const Option<T>.

Since nightly from 2020-02-12 using both pointer types interchangeably
causes segmentation faults as their memory layouts are not actually
compatible and never were, but due to some compiler changes this
actually causes crashes at runtime now.

See https://github.com/rust-lang/rust/issues/69102 for details.

As an Option<_> was only used for some paranoid runtime checks that are
not really needed, simply store T directly as impl/private data of
subclasses.

This has the side-effect of having zero-sized impl/private data types to
actually occupy zero bytes of memory, and in some other cases to save a
few bytes of the allocation.

----

CC @EPashkin @GuillaumeGomez

### DIFF
--- a/src/subclass/object.rs
+++ b/src/subclass/object.rs
@@ -565,6 +565,17 @@ mod test {
     }
 
     #[test]
+    fn test_create_child_object() {
+        let type_ = ChildObject::get_type();
+        let obj = Object::new(type_, &[]).unwrap();
+
+        // ChildObject is a zero-sized type and we map that to the same pointer as the object
+        // itself. No private/impl data is allocated for zero-sized types.
+        let imp = ChildObject::from_instance(&obj);
+        assert_eq!(imp as *const _ as *const (), obj.as_ptr() as *const _);
+    }
+
+    #[test]
     fn test_set_properties() {
         let obj = Object::new(SimpleObject::get_type(), &[]).unwrap();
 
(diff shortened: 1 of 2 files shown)


sdroege:
@GuillaumeGomez This wants another bugfix release btw, it's a quite bad bug and I don't fully understand why it didn't explode from the very beginning :)

GuillaumeGomez:
Indeed, thanks for coming up so quickly with this fix!

👍

sdroege:
I have a small additional fix tomorrow, so don't do a release just yet. Tomorrow :)
-- 

GuillaumeGomez:
It's a bit late for a release you know... We still have a few details to discuss about as well!

src/subclass/object.rs
```diff
@@ -564,6 +564,17 @@ mod test {
         assert!(weak.upgrade().is_none());
     }
 
+    #[test]
+    fn test_create_child_object() {
+        let type_ = ChildObject::get_type();
+        let obj = Object::new(type_, &[]).unwrap();
```
GuillaumeGomez: Oh also, please use expect, even in test!

sdroege: The tests in the whole file and in various other places are like this. Let's handle this in a separate PR to make it consistent

sdroege: Created https://github.com/gtk-rs/glib/issues/592 for that.

EPashkin:
👍 Thanks.

sdroege:
Thanks for reviewing :)

sdroege:
Released as 0.9.3 on crates.io

PR #590, opened 2020-02-12
Don't pass an owned string to g_quark_from_static_string()

sdroege:
When called more than once or for a quark that already exists otherwise,
we would always leak the string. Instead pass an unowned string to
g_quark_from_string().

----

CC @GuillaumeGomez @EPashkin

### DIFF
--- a/src/quark.rs
+++ b/src/quark.rs
@@ -13,7 +13,7 @@ pub struct Quark(glib_sys::GQuark);
 
 impl Quark {
     pub fn from_string(s: &str) -> Quark {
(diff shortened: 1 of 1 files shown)


sdroege:
@GuillaumeGomez This should also be part of the bugfix release made for https://github.com/gtk-rs/glib/pull/589. While leaks are safe, they're still annoying and not good :)

GuillaumeGomez:
👍

GuillaumeGomez:
We can add the "from_static_string" implementation by asking a `&'static str` though. Sending a PR for that.

EPashkin:
👍

PR #591, opened 2020-02-12
Add Quark::from_static_string

GuillaumeGomez:
Follow-up of #590.

cc @sdroege @EPashkin

### DIFF
--- a/src/quark.rs
+++ b/src/quark.rs
@@ -16,6 +16,10 @@ impl Quark {
         unsafe { from_glib(glib_sys::g_quark_from_static_string(s.to_glib_full())) }
     }
 
+    pub fn from_static_string(s: &'static str) -> Quark {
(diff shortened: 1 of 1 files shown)


sdroege:
That has exactly the same problem and leaks :)
-- 

GuillaumeGomez:
Then I didn't fully grasp the whole problem. Let's talk about it tomorrow.

sdroege:
The problem is as follows
- First call: we allocate a string (`to_glib_full()`) and pass it to the function, the function stores it
- Second call: we allocate a string and pass it to the function, the function sees that a quark with that name is already registered so does nothing. It leaks the string
- Third call: see above, another leak

The first might also have happened from another part of the code already, in which case we would directly start with the second call and leak.

The only thing that we could do here is to pass the `&'static str` directly as a pointer to this function, but that requires the caller to make sure it is NUL-terminated (which is normally not the case, requires writing `"foo\0"`).

PR #593, opened 2020-02-13
Replace unwrap calls with expect in tests

### LINKED ISSUE #592: Use expect() instead of unwrap() in tests
See https://github.com/gtk-rs/glib/pull/589#discussion_r378562349

CC @GuillaumeGomez

GuillaumeGomez:
Fixes #592.

cc @sdroege @EPashkin

### DIFF
--- a/glib-macros/src/genum_derive.rs
+++ b/glib-macros/src/genum_derive.rs
@@ -261,7 +261,7 @@ pub fn impl_genum(input: &syn::DeriveInput) -> TokenStream {
                     },
                 ];
 
-                let name = std::ffi::CString::new(#gtype_name).unwrap();
+                let name = std::ffi::CString::new(#gtype_name).expect("CString::new failed");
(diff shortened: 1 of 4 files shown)


sdroege:
You didn't update the ones you initially complained about :) See src/subclass/object.rs tests.

GuillaumeGomez:
That was merged after. I'll update the PR shortly.

sdroege:
The tests were already there before that PR :)

GuillaumeGomez:
Then it's cat invasion! :running_man:

GuillaumeGomez:
I think I did all of the ones you added?

sdroege:
Let's merge then. Compared to an `unwrap()` these messages don't bring myself much additional information but you seem to care :)

GuillaumeGomez:
It helps a bit from my point of view so I do care. :)

PR #596, opened 2020-02-16
Add print in case an upgrade failed

GuillaumeGomez:
Just one question: should we put it behind a feature instead?

cc @sdroege @EPashkin

### DIFF
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -135,13 +135,27 @@ macro_rules! to_type_after {
     (@weak $variable:ident , $return_value:expr) => {
         let $variable = match $crate::clone::Upgrade::upgrade(&$variable) {
             Some(val) => val,
-            None => return ($return_value)(),
+            None => {
+                $crate::g_debug!(
+                    $crate::CLONE_MACRO_LOG_DOMAIN,
+                    "Failed to upgrade {}",
+                    stringify!($variable)
+                );
+                return ($return_value)();
+            }
         };
     };
     (as $rename:ident @weak $($variable:ident).+ , $return_value:expr) => {
         let $rename = match $crate::clone::Upgrade::upgrade(&$rename) {
             Some(val) => val,
-            None => return ($return_value)(),
+            None => {
+                $crate::g_debug!(
+                    $crate::CLONE_MACRO_LOG_DOMAIN,
+                    "Failed to upgrade {} {}",
+                    stringify!($rename), "yolo",
+                );
+                return ($return_value)();
+            }
         };
     };
     ($(as $rename:ident)? @strong $($variable:ident).+ , $return_value:expr) => {};
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -195,16 +195,19 @@ macro_rules! to_return_value {
 /// In case something goes wrong inside the `clone!` macro, we use the [`g_debug`] macro. Meaning
 /// that if you want to see these debug messages, you'll have to set the `G_MESSAGES_DEBUG`
 /// environment variable when running your code (either in the code directly or when running the
-/// binary):
+/// binary) to either "all" or [`CLONE_MACRO_LOG_DOMAIN`][crate::CLONE_MACRO_LOG_DOMAIN]:
 ///
 /// ```rust
-/// ::std::env::set_var("G_MESSAGES_DEBUG", "1");
+/// use glib::CLONE_MACRO_LOG_DOMAIN;
+///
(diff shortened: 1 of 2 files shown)


EPashkin:
`if cfg!(debug_assertions)` ?

EPashkin:
same as in https://doc.rust-lang.org/std/macro.debug_assert.html

sdroege:
I don't agree with this change. Upgrade failing is not an error but often wanted behavior. 
-- 

GuillaumeGomez:
No, it's definitely not. Take a look at https://github.com/gtk-rs/examples/pull/294.

GuillaumeGomez:
I'm not too certain using `cfg(debug_assertions)` is a great idea though... Having a debug feature or something along the line might be better I think.

Also, I'll need to add documentation about this as well.

sdroege:
On February 16, 2020 6:00:46 PM GMT+01:00, Guillaume Gomez <notifications@github.com> wrote:
>No, it's definitely not. Take a look at
>https://github.com/gtk-rs/examples/pull/294.

Yes that's a bug in the usage there. I don't want random noise to be printed in correct usage. That the value can disappear safely is the main feature of weak references and not an error.

-- 

GuillaumeGomez:
That's why I want to put it behind a macro. In case you didn't use it correctly and you need more information to debug.

sdroege:
As a separate debugging feature would be acceptable for me.

GuillaumeGomez:
Then I'll do that instead.

EPashkin:
@GuillaumeGomez note that current version don't build
```
error: variable 'variable' is still repeating at this depth
   --> src\clone.rs:136:65
    |
136 |                     println!("Failed to upgrade {}", stringify!($variable));
    |       
```

GuillaumeGomez:
Ah? Strange, it worked fine for me...

GuillaumeGomez:
Fixed the formatting...

src/lib.rs
```diff
@@ -206,6 +206,8 @@ pub use source_futures::*;
 mod thread_pool;
 pub use thread_pool::ThreadPool;
 
+pub const CLONE_MACRO_LOG_DOMAIN: &str = "clone-macro-debug";
```
sdroege: Nobody's going to know what that domain is :) `glib-rs-clone` maybe?

GuillaumeGomez:
Updated the domain as well.

src/clone.rs
```diff
@@ -169,6 +190,23 @@ macro_rules! to_return_value {
 /// closure is immediately returning an optional default return value. If none is provided, `()` is
 /// returned.
 ///
+/// ### Debugging
+///
+/// In case something goes wrong inside the `clone!` macro, we use the [`g_debug`] macro. Meaning
+/// that if you want to see these debug messages, you'll have to set the `G_MESSAGES_DEBUG`
```
sdroege: set to what? :) Don't you need `G_MESSAGES_DEBUG=all` or `G_MESSAGES_DEBUG=glib-rs-clone` or so? The latter for only having logs from this specific domain.

src/lib.rs
```diff
@@ -206,6 +206,8 @@ pub use source_futures::*;
 mod thread_pool;
 pub use thread_pool::ThreadPool;
 
+pub const CLONE_MACRO_LOG_DOMAIN: &str = "glib-rs-clone";
```
sdroege: Maybe `#[doc(hidden)]`?

sdroege:
Otherwise good to be merged for me.

src/clone.rs
GuillaumeGomez: I used `G_MESSAGES_DEBUG=1` and it worked so I don't know? XD

GuillaumeGomez: I wrote how just below already. :p

sdroege: No, you didn't say which log domain it is using. And `G_MESSAGES_DEBUG` takes a string (comma separated listed of domains) or `"all"` for everything. Not `"1"`

src/lib.rs
```diff
@@ -206,6 +206,10 @@ pub use source_futures::*;
 mod thread_pool;
 pub use thread_pool::ThreadPool;
 
+/// This is the log domain used by the [`clone!`][crate::clone] macro. If you want to customize the
+/// message, you can set your own logger using the corresponding `log` functions.
```
sdroege: It's not about customizing the message. It's about having a custom logger that does not go to stdout but for example to syslog.

sdroege:
:+1:

PR #600, opened 2020-02-24
Add manual traits check

GuillaumeGomez:
I can't really make the build simpler considering that all rust version and platforms kinda have unique things to run. It's kind of a mess so I won't touch for the moment. :)

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez thanks,  👍

PR #601, opened 2020-02-27
Various container translation cleanups/fixes

sdroege:
@GuillaumeGomez @EPashkin 

See individual commits. I tried splitting it up as much as possible to make it easier to review. This is also a candidate for a new bugfix release.

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -1792,19 +1792,36 @@ where
         + FromGlibPtrNone<<T as GlibPtrDefault>::GlibType>
         + FromGlibPtrFull<<T as GlibPtrDefault>::GlibType>,
 {
-    unsafe fn from_glib_none_as_vec(ptr: *mut glib_sys::GSList) -> Vec<T> {
-        let num = glib_sys::g_slist_length(ptr) as usize;
-        FromGlibContainer::from_glib_none_num(ptr, num)
+    unsafe fn from_glib_none_as_vec(mut ptr: *mut glib_sys::GSList) -> Vec<T> {
+        let mut res = Vec::new();
+        while !ptr.is_null() {
+            let item_ptr: <T as GlibPtrDefault>::GlibType = Ptr::from((*ptr).data);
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Looks good to me, thanks! Question though: any way to add tests for those fixes to prevent regression?

sdroege:
> Looks good to me, thanks! Question though: any way to add tests for those fixes to prevent regression?

* The first commit is a memory leak, so unless we run tests in valgrind no :)
* The second commit could be tested in theory but it's mostly to mitigate bugs in other codes calling us. It's a situation that must not really happen but is easy to check for in our case.
* The third commit is just an optimization, not functional difference.
* The fourth commit is only minimally reducing the work for the compiler and otherwise makes no difference.

It would be useful to have tests for these things, but they're already tested in other crates via API that actually makes use of the translations. For the list translations I have various tests in gstreamer-rs that make use of them indirectly.

sdroege:
Also the way how I found the second commit was a race condition in gstreamer-rs, insufficient mutex use in unsafe code: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/426

That's not really something testable and something that really shouldn't happen and be fixed elsewhere then :)

GuillaumeGomez:
Ok... It doesn't put me at ease but ok...

sdroege:
We could write some artificial tests for all the translations in glib, but I'd do that as a separate PR. Would also be a good thing for a newcomer to get some better understanding of how the translations are all working.

sdroege:
> We could write some artificial tests for all the translations in glib

I think that would be a good idea in any case, just a bit boring :) Should I create an issue for that, will you, ...?

GuillaumeGomez:
If it's intended for newcomers, please do and add a lot of details. :)

sdroege:
Will do later

sdroege:
macOS builds timing out as usual

PR #603, opened 2020-03-02
glib-macros: add GBoxed derive

gdesmott:
Is there any reason to keep `glib_boxed_type!` and `glib_boxed_derive_traits!` or should I just in-lined them in the Derive macro?

### DIFF
--- a/glib-macros/src/gboxed_derive.rs
+++ b/glib-macros/src/gboxed_derive.rs
@@ -2,8 +2,9 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
-use proc_macro2::{Ident, TokenStream};
+use proc_macro2::{Ident, Span, TokenStream};
 use proc_macro_error::abort_call_site;
+use proc_macro_crate::crate_name;
 use quote::quote;
 
 use crate::utils::{find_attribute_meta, find_nested_meta, parse_type_name};
@@ -40,12 +41,12 @@ fn gen_ptr_to_option(name: &Ident, nullable: bool) -> TokenStream {
     }
 }
 
-fn gen_impl_from_value(name: &Ident) -> TokenStream {
+fn gen_impl_from_value(name: &Ident, crate_ident: &Ident) -> TokenStream {
     quote! {
-        impl<'a> ::glib::value::FromValue<'a> for &'a #name {
-            unsafe fn from_value(value: &'a ::glib::value::Value) -> Self {
-                let ptr = ::glib::gobject_sys::g_value_get_boxed(
-                    ::glib::translate::ToGlibPtr::to_glib_none(value).0,
+        impl<'a> #crate_ident::value::FromValue<'a> for &'a #name {
+            unsafe fn from_value(value: &'a #crate_ident::value::Value) -> Self {
+                let ptr = #crate_ident::gobject_sys::g_value_get_boxed(
+                    #crate_ident::translate::ToGlibPtr::to_glib_none(value).0,
                 );
                 assert!(!ptr.is_null());
                 &*(ptr as *mut #name)
@@ -65,6 +66,16 @@ pub fn impl_gboxed(input: &syn::DeriveInput) -> TokenStream {
         ),
     };
 
+    let crate_name = match crate_name("glib") {
+        Ok(x) => x,
+	Err(_) => {
+            // In case we use it directly from glib itself (it cannot find glib as a dependency
+	    // in this case)
+            "glib".to_owned()
+	}
+    };
+    let crate_ident = Ident::new(&crate_name, Span::call_site());
+
     let meta = find_attribute_meta(&input.attrs, "gboxed")
         .unwrap()
         .unwrap();
@@ -73,7 +84,7 @@ pub fn impl_gboxed(input: &syn::DeriveInput) -> TokenStream {
     let option_to_ptr = gen_option_to_ptr(nullable);
     let ptr_to_option = gen_ptr_to_option(name, nullable);
     let impl_from_value = if !nullable {
-        gen_impl_from_value(name)
+        gen_impl_from_value(name, &crate_ident)
     } else {
         quote! {}
     };
@@ -82,12 +93,12 @@ pub fn impl_gboxed(input: &syn::DeriveInput) -> TokenStream {
         impl BoxedType for #name {
             const NAME: &'static str = #gtype_name;
 
-            fn get_type() -> ::glib::Type {
-                static mut TYPE_: ::glib::Type = ::glib::Type::Invalid;
+            fn get_type() -> #crate_ident::Type {
+                static mut TYPE_: #crate_ident::Type = #crate_ident::Type::Invalid;
                 static ONCE: ::std::sync::Once = ::std::sync::Once::new();
 
                 ONCE.call_once(|| {
-                    let type_ = ::glib::subclass::register_boxed_type::<Self>();
+                    let type_ = #crate_ident::subclass::register_boxed_type::<Self>();
                     unsafe {
                         TYPE_ = type_;
                     }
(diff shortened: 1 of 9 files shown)


glib-macros/src/lib.rs
```diff
@@ -17,3 +18,29 @@ pub fn genum_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+/// Derive macro for defining a [`BoxedType`]`::get_type` function and
+/// the [`glib::Value`] traits.
+///
+/// # Example
+///
+/// ```
+/// use ::glib_macros::GBoxed;
+/// use glib::prelude::*;
+/// use glib::subclass::prelude::*;
+/// use glib::{glib_boxed_type, glib_boxed_derive_traits};
```
gdesmott: Needed so the doc test passes. What would be the good way to import the macros here as I cannot use `#[macro_use]
extern crate glib;`  

```
---- src/lib.rs - gboxed_derive (line 6) stdout ----
error[E0468]: an `extern crate` loading macros must be at the crate root
 --> src/lib.rs:11:1
  |
8 | extern crate glib;
  | ^^^^^^^^^^^^^^^^^^
```

EPashkin: Why you don't like this use? It seems normal for edition2018.
But for better usability IMHO we need find way to glob-use all macros.

EPashkin: Also `/// ##[macro_use] extern crate glib;` works fine for me, see https://doc.rust-lang.org/rustdoc/documentation-tests.html#documenting-macros

EPashkin: Adding `glib_boxed_wrapper, glib_boxed_type, glib_boxed_derive_traits,` to glib::prelude make
`extern crate glib` and `use glib::{glib_boxed_type, ..}` unneeded.

gdesmott: Indeed, I guess I didn't put the `extern crate` before the `use` when I tried yesterday.

sdroege:
> Is there any reason to keep `glib_boxed_type!` and `glib_boxed_derive_traits!` or should I just in-lined them in the Derive macro?

Should go into the derive macro, yes.

glib-macros/src/gboxed_derive.rs
```diff
@@ -0,0 +1,30 @@
...
+        Err(e) => abort_call_site!(
+            "{}: derive(GBoxed) requires #[gboxed(type_name = \"BoxedTypeName\")]",
+            e
+        ),
+    };
+
+    quote! {
+        impl BoxedType for #name {
+            const NAME: &'static str = #gtype_name;
+            glib_boxed_type!();
```
sdroege: Move these into the macro, no need to have them separately anymore

glib-macros/src/gboxed_derive.rs
```diff
@@ -0,0 +1,89 @@
...
+                    glib::translate::ToGlibPtrMut::to_glib_none_mut(value).0,
+                    ptr as *mut _,
+                );
+            }
+        }
+
+        impl<'a> glib::value::FromValueOptional<'a> for &'a #name {
+            unsafe fn from_value_optional(value: &'a glib::value::Value) -> Option<Self> {
+                let ptr = glib::gobject_sys::g_value_get_boxed(
+                    glib::translate::ToGlibPtr::to_glib_none(value).0,
```
sdroege: This is problematic as it assumes that `glib` is available in scope with that specific name. That's not a given.

It would be better to use `$crate::glib` or so here (and if needed re-export `glib` from the crate root with `#[doc(hidden)]`). I don't know how that works with procedural macros though

glib-macros/src/gboxed_derive.rs
```diff
@@ -0,0 +1,89 @@
...
+                glib::gobject_sys::g_value_take_boxed(
+                    glib::translate::ToGlibPtrMut::to_glib_none_mut(value).0,
+                    ptr as *mut _,
+                );
+            }
+        }
+
+        impl glib::value::SetValueOptional for #name {
+            unsafe fn set_value_optional(value: &mut glib::value::Value, this: Option<&Self>) {
+                let this = this.expect("None not allowed");
```
sdroege: This is interesting now. With configuration on the derive macro we could allow nullable and non-nullable types :)

glib-macros/src/gboxed_derive.rs
gdesmott: How is it different from the other `glib::` usages in the existing `GEnum` macro?

`glib-macro` is a dep of `glib` so I'm not sure how to re-export `glib` from the macro crate without introducing a circular dep.

(Btw, isn't this only a problem when not using edition 2018 as external crate are automatically in scope with 2018?)

glib-macros/src/gboxed_derive.rs
gdesmott: Something like that (see new top commit)?

glib-macros/src/gboxed_derive.rs
sdroege: In 2018 edition you can rename crates in your `Cargo.toml`, which I usually do for GStreamer for example (`gstreamer` -> `gst`).

I don't know how that would work with procedural macros. Maybe you can take a look at `failure` / `failure_derive`, which should have the same problem (the derived trait is part of the `failure` crate).

glib-macros/src/gboxed_derive.rs
```diff
@@ -0,0 +1,127 @@
...
+        impl<'a> glib::value::FromValueOptional<'a> for &'a #name {
+            unsafe fn from_value_optional(value: &'a glib::value::Value) -> Option<Self> {
+                let ptr = glib::gobject_sys::g_value_get_boxed(
+                    glib::translate::ToGlibPtr::to_glib_none(value).0,
+                );
+                #ptr_to_option
+            }
+        }
+
+        impl<'a> glib::value::FromValue<'a> for &'a #name {
```
sdroege: This one should not be implemented for nullable types

glib-macros/src/gboxed_derive.rs
gdesmott: They just uses `::failure`, see https://github.com/rust-lang-nursery/failure/blob/master/failure_derive/src/lib.rs#L76

[structop-dervive does the same](https://github.com/TeXitoi/structopt/blob/master/structopt-derive/src/lib.rs#L84) so I'm afraid there is no magic solution here.

glib-macros/src/gboxed_derive.rs
gdesmott: fixed

glib-macros/src/gboxed_derive.rs
sdroege: Then let's do the same here I guess. `::glib`

sdroege: I.e. adding some leading `::` :)

gdesmott: That breaks the doc example and I didn't find any workaround to make it work.
Any idea? I pushed the updated branch.

```
error[E0433]: failed to resolve: could not find `glib` in `{{root}}`
   --> src/subclass/boxed.rs:139:43
    |
139 |     #[derive(Clone, Debug, PartialEq, Eq, glib::GBoxed)]
    |                                           ^^^^^^^^^^^^ could not find `glib` in `{{root}}`
```

sdroege: That's a question for @GuillaumeGomez :) Doesn't the doc example have an `extern crate glib` in it and thus get `glib` at its root?

gdesmott: No, tried that and got: `can't find crate`

GuillaumeGomez: Intriguing. I'll try to take a look today (but don't hesitate to ping me so I don't forget :) ).

sdroege: @GuillaumeGomez ping? :)

GuillaumeGomez: I'm completely overloaded... Will try to take a look this evening. Keeping the tab open.

GuillaumeGomez:
I committed a potential fix to be able to call glib, whatever the context.

sdroege:
If this works on the CI, @gdesmott good to merge from your side?

gdesmott:
Yep, many thanks to @GuillaumeGomez for the fix!

Issue #599, opened 2020-02-23
Change from_glib_borrow() to return a wrapper struct


sdroege:
The idea would be to get rid of the additional bool that is inside all our wrapper structs. It's a waste of space and generally wastes one additional pointer (due to padding) per object :)

Instead that special case would be completely removed from the wrapper structs and moved into a new wrapper struct on top of it, i.e.

```rust
fn from_glib_borrow(ptr) -> glib::Borrowed<T>;

struct Borrowed<T>(T);

impl<T> Deref for Borrowed<T> { ... }
impl<T> AsRef for Borrowed<T> { ... }

impl<T> Drop for Borrowed<T> {
    fn drop(&mut self) {
        mem::forget(self.0);
    }
}
```

This would also make it much less likely that it would accidentally be passed to safe Rust code.

Opinions @EPashkin @GuillaumeGomez ?

GuillaumeGomez:
I need to check the exact differences between the two. Will check tomorrow. We can talk about it directly tomorrow if you prefer too?

EPashkin:
@sdroege I like this idea

sdroege:
@GuillaumeGomez sure just let me know when you're available

sdroege:
> <imperio> sounds good to me, thanks for the explanations!

So now it only has to be implemented and code be changed everywhere :)

sdroege:
cairo needs to be manually updated to a similar approach

PR #605, opened 2020-03-07
Add translate::Borrowed wrapper struct for from_glib_borrow()

sdroege:
Fixes https://github.com/gtk-rs/glib/issues/599

@EPashkin @GuillaumeGomez Any comments/opinions?


src/translate.rs
```diff
@@ -184,6 +184,40 @@ pub struct StashMut<'a, P: Copy, T: ?Sized>(pub P, pub <T as ToGlibPtrMut<'a, P>
...
 
+#[derive(Debug)]
+pub struct Borrowed<T>(Option<T>);
+
+impl<T> Borrowed<T> {
+    pub fn new(val: T) -> Self {
+        Self(Some(val))
+    }
+
+    pub unsafe fn into_inner(mut self) -> T {
```
EPashkin: This IMHO need comment about limits of usage this function

EPashkin: Or maybe better instead add `from_borrowed` constructor?

GuillaumeGomez:
No obvious issues that I can see...

src/translate.rs
sdroege: How would `from_borrowed()` work? I was thinking of that but it didn't seem like an improvement with what I came up with :)

Docs will definitely come if the approach is considered ok

sdroege:
> No obvious issues that I can see...

Any other objections? :)

GuillaumeGomez:
I'll need to reread the PR completely and more carefully and I'll be rested.

src/translate.rs
```diff
@@ -184,6 +184,40 @@ pub struct StashMut<'a, P: Copy, T: ?Sized>(pub P, pub <T as ToGlibPtrMut<'a, P>
 where
     T: ToGlibPtrMut<'a, P>;
 
+#[derive(Debug)]
+pub struct Borrowed<T>(Option<T>);
```
sdroege:
```suggestion
pub struct Borrowed<T>(mem::ManuallyDrop<T>);
```

This is better, it does not involve the annoying `Option` wrapping

EPashkin: What will be with `Borrowed::new(None)` for `Option` ?

src/translate.rs
EPashkin: I think about something like `fn from_borrowed<E, F: FnOnce(E) -> T>(other: Borrowed<E>, f:F)-> Self`
With call
```
                $crate::translate::Borrowed::from_borrowed(
$crate::translate::from_glib_borrow::<_, $crate::boxed::Boxed<_, _>>(ptr),
          |v|   $name( v )
                )
```
If this don't work the :+1: for `into_inner`

EPashkin: or we can just add `fn map<E, F: FnOnce(T) -> E>(self, f:F)-> Borrowed<E>`

src/translate.rs
sdroege: I don't understand :) In that case `T` is `Option<_>` and all is good

EPashkin: Really, I missed that :(

src/translate.rs
sdroege: That's all doable but does not really improve the code and also does not make it any safer :) You still need to ensure that the value passed to the closure is not dropped anywhere.

I'd keep it as is.

sdroege:
Added some docs and cleanup. Any other comments? Otherwise I would start updating everything else and gir

EPashkin:
@sdroege Thanks again and 👍

sdroege:
I have an initial changeset for gir (easier than expected), and gio is compiling again. Will create PRs once everything's compiling.

sdroege:
This would be ready now but don't merge yet (want to prepare the other PRs first).

src/auto/versions.txt
```diff
@@ -1,2 +1,2 @@
-Generated by gir (https://github.com/gtk-rs/gir @ 83cc9ad)
+Generated by gir (https://github.com/gtk-rs/gir @ 997cbcb)
 from gir-files (https://github.com/gtk-rs/gir-files @ 22748fa)
```
EPashkin, line 2: For be same as other maybe do gir-files update too, to 480c117

sdroege: Done

sdroege:
All 9 PRs in place, @GuillaumeGomez please merge and then all will be green again :)

EPashkin:
👍 for all PRs, it builds locally for me.
Only last to do is PR against sourceview when all merged.

Issue #597, opened 2020-02-18
confusing error message from clone!


bilelmoussaoui:
I have tried to use the clone! macro with two use cases and I'm getting a confusing error message 

test case 1 
```
clone!(@strong row_sender => async move {
                    google_font.download().await.and_then(|()| {
                        send!(row_sender, RowAction::Installed);
                        Ok(())
                    });
                });
```

test case 2 
```
       let on_entry_changed = clone!(@weak self.fg_entry, self.bg_entry => move |_entry: &gtk::Entry| {
            let fg_colour = fg_entry.get_text();
            let bg_colour = bg_entry.get_text();

            state.borrow_mut().set_bg_colour(Some(&bg_colour));
            state.borrow_mut().set_fg_colour(Some(&fg_colour));

            send!(sender, Action::ColorChanged);
        });
```

In both cases I'm getting `Variables need to be valid identifiers, e.g. field accesses are not allowed as is, you must rename it!` which is a bit confusing

sdroege:
The second one seems to tell you exactly what you did wrong there :) You need to rename the fields.

The first should work as is in that regard, but you have to provide a closure and not an async block. So the error message could be improved a bit there.

GuillaumeGomez:
And for the first, I think the macro is failing to detect the missing "|whatever|" block.

sdroege:
@GuillaumeGomez Is this fixed now with the latest change?

GuillaumeGomez:
Checking.

GuillaumeGomez:
The first test case now says "no rules expected the token \`async\`". The second says what it should. I'm adding a test case for the first just in case.

PR #609, opened 2020-03-12
Add clone test case for "async"

GuillaumeGomez:
Fixes https://github.com/gtk-rs/glib/issues/597

cc @sdroege @EPashkin

### DIFF
--- a/clone_tests/clone_compilation_errors.py
+++ b/clone_tests/clone_compilation_errors.py
@@ -50,7 +50,7 @@ TESTS = [
     ("clone!(v => move || {})",
         "You need to specify if this is a weak or a strong clone."),
     ("clone!(@strong v => async move {println!(\"foo\");});",
-        "no rules expected the token `async`"),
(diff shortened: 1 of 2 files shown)


clone_tests/clone_compilation_errors.py
```diff
@@ -49,6 +49,8 @@
         "Unknown keyword, only `weak` and `strong` are allowed"),
     ("clone!(v => move || {})",
         "You need to specify if this is a weak or a strong clone."),
+    ("clone!(@strong v => async move {println!(\"foo\");});",
+        "no rules expected the token `async`"),
```
sdroege: This seems like it could get a custom error message on our side. "async blocks are not supported by the clone! macro" or something like that.

GuillaumeGomez: Fine by me!

GuillaumeGomez:
Updated the sentence.

sdroege:
Will merge once the CI is happy. Thanks! :)

PR #611, opened 2020-03-15
Logs

### LINKED ISSUE #610: Add bindings for GLib log
https://developer.gnome.org/glib/stable/glib-Message-Logging.html

https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html

GuillaumeGomez:
Fixes #610.

cc @EPashkin @sdroege

### DIFF
--- a/src/auto/enums.rs
+++ b/src/auto/enums.rs
@@ -314,6 +314,56 @@ impl ErrorDomain for KeyFileError {
     }
 }
 
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
+pub enum LogWriterOutput {
+    Handled,
+    Unhandled,
+    #[doc(hidden)]
+    __Unknown(i32),
+}
+
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+impl fmt::Display for LogWriterOutput {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "LogWriterOutput::{}",
+            match *self {
+                LogWriterOutput::Handled => "Handled",
+                LogWriterOutput::Unhandled => "Unhandled",
+                _ => "Unknown",
+            }
+        )
+    }
+}
+
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+#[doc(hidden)]
+impl ToGlib for LogWriterOutput {
+    type GlibType = glib_sys::GLogWriterOutput;
+
+    fn to_glib(&self) -> glib_sys::GLogWriterOutput {
+        match *self {
+            LogWriterOutput::Handled => glib_sys::G_LOG_WRITER_HANDLED,
+            LogWriterOutput::Unhandled => glib_sys::G_LOG_WRITER_UNHANDLED,
+            LogWriterOutput::__Unknown(value) => value,
+        }
+    }
+}
+
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+#[doc(hidden)]
+impl FromGlib<glib_sys::GLogWriterOutput> for LogWriterOutput {
+    fn from_glib(value: glib_sys::GLogWriterOutput) -> Self {
+        match value {
+            1 => LogWriterOutput::Handled,
+            0 => LogWriterOutput::Unhandled,
+            value => LogWriterOutput::__Unknown(value),
+        }
+    }
+}
+
 #[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
 pub enum OptionArg {
     None,
--- a/src/auto/enums.rs
+++ b/src/auto/enums.rs
@@ -314,56 +314,6 @@ impl ErrorDomain for KeyFileError {
     }
 }
 
-#[cfg(any(feature = "v2_50", feature = "dox"))]
-#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
-pub enum LogWriterOutput {
-    Handled,
-    Unhandled,
-    #[doc(hidden)]
-    __Unknown(i32),
-}
-
-#[cfg(any(feature = "v2_50", feature = "dox"))]
-impl fmt::Display for LogWriterOutput {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(
-            f,
-            "LogWriterOutput::{}",
-            match *self {
-                LogWriterOutput::Handled => "Handled",
-                LogWriterOutput::Unhandled => "Unhandled",
-                _ => "Unknown",
-            }
-        )
-    }
-}
-
-#[cfg(any(feature = "v2_50", feature = "dox"))]
-#[doc(hidden)]
-impl ToGlib for LogWriterOutput {
-    type GlibType = glib_sys::GLogWriterOutput;
-
-    fn to_glib(&self) -> glib_sys::GLogWriterOutput {
-        match *self {
-            LogWriterOutput::Handled => glib_sys::G_LOG_WRITER_HANDLED,
-            LogWriterOutput::Unhandled => glib_sys::G_LOG_WRITER_UNHANDLED,
-            LogWriterOutput::__Unknown(value) => value,
-        }
-    }
-}
-
-#[cfg(any(feature = "v2_50", feature = "dox"))]
-#[doc(hidden)]
-impl FromGlib<glib_sys::GLogWriterOutput> for LogWriterOutput {
-    fn from_glib(value: glib_sys::GLogWriterOutput) -> Self {
-        match value {
-            1 => LogWriterOutput::Handled,
-            0 => LogWriterOutput::Unhandled,
-            value => LogWriterOutput::__Unknown(value),
-        }
-    }
-}
-
 #[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
 pub enum OptionArg {
     None,
--- a/src/auto/functions.rs
+++ b/src/auto/functions.rs
@@ -724,12 +724,6 @@ pub fn listenv() -> Vec<std::ffi::OsString> {
     unsafe { FromGlibPtrContainer::from_glib_full(glib_sys::g_listenv()) }
 }
 
-pub fn log_remove_handler(log_domain: &str, handler_id: u32) {
-    unsafe {
-        glib_sys::g_log_remove_handler(log_domain.to_glib_none().0, handler_id);
-    }
-}
-
 pub fn log_set_always_fatal(fatal_mask: LogLevelFlags) -> LogLevelFlags {
     unsafe { from_glib(glib_sys::g_log_set_always_fatal(fatal_mask.to_glib())) }
 }
--- a/src/auto/functions.rs
+++ b/src/auto/functions.rs
@@ -16,8 +16,6 @@ use Error;
 use FileTest;
 use FormatSizeFlags;
 use GString;
-#[cfg(any(feature = "v2_50", feature = "dox"))]
-use LogField;
 use LogLevelFlags;
 use Pid;
 use Source;
@@ -745,13 +743,10 @@ pub fn log_set_fatal_mask(log_domain: &str, fatal_mask: LogLevelFlags) -> LogLev
     }
 }
 
-#[cfg(any(feature = "v2_50", feature = "dox"))]
-pub fn log_structured_array(log_level: LogLevelFlags, fields: &[&LogField]) {
-    let n_fields = fields.len() as usize;
-    unsafe {
-        glib_sys::g_log_structured_array(log_level.to_glib(), fields.to_glib_none().0, n_fields);
-    }
-}
+//#[cfg(any(feature = "v2_50", feature = "dox"))]
+//pub fn log_structured_array(log_level: LogLevelFlags, fields: /*Ignored*/&[&LogField]) {
+//    unsafe { TODO: call glib_sys:g_log_structured_array() }
+//}
 
 //pub fn log_structured_standard(log_domain: &str, log_level: LogLevelFlags, file: &str, line: &str, func: &str, message_format: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) {
 //    unsafe { TODO: call glib_sys:g_log_structured_standard() }
@@ -769,34 +764,22 @@ pub fn log_variant(log_domain: Option<&str>, log_level: LogLevelFlags, fields: &
 }
 
 //#[cfg(any(feature = "v2_50", feature = "dox"))]
-//pub fn log_writer_default(log_level: LogLevelFlags, fields: &[&LogField], user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> LogWriterOutput {
+//pub fn log_writer_default(log_level: LogLevelFlags, fields: /*Ignored*/&[&LogField], user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> LogWriterOutput {
 //    unsafe { TODO: call glib_sys:g_log_writer_default() }
 //}
 
-#[cfg(any(feature = "v2_50", feature = "dox"))]
-pub fn log_writer_format_fields(
-    log_level: LogLevelFlags,
-    fields: &[&LogField],
-    use_color: bool,
-) -> Option<GString> {
-    let n_fields = fields.len() as usize;
-    unsafe {
-        from_glib_full(glib_sys::g_log_writer_format_fields(
-            log_level.to_glib(),
-            fields.to_glib_none().0,
-            n_fields,
-            use_color.to_glib(),
-        ))
-    }
-}
+//#[cfg(any(feature = "v2_50", feature = "dox"))]
+//pub fn log_writer_format_fields(log_level: LogLevelFlags, fields: /*Ignored*/&[&LogField], use_color: bool) -> Option<GString> {
+//    unsafe { TODO: call glib_sys:g_log_writer_format_fields() }
+//}
 
 //#[cfg(any(feature = "v2_50", feature = "dox"))]
-//pub fn log_writer_journald(log_level: LogLevelFlags, fields: &[&LogField], user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> LogWriterOutput {
+//pub fn log_writer_journald(log_level: LogLevelFlags, fields: /*Ignored*/&[&LogField], user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> LogWriterOutput {
 //    unsafe { TODO: call glib_sys:g_log_writer_journald() }
 //}
 
 //#[cfg(any(feature = "v2_50", feature = "dox"))]
-//pub fn log_writer_standard_streams(log_level: LogLevelFlags, fields: &[&LogField], user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> LogWriterOutput {
+//pub fn log_writer_standard_streams(log_level: LogLevelFlags, fields: /*Ignored*/&[&LogField], user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> LogWriterOutput {
 //    unsafe { TODO: call glib_sys:g_log_writer_standard_streams() }
 //}
 
--- a/src/auto/functions.rs
+++ b/src/auto/functions.rs
@@ -16,10 +16,15 @@ use Error;
 use FileTest;
 use FormatSizeFlags;
 use GString;
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+use LogField;
+use LogLevelFlags;
 use Pid;
 use Source;
 use SpawnFlags;
 use UserDirectory;
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+use Variant;
 
 pub fn access<P: AsRef<std::path::Path>>(filename: P, mode: i32) -> i32 {
     unsafe { glib_sys::g_access(filename.as_ref().to_glib_none().0, mode) }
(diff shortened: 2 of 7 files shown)


src/auto/log_field.rs
```diff
@@ -0,0 +1,5 @@
+// This file was generated by gir (https://github.com/gtk-rs/gir)
+// from gir-files (https://github.com/gtk-rs/gir-files)
+// DO NOT EDIT
+
+use glib_sys;
```
EPashkin: Empty file

GuillaumeGomez: Arf, you saw it before I had time to remove it!

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+}
+
+#[doc(hidden)]
+impl ToGlib for LogField {
+    type GlibType = glib_sys::GLogField;
+
+    fn to_glib(&self) -> glib_sys::GLogField {
+        match self.value {
+            LogValue::String(ref s) => {
+                let s: *const libc::c_char = s.to_glib_none().0;
```
EPashkin: IMHO ToGlibNone for String creates temporary CString and it freed immediately

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+                glib_sys::GLogField {
+                    key: self.key.to_glib_none().0,
+                    value: s as _,
+                    length: -1,
+                }
+            }
+            LogValue::Value(ref v) => glib_sys::GLogField {
+                key: self.key.to_glib_none().0,
```
EPashkin: Not fully sure that this works fine to, we need more comment in GString implementation.

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+    Value(Vec<u8>),
+}
+
+pub struct LogField {
+    key: GString,
+    value: LogValue,
+}
+
+#[doc(hidden)]
+impl ToGlib for LogField {
```
EPashkin: Not better implement `ToGlibPtr` instead? LogField is not "simple type"

src/auto/functions.rs
```diff
@@ -721,86 +726,81 @@ pub fn listenv() -> Vec<std::ffi::OsString> {
...
+        ))
+    }
+}
 
-//#[cfg(any(feature = "v2_50", feature = "dox"))]
-//pub fn log_structured_array(log_level: /*Ignored*/LogLevelFlags, fields: /*Ignored*/&[&LogField]) {
-//    unsafe { TODO: call glib_sys:g_log_structured_array() }
-//}
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+pub fn log_structured_array(log_level: LogLevelFlags, fields: &[&LogField]) {
```
sdroege: @GuillaumeGomez You were not listening to me :P Don't bind the structured logging functions in this round, they require much more effort to get right.

src/log.rs
sdroege: Yes

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+use GString;
+use LogLevelFlags;
+use LogWriterOutput;
+
+pub enum LogValue {
+    String(String),
+    Value(Vec<u8>),
+}
+
+pub struct LogField {
```
sdroege: This and `LogValue` can go away, only needed for structured logging

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+            unsafe {
+                *(ptr.offset(pos as _)) = x.to_glib();
+            }
+        }
+        ptr as *const _
+    }
+}
+
+#[cfg(any(feature = "v2_46", feature = "dox"))]
+pub fn log_set_handler_full<P: Fn(&str, &LogLevelFlags, &str) + Send + Sync + 'static>(
```
sdroege:
```suggestion
pub fn log_set_handler<P: Fn(&str, &LogLevelFlags, &str) + Send + Sync + 'static>(
```

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+            log_domain.to_glib_none().0,
+            log_levels.to_glib(),
+            log_func,
+            Box_::into_raw(super_callback0) as *mut _,
+            destroy_call4,
+        )
+    }
+}
+
+pub fn log_set_handler<P: Fn(&str, &LogLevelFlags, &str) + Send + Sync + 'static>(
```
sdroege: This one will leak the callback, so remove it :) That's why the full variant exists

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+        ptr as *const _
+    }
+}
+
+#[cfg(any(feature = "v2_46", feature = "dox"))]
+pub fn log_set_handler_full<P: Fn(&str, &LogLevelFlags, &str) + Send + Sync + 'static>(
+    log_domain: Option<&str>,
+    log_levels: LogLevelFlags,
+    log_func: P,
+) -> u32 {
```
sdroege: I would make a custom type for the return value, just like `SourceId`. And then also use that for the removal

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+        glib_sys::g_log_set_handler(
+            log_domain.to_glib_none().0,
+            log_levels.to_glib(),
+            log_func,
+            Box_::into_raw(super_callback0) as *mut _,
+        )
+    }
+}
+
+static PRINT_HANDLER: Lazy<Mutex<Option<Box_<Box_<dyn Fn(GString) + Send + Sync + 'static>>>>> =
```
sdroege:
```suggestion
static PRINT_HANDLER: Lazy<Mutex<Option<Box_<Box_<dyn Fn(&str) + Send + Sync + 'static>>>>> =
```

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+            }
+            Err(_) => {
+                // should we log something here?
+            }
+        }
+    }
+    let callback = if func.is_some() {
+        Some(func_func as _)
+    } else {
+        None
```
sdroege: Split this into a `set_print_handler()` and `unset_print_handler()`. Otherwise this is annoying to use for callers

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+            log_func,
+            Box_::into_raw(super_callback0) as *mut _,
+        )
+    }
+}
+
+static PRINT_HANDLER: Lazy<Mutex<Option<Box_<Box_<dyn Fn(GString) + Send + Sync + 'static>>>>> =
+    Lazy::new(|| Mutex::new(None));
+
+pub fn set_print_handler<P: Fn(GString) + Send + Sync + 'static>(func: Option<P>) {
```
sdroege:
```suggestion
pub fn set_print_handler<P: Fn(&str) + Send + Sync + 'static>(func: Option<P>) {
```

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+            // should we log something?
+        }
+    }
+    unsafe { glib_sys::g_set_printerr_handler(callback) };
+}
+
+static PRINTERR_HANDLER: Lazy<Mutex<Option<Box_<Box_<dyn Fn(GString) + Send + Sync + 'static>>>>> =
+    Lazy::new(|| Mutex::new(None));
+
+pub fn set_printerr_handler<P: Fn(GString) + Send + Sync + 'static>(func: Option<P>) {
```
sdroege: All the same comments as for the other function above

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+        let log_domain: GString = from_glib_borrow(log_domain);
+        let log_level = from_glib(log_level);
+        let message: GString = from_glib_borrow(message);
+        let callback: &P = &*(user_data as *mut _);
+        (*callback)(log_domain.as_str(), &log_level, message.as_str());
+    }
+    let log_func = Some(log_func_func::<P> as _);
+    let super_callback: Box_<P> = log_func_data;
+    unsafe {
+        glib_sys::g_log_set_default_handler(log_func, Box_::into_raw(super_callback) as *mut _);
```
sdroege: As this also does not allow a destroy notify, you would leak the callback. Do the local storage like you did for the other two functions above :) There can only be one.

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+/// use glib::{LogLevelFlags, g_log};
+///
+/// g_log!(None, LogLevelFlags::FLAG_RECURSION, "test");
+/// g_log!(Some("test"), LogLevelFlags::FLAG_FATAL, "test");
+///
+/// // You can also pass arguments like in format! or println!:
+/// let x = 12;
+/// g_log!(None, LogLevelFlags::FLAG_RECURSION, "test: {}", x);
+/// g_log!(Some("test"), LogLevelFlags::FLAG_RECURSION, "test: {}", x);
+/// g_log!(None, LogLevelFlags::FLAG_FATAL, "test: {} {}", x, "a");
```
sdroege: I don't think the domain is optional

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+///
+/// ```no_run
+/// use glib::{LogLevelFlags, g_log};
+///
+/// g_log!(None, LogLevelFlags::FLAG_RECURSION, "test");
+/// g_log!(Some("test"), LogLevelFlags::FLAG_FATAL, "test");
+///
+/// // You can also pass arguments like in format! or println!:
+/// let x = 12;
+/// g_log!(None, LogLevelFlags::FLAG_RECURSION, "test: {}", x);
```
sdroege: The flags should not be exposed as flags. They're an enum, with the flags part being internal. Check the docs :)

You basicall want `enum LogLevel { Error, Critical, Warning, Message, Info, Debug }`

src/log.rs
```diff
@@ -0,0 +1,396 @@
...
+/// Macro used to log using GLib logging system. Is uses [g_log].
+///
+/// [g_log]: https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-log)
+///
+/// Example:
+///
+/// ```no_run
+/// use glib::{LogLevelFlags, g_log};
+///
+/// g_log!(None, LogLevelFlags::FLAG_RECURSION, "test");
```
sdroege: In addition you probably want macros like `g_error!("domain", "foo {}", 1)` for each of the levels.

src/log.rs
GuillaumeGomez: I'm not sure to follow the issue here?

sdroege:
```rust
Some(|| {

})
```

is annoying to type/read, and we can easily make it nicer

GuillaumeGomez: I'm still completely lost... What's the link with `(un)set_print_handler`?

sdroege: Currently it's `set_print_handler(Some(|...| { ...})` and that sucks. It should be `set_print_handler(|..| { ... })` and `unset_print_handler()`.

GuillaumeGomez: Oh. I see! Makes sense. You weren't talking about this part of the code specifically but of the function more globally!

sdroege: Yes, sorry :)

src/log.rs
GuillaumeGomez: The problem is that we also have `FLAG_RECURSION` and `FLAG_FATAL`. I think this is why this is not an enum... So do we just ignore those two?

GuillaumeGomez:
I made a first push which clean a few things. However, I still need to add the log macros and we need to clarify the `LogLevelFlags` status.

sdroege:
> I made a first push which clean a few things.

Can you close the discussions above that are handled? That makes it easier to track the status :)

> However, I still need to add the log macros and we need to clarify the `LogLevelFlags` status.

What `LogLevelFlags` status?

src/log.rs
sdroege: They're marked as internal flags in the documentation, so yes :)

GuillaumeGomez:
> Can you close the discussions above that are handled? That makes it easier to track the status :)

I did already, refresh your page. ;)

> What `LogLevelFlags` status?

This one: https://github.com/gtk-rs/glib/pull/611#discussion_r393040477

src/log.rs
sdroege: I.e. ignore them and make this a proper enum

sdroege: Oh no:

```
Sets the log handler for a domain and a set of log levels. To handle fatal and recursive messages the log_levels parameter must be combined with the G_LOG_FLAG_FATAL and G_LOG_FLAG_RECURSION bit flags.
```

I'd handle these two flags as separate boolean parameters to `g_log_set_handler()`. The more tricky bit is `g_log_set_fatal_mask()`, which wants a mask of log levels.

But maybe we need both things then. For normal logging usage you really want an enum for simple and correct use.

GuillaumeGomez: Oh indeed! Makes my life a lot easier. :) I'll keep the flags though (since we're having the C values generated, we should use them after all!), but we should doc hide them. Will check if it's already supported in gir, otherwise I'll add it.

GuillaumeGomez: Ah. Then we can have both: a type builder for log levels which can be set as fatal?

sdroege: You must not pass `FATAL` or `RECURSIVE` to any function apart from `set_handler()`, and a combination of log levels (not the other flags) to `set_fatal_mask()`.

GuillaumeGomez: Ok, so I need to create two separate enums.

sdroege: One enum and one or two flags types, I guess. One flags type if you handle `FATAL` / `RECURSIVE` as additional booleans in the API.

GuillaumeGomez:
Ok so I replaced all `LogLevelFlags` related things. If it's ok with you, I'll then add the missing macros.

src/auto/enums.rs
```diff
@@ -314,6 +314,56 @@ impl ErrorDomain for KeyFileError {
     }
 }
 
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
+pub enum LogWriterOutput {
```
sdroege: Part of structured logging, ignore for now

src/log.rs
```diff
@@ -0,0 +1,443 @@
...
+    fn to_glib(&self) -> u32 {
+        self.0
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum LogFlag {
+    None,
+    Recursion,
+    Fatal,
```
sdroege: This is a flag, use `bitflags!` :)

src/log.rs
```diff
@@ -0,0 +1,443 @@
...
+        } else if value & LogLevelFlags::LEVEL_DEBUG.bits() != 0 {
+            LogLevel::Debug
+        } else {
+            panic!("Unknown log level: {}", value)
+        }
+    }
+}
+
+#[cfg(any(feature = "v2_46", feature = "dox"))]
+pub fn log_set_handler<P: Fn(&str, LogLevel, LogFlag, &str) + Send + Sync + 'static>(
```
sdroege: The level here is a bitmask of all levels that should be allowed

src/log.rs
```diff
@@ -0,0 +1,443 @@
...
+
+pub fn log_remove_handler(log_domain: &str, handler_id: LogHandlerId) {
+    unsafe {
+        glib_sys::g_log_remove_handler(log_domain.to_glib_none().0, handler_id.to_glib());
+    }
+}
+
+pub fn log_set_always_fatal(
+    fatal_level: LogLevel,
+    fatal_flag: LogFlag,
```
sdroege: The flags shouldn't be passed here and the log level is a bitmask of log levels

src/log.rs
```diff
@@ -0,0 +1,443 @@
...
+            fatal_level.to_glib() | fatal_flag.to_glib(),
+        )
+    };
+    (from_glib(ret), from_glib(ret))
+}
+
+pub fn log_set_fatal_mask(
+    log_domain: &str,
+    fatal_level: LogLevel,
+    fatal_flag: LogFlag,
```
sdroege: Same as https://github.com/gtk-rs/glib/pull/611/files#r393632351

src/log.rs
```diff
@@ -0,0 +1,443 @@
...
+        glib_sys::g_log_set_fatal_mask(
+            log_domain.to_glib_none().0,
+            fatal_level.to_glib() | fatal_flag.to_glib(),
+        )
+    };
+    (from_glib(ret), from_glib(ret))
+}
+
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+pub fn log_variant(log_domain: Option<&str>, log_level: LogLevel, fields: &Variant) {
```
sdroege: Structured logging

src/log.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+    unsafe { glib_sys::g_set_printerr_handler(None) };
+}
+
+static DEFAULT_HANDLER: Lazy<
+    Mutex<Option<Box_<Box_<dyn Fn(&str, LogLevel, LogFlag, &str) + Send + Sync + 'static>>>>,
+> = Lazy::new(|| Mutex::new(None));
+
+/// To set back the default print handler, use the [`log_unset_default_handler`] function.
+pub fn log_set_default_handler<P: Fn(&str, LogLevel, LogFlag, &str) + Send + Sync + 'static>(
```
sdroege: Same as https://github.com/gtk-rs/glib/pull/611/files#r393632351

src/log.rs
GuillaumeGomez: And that is a perfect demonstration of why code examples should always be provided. I completely misunderstood the docs...

sdroege: I'm only going from the documentation, so if you disagree with my interpretation let me know :)

src/auto/enums.rs
sdroege: Still there, please remove :)

src/lib.rs
```diff
@@ -181,6 +181,17 @@ mod param_spec;
 pub use param_spec::ParamSpec;
 mod quark;
 pub use quark::Quark;
+#[macro_use]
+mod log;
+#[cfg(any(feature = "v2_46", feature = "dox"))]
```
sdroege: Why 2.46? This existed always or not?

src/lib.rs
```diff
@@ -181,6 +181,17 @@ mod param_spec;
...
+pub use log::log_set_handler;
+#[cfg(any(feature = "v2_50", feature = "dox"))]
+pub use log::log_variant;
```
sdroege: Structured logging, remove

src/log.rs
```diff
@@ -0,0 +1,443 @@
...
+impl ToGlib for LogHandlerId {
+    type GlibType = u32;
+
+    fn to_glib(&self) -> u32 {
+        self.0
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum LogFlag {
```
sdroege: Needs to be bitmask and don't have a none variant

src/log.rs
sdroege: Needs to be a `LogLevelFlags` here

src/log.rs
sdroege: Needs to be a `LogLevelFlags` here

src/log.rs
sdroege: Needs to be a `LogLevelFlags` here

src/log.rs
sdroege: Needs to be a `LogLevelFlags` here

src/log.rs
sdroege: Structured logging, remove

src/lib.rs
sdroege: Still relevant

src/log.rs
```diff
@@ -0,0 +1,442 @@
...
+    type GlibType = u32;
+
+    fn to_glib(&self) -> u32 {
+        self.0
+    }
+}
+
+bitflags! {
+    pub struct LogFlag: u32 {
+        const FLAG_NONE = 0;
```
sdroege: None flag not needed, that's already provided by `empty()`

src/log.rs
```diff
@@ -0,0 +1,442 @@
...
+
+bitflags! {
+    pub struct LogLevels: u32 {
+        const LEVEL_ERROR = 4;
+        const LEVEL_CRITICAL = 8;
+        const LEVEL_WARNING = 16;
+        const LEVEL_MESSAGE = 32;
+        const LEVEL_INFO = 64;
+        const LEVEL_DEBUG = 128;
+        const LEVEL_MASK = 4294967292;
```
sdroege: MASK not needed, that's already provided by `all()`

src/log.rs
```diff
@@ -0,0 +1,442 @@
...
+}
+
+#[cfg(any(feature = "v2_46", feature = "dox"))]
+pub fn log_set_handler<P: Fn(&str, LogLevelFlags, &str) + Send + Sync + 'static>(
+    log_domain: &str,
+    log_levels: LogLevelFlags,
+    log_func: P,
+) -> LogHandlerId {
+    let log_func_data: Box_<P> = Box_::new(log_func);
+    unsafe extern "C" fn log_func_func<P: Fn(&str, LogLevelFlags, &str) + Send + Sync + 'static>(
```
sdroege: I think this should only take a `LogLevel`, single value

src/log.rs
```diff
@@ -0,0 +1,442 @@
...
+
+#[doc(hidden)]
+impl FromGlib<glib_sys::GLogLevelFlags> for LogLevels {
+    fn from_glib(value: glib_sys::GLogLevelFlags) -> LogLevels {
+        LogLevels::from_bits_truncate(value)
+    }
+}
+
+#[cfg(any(feature = "v2_46", feature = "dox"))]
+pub fn log_set_handler<P: Fn(&str, LogLevelFlags, &str) + Send + Sync + 'static>(
```
sdroege: Needs to take the recursion/fatal flags

sdroege: I.e. should take `LogLevels` and `LogFlag` parameter

src/log.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+    unsafe { glib_sys::g_set_printerr_handler(None) };
+}
+
+static DEFAULT_HANDLER: Lazy<
+    Mutex<Option<Box_<Box_<dyn Fn(&str, LogLevels, &str) + Send + Sync + 'static>>>>,
+> = Lazy::new(|| Mutex::new(None));
+
+/// To set back the default print handler, use the [`log_unset_default_handler`] function.
+pub fn log_set_default_handler<P: Fn(&str, LogLevels, &str) + Send + Sync + 'static>(log_func: P) {
```
sdroege: Same comments as for set_log_handler()

src/log.rs
```diff
@@ -0,0 +1,442 @@
...
+    unsafe extern "C" fn func_func(
+        log_domain: *const libc::c_char,
+        log_levels: glib_sys::GLogLevelFlags,
+        message: *const libc::c_char,
+        _user_data: glib_sys::gpointer,
+    ) {
+        match DEFAULT_HANDLER.lock() {
```
sdroege: This lock (and also elsewhere) is problematic, as you hold it while calling the function. If the function causes more logging then we'll have a deadlock.

Better to store an `Arc` inside the mutex, lock, clone the arc, unlock, then call the function.

src/lib.rs
GuillaumeGomez: https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-log-set-handler-full -> 2.46

sdroege: Ah I looked at the non-full version. Maybe we should also do the mutex trick here then to make it more useful? Same effect in the end...

sdroege: Hrm no, not same effect. You can have multiple loggers. Nevermind.

src/log.rs
```diff
@@ -0,0 +1,467 @@
...
+            Err(_) => {
+                // should we log something here?
+                None
+            }
+        } {
+            let string: GString = from_glib_borrow(string);
+            (*callback)(string.as_str())
+        }
+    }
+    let func = Some(Arc::new(Box_::new(Box_::new(func))));
```
sdroege: Make sure to give this the expected type, because below you do a blanket transmute and otherwise this is a bit dangerous.

src/log.rs
```diff
@@ -0,0 +1,471 @@
...
+            LogLevel::Debug
+        } else {
+            panic!("Unknown log level: {}", value)
+        }
+    }
+}
+
+bitflags! {
+    pub struct LogLevels: u32 {
+        const LEVEL_ERROR = 4;
```
sdroege: You don't want to use the values from `glib_sys` here instead of hardcoding 4, etc?

src/log.rs
```diff
@@ -0,0 +1,471 @@
...
+
+#[doc(hidden)]
+impl FromGlib<glib_sys::GLogLevelFlags> for LogLevels {
+    fn from_glib(value: glib_sys::GLogLevelFlags) -> LogLevels {
+        LogLevels::from_bits_truncate(value)
+    }
+}
+
+#[cfg(any(feature = "v2_46", feature = "dox"))]
+pub fn log_set_handler<P: Fn(&str, LogLevel, LogFlags, &str) + Send + Sync + 'static>(
```
sdroege: I don't think the flags should be passed to the closure

src/log.rs
```diff
@@ -0,0 +1,471 @@
...
+
+/// To set back the default print handler, use the [`unset_print_handler`] function.
+pub fn set_print_handler<P: Fn(&str) + Send + Sync + 'static>(func: P) {
+    unsafe extern "C" fn func_func(string: *const libc::c_char) {
+        if let Some(callback) = match PRINT_HANDLER.lock() {
+            Ok(handler) => {
+                if let Some(ref handler) = *handler {
+                    Some(Arc::clone(handler))
+                } else {
+                    panic!("PRINT_HANDLER cannot be None!");
```
sdroege: Why not? You can unset it :)

GuillaumeGomez: If unset, then this callback shouldn't be called.

src/log.rs
```diff
@@ -0,0 +1,471 @@
...
+        }
+    }
+    let func: Option<Arc<Box_<Box_<dyn Fn(&str) + Send + Sync + 'static>>>> =
+        Some(Arc::new(Box_::new(Box_::new(func))));
+    match PRINT_HANDLER.lock() {
+        Ok(mut handler) => {
+            *handler = func;
+        }
+        Err(_) => {
+            // should we log something?
```
sdroege: If locking a mutex fails then things are all wrong anyway. replace with `lock().unwrap()` or `lock().expect("...")` everywhere

src/log.rs
```diff
@@ -0,0 +1,471 @@
...
+        }
+    }
+    unsafe { glib_sys::g_set_printerr_handler(None) };
+}
+
+static DEFAULT_HANDLER: Lazy<
+    Mutex<Option<Arc<Box_<Box_<dyn Fn(&str, LogLevel, LogFlags, &str) + Send + Sync + 'static>>>>>,
```
sdroege: No flags into the closure

src/log.rs
GuillaumeGomez: Oh right! It'd be much better, thanks!

src/log.rs
sdroege: It might've just been called, this is slightly racy so you need to handle that case unfortunately (by doing nothing, just return).

GuillaumeGomez:
Ok, I think I fixed all comments.

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+    type GlibType = u32;
+
+    fn to_glib(&self) -> u32 {
+        match *self {
+            LogLevel::Error => LogLevelFlags::LEVEL_ERROR.bits(),
+            LogLevel::Critical => LogLevelFlags::LEVEL_CRITICAL.bits(),
+            LogLevel::Warning => LogLevelFlags::LEVEL_WARNING.bits(),
+            LogLevel::Message => LogLevelFlags::LEVEL_MESSAGE.bits(),
+            LogLevel::Info => LogLevelFlags::LEVEL_INFO.bits(),
+            LogLevel::Debug => LogLevelFlags::LEVEL_DEBUG.bits(),
```
sdroege: A bit indirect, you could also just go via `glib_sys::G_LOG_LEVEL_DEBUG` etc :)

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    fn to_glib(&self) -> u32 {
+        self.0
+    }
+}
+
+bitflags! {
+    pub struct LogFlags: u32 {
+        const FLAG_RECURSION = LogLevelFlags::FLAG_RECURSION.bits();
+        const FLAG_FATAL = LogLevelFlags::FLAG_FATAL.bits();
```
sdroege: And here `glib_sys::G_LOG_LEVEL_FATAL`

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+bitflags! {
+    pub struct LogLevels: u32 {
+        const LEVEL_ERROR = LogLevelFlags::LEVEL_ERROR.bits();
+        const LEVEL_CRITICAL = LogLevelFlags::LEVEL_CRITICAL.bits();
+        const LEVEL_WARNING = LogLevelFlags::LEVEL_WARNING.bits();
+        const LEVEL_MESSAGE = LogLevelFlags::LEVEL_MESSAGE.bits();
+        const LEVEL_INFO = LogLevelFlags::LEVEL_INFO.bits();
+        const LEVEL_DEBUG = LogLevelFlags::LEVEL_DEBUG.bits();
```
sdroege: And here `glib_sys::G_LOG_LEVEL_DEBUG`, etc

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+    }
+    let func: Option<Arc<Box_<Box_<dyn Fn(&str) + Send + Sync + 'static>>>> =
+        Some(Arc::new(Box_::new(Box_::new(func))));
+    match PRINTERR_HANDLER.lock() {
+        Ok(mut handler) => {
+            *handler = func;
+        }
+        Err(_) => {
+            panic!("Failed to lock PRINTERR_HANDLER to change callback");
```
sdroege: Just `expect()` above :)

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+/// To set the default print handler, use the [`set_print_handler`] function.
+pub fn unset_print_handler() {
+    match PRINT_HANDLER.lock() {
+        Ok(mut handler) => {
+            *handler = None;
+        }
+        Err(_) => {
+            panic!("Failed to lock PRINT_HANDLER to remove callback");
```
sdroege: Just `expect()`

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+    }
+    let func: Option<Arc<Box_<Box_<dyn Fn(&str) + Send + Sync + 'static>>>> =
+        Some(Arc::new(Box_::new(Box_::new(func))));
+    match PRINT_HANDLER.lock() {
+        Ok(mut handler) => {
+            *handler = func;
+        }
+        Err(_) => {
+            panic!("Failed to lock PRINT_HANDLER to change callback");
```
sdroege: Just `expect()`

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+/// To set the default print handler, use the [`set_printerr_handler`] function.
+pub fn unset_printerr_handler() {
+    match PRINTERR_HANDLER.lock() {
+        Ok(mut handler) => {
+            *handler = None;
+        }
+        Err(_) => {
+            panic!("Failed to lock PRINTERR_HANDLER to remove callback");
```
sdroege: `expect()`

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+    }
+    let log_func: Option<Arc<Box_<Box_<dyn Fn(&str, LogLevel, &str) + Send + Sync + 'static>>>> =
+        Some(Arc::new(Box_::new(Box_::new(log_func))));
+    match DEFAULT_HANDLER.lock() {
+        Ok(mut handler) => {
+            *handler = log_func;
+        }
+        Err(_) => {
+            panic!("Failed to lock DEFAULT_HANDLER to change callback");
```
sdroege: `expect()`

src/log.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+/// To set the default print handler, use the [`log_set_default_handler`] function.
+pub fn log_unset_default_handler() {
+    match DEFAULT_HANDLER.lock() {
+        Ok(mut handler) => {
+            *handler = None;
+        }
+        Err(_) => {
+            panic!("Failed to lock DEFAULT_HANDLER to remove callback");
```
sdroege: `expect()`

src/log.rs
```diff
@@ -0,0 +1,572 @@
...
+                $log_level.to_glib(),
+                format!($format, $($arg),*)
+                    // to prevent the glib formatter to look for arguments which don't exist
+                    .replace("%", "%%").to_glib_none().0,
+            );
+        }
+    }};
+}
+
+/// Macro used to log using GLib logging system. Is uses [g_log].
```
sdroege:
```suggestion
/// Macro used to log using GLib logging system. It uses [g_log].
```

sdroege:
And you also want to run `cargo fmt` once

GuillaumeGomez:
Updated.

sdroege:
Will merge once the CI is through. I assume you tested both usage of the logging and the callbacks? :)

Issue #616, opened 2020-03-19
g_log not working with default_handler


GuillaumeGomez:
With the following code, it crashes *after* the callback has been called. I haven't been able to understand why yet...

```rust
extern crate glib;

use glib::*;

use std::sync::{Arc, Mutex};

#[derive(Default)]
struct Counters {
    errors: usize,
    criticals: usize,
    warnings: usize,
    messages: usize,
    infos: usize,
    debugs: usize,
}

fn assert_counts(
    count: &Arc<Mutex<Counters>>,
    errors: usize,
    criticals: usize,
    warnings: usize,
    messages: usize,
    infos: usize,
    debugs: usize,
) {
    let count = count.lock().expect("failed to lock 1");
    assert_eq!(count.errors, errors);
    assert_eq!(count.criticals, criticals);
    assert_eq!(count.warnings, warnings);
    assert_eq!(count.messages, messages);
    assert_eq!(count.infos, infos);
    assert_eq!(count.debugs, debugs);
}

fn main() {
    let count = Arc::new(Mutex::new(Counters::default()));
    let c_count = Arc::clone(&count);
    log_set_default_handler(move |_, level, _| {
        println!("-> {:?}", level);
        match level {
            LogLevel::Error => { (*c_count.lock().expect("failed to lock 2")).errors += 1; }
            LogLevel::Critical => { (*c_count.lock().expect("failed to lock 3")).criticals += 1; }
            LogLevel::Warning => { (*c_count.lock().expect("failed to lock 4")).warnings += 1; }
            LogLevel::Message => { (*c_count.lock().expect("failed to lock 5")).messages += 1; }
            LogLevel::Info => { (*c_count.lock().expect("failed to lock 6")).infos += 1; }
            LogLevel::Debug => { (*c_count.lock().expect("failed to lock 7")).debugs += 1; }
        }
	println!("done!");
    });
    assert_counts(&count, 0, 0, 0, 0, 0, 0);
    g_error!("domain", "hello");
    println!("done?");
    assert_counts(&count, 1, 0, 0, 0, 0, 0);
    g_error!("domain", "hello");
    g_critical!("domain", "hello");
    g_warning!("domain", "hello");
    g_message!("domain", "hello");
    g_info!("domain", "hello");
    g_debug!("domain", "hello");
    g_info!("domain", "hello");
    assert_counts(&count, 2, 1, 1, 1, 2, 1);

    // We now unset our callback and check if it has really been unset.
[...]

PR #617, opened 2020-03-19
Clean up log functions

GuillaumeGomez:
We first need to fix https://github.com/gtk-rs/glib/issues/616.

cc @sdroege @EPashkin

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -115,7 +115,7 @@ fn to_log_flags(fatal: bool, recursion: bool) -> u32 {
 
 #[cfg(any(feature = "v2_46", feature = "dox"))]
 pub fn log_set_handler<P: Fn(&str, LogLevel, &str) + Send + Sync + 'static>(
-    log_domain: &str,
+    log_domain: Option<&str>,
     log_levels: LogLevels,
     fatal: bool,
     recursion: bool,
@@ -152,7 +152,7 @@ pub fn log_set_handler<P: Fn(&str, LogLevel, &str) + Send + Sync + 'static>(
     }
 }
 
-pub fn log_remove_handler(log_domain: &str, handler_id: LogHandlerId) {
+pub fn log_remove_handler(log_domain: Option<&str>, handler_id: LogHandlerId) {
     unsafe {
         glib_sys::g_log_remove_handler(log_domain.to_glib_none().0, handler_id.to_glib());
     }
--- a/src/log.rs
+++ b/src/log.rs
@@ -324,15 +324,13 @@ macro_rules! g_log {
         check_log_args(&$log_domain, $log_level, $format);
         // the next line is used to enforce the type for the macro checker...
         let log_domain: &str = $log_domain;
+        // to prevent the glib formatter to look for arguments which don't exist
         let f = $format.replace("%", "%%");
         unsafe {
             $crate::glib_sys::g_log(
-                // log_domain.to_glib_none().0,
-                b"domain\0".as_ptr() as *const _,
+                log_domain.to_glib_none().0,
                 $log_level.to_glib(),
-                // to prevent the glib formatter to look for arguments which don't exist
-                // f.to_glib_none().0,
-                b"\0".as_ptr() as *const _,
+                f.to_glib_none().0,
             );
         }
     }};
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
As reference, here's the C code:

```C
#include <glib.h>
#include <stdio.h>

#define G_LOG_DOMAIN "osef"

void log_handler(const gchar *log_domain, GLogLevelFlags log_level, const gchar *message, gpointer user_data) {
  printf("[%s:%d] [%s]\n", log_domain, log_level, message);
}

int main() {
  g_log_set_handler(G_LOG_DOMAIN, G_LOG_LEVEL_MASK, log_handler, NULL);
  g_log(G_LOG_DOMAIN, G_LOG_LEVEL_INFO, "hello!");
  g_log(G_LOG_DOMAIN, G_LOG_LEVEL_DEBUG, "hello!");
  return 0;
}
```

compiled with:

```bash
$ gcc test.c `pkg-config --cflags glib-2.0` `pkg-config --libs glib-2.0`
```

sdroege:
> As reference, here's the C code:

Works fine for me without crashes, warnings or valgrind warnings. Trying the Rust code now.

tests/log.rs
```diff
@@ -0,0 +1,105 @@
...
+    assert_eq!(count.warnings, warnings);
+    assert_eq!(count.messages, messages);
+    assert_eq!(count.infos, infos);
+    assert_eq!(count.debugs, debugs);
+}
+
+#[test]
+fn check_log_handlers() {
+    //
+    // log_set_default_handler check part
```
sdroege: Make sure to set the fatal_mask here explicitly so that criticals/warnings are not fatal :) Otherwise the environment might contain `G_DEBUG=fatal_warnings` and things would simply abort on the first critical/warning.

tests/log.rs
```diff
@@ -0,0 +1,105 @@
...
+            }));
+        assert_counts(&count, 0, 0, 0, 0, 0);
+        g_warning!("domain", "hello");
+        assert_counts(&count, 0, 1, 0, 0, 0);
+        g_critical!("domain", "hello");
+        g_message!("domain", "hello");
+        g_info!("domain", "hello");
+        assert_counts(&count, 0, 1, 0, 0, 0);
+        g_debug!("domain", "hello");
+        assert_counts(&count, 0, 1, 0, 0, 1);
```
sdroege: Do some printfs with a different domain to see that your callback is not called for them

tests/log.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    //
+    // log_set_handler check part
+    //
+    #[cfg(feature = "v2_46")]
+    {
+        let count = Arc::new(Mutex::new(Counters::default()));
+        // We set the handler for both warning and debug.
+        let handler_id = log_set_handler(
+            "domain",
```
sdroege: The domain can be `NULL`, so we should make it an `Option<_>`. If `NULL` it matches on all domains.

GuillaumeGomez:
Updated!

src/log.rs
```diff
@@ -114,27 +114,28 @@ fn to_log_flags(fatal: bool, recursion: bool) -> u32 {
 }
 
 #[cfg(any(feature = "v2_46", feature = "dox"))]
-pub fn log_set_handler<P: Fn(&str, LogLevel, &str) + Send + Sync + 'static>(
-    log_domain: &str,
+pub fn log_set_handler<P: Fn(Option<&str>, LogLevel, &str) + Send + Sync + 'static>(
```
sdroege: No, the domain that is passed to the callback is always not `NULL` :)

```suggestion
pub fn log_set_handler<P: Fn(&str, LogLevel, &str) + Send + Sync + 'static>(
```

src/log.rs
```diff
@@ -114,27 +114,28 @@ fn to_log_flags(fatal: bool, recursion: bool) -> u32 {
...
+    log_domain: Option<&str>,
     log_levels: LogLevels,
     fatal: bool,
     recursion: bool,
     log_func: P,
 ) -> LogHandlerId {
     let log_func_data: Box_<P> = Box_::new(log_func);
-    unsafe extern "C" fn log_func_func<P: Fn(&str, LogLevel, &str) + Send + Sync + 'static>(
+    unsafe extern "C" fn log_func_func<P: Fn(Option<&str>, LogLevel, &str) + Send + Sync + 'static>(
```
sdroege:
```suggestion
    unsafe extern "C" fn log_func_func<P: Fn(&str, LogLevel, &str) + Send + Sync + 'static>(
```

sdroege: etc

src/log.rs
```diff
@@ -114,27 +114,28 @@ fn to_log_flags(fatal: bool, recursion: bool) -> u32 {
 }
 
 #[cfg(any(feature = "v2_46", feature = "dox"))]
-pub fn log_set_handler<P: Fn(&str, LogLevel, &str) + Send + Sync + 'static>(
-    log_domain: &str,
+pub fn log_set_handler<P: Fn(Option<&str>, LogLevel, &str) + Send + Sync + 'static>(
+    log_domain: Option<&str>,
```
sdroege: And this is ok now :)

GuillaumeGomez:
Updated!

sdroege:
You want to run `cargo fmt`, otherwise we're good here now.

PR #618, opened 2020-03-24
Improve clone! error messages

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -94,10 +94,25 @@ macro_rules! to_type_before {
     (@weak $($variable:ident).+ as $rename:ident) => (
         let $rename = $crate::clone::Downgrade::downgrade(&$($variable).+);
     );
-    (@ $keyword:ident $($variable:ident).+ $(as $rename:ident)?) => (
+    // The two following cases are just here so "@strong" and "@weak" aren't detected as invalid
+    // when passing an expression (like "@default-return" => "-return" is the start of an expression
+    // there).
+    (@strong $variable:expr) => (
+        let $variable = $variable.clone();
+    );
+    (@weak $variable:expr) => (
+        let $variable = $crate::clone::Downgrade::downgrade(&$variable);
+    );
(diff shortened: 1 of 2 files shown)


src/clone.rs
```diff
@@ -94,10 +94,16 @@ macro_rules! to_type_before {
     (@weak $($variable:ident).+ as $rename:ident) => (
         let $rename = $crate::clone::Downgrade::downgrade(&$($variable).+);
     );
-    (@ $keyword:ident $($variable:ident).+ $(as $rename:ident)?) => (
+    (@strong $variable:expr) => (
+        let $variable = $variable.clone();
+    );
+    (@weak $variable:expr) => (
```
sdroege: Can we get tests for these two cases?

src/clone.rs
```diff
@@ -420,6 +426,9 @@ macro_rules! clone {
     ($($(@ $strength:ident)? $($variables:ident).+ $(as $rename:ident)?),+ => async $($x:tt)+ ) => (
         compile_error!("async blocks are not supported by the clone! macro");
     );
+    ($($(@ $strength:ident)? $variables:expr),+ => move || $($_:tt)* ) => (
```
sdroege: And for this?

src/clone.rs
GuillaumeGomez: They can't get called actually, it's only to prevent the macro to fail on "$variable:expr". I can add a comment explaining it though.

src/clone.rs
GuillaumeGomez: I added a test for it (the "default-return" one).

src/clone.rs
sdroege: If it can't be called, why would it be needed then? :) (Yes please add an explanation comment)

GuillaumeGomez: I added the comment. Can you tell me if it's clear enough please?

PR #620, opened 2020-03-31
glib-macros: add gflags attribute macro

gdesmott:


glib-macros/Cargo.toml
```diff
@@ -19,6 +19,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = "1.0"
 proc-macro-crate = "0.1"
+bitflags = "1.0"
```
sdroege: Do we have to depend on it if we only expand into code that uses the macro? The user of our macro will have to depend on bitflags AFAIU.

Or are you re-exporting it and using the re-exported version from the generated code?

glib-macros/src/gflags_attribute.rs
```diff
@@ -0,0 +1,184 @@
...
+        let disc = v.discriminant.as_ref().expect("missing discriminant");
+        let value = &disc.1;
+
+        quote_spanned! {v.span()=>
+            const #name = #value;
+        }
+    });
+
+    quote! {
+        bitflags::bitflags! {
```
sdroege: Maybe should re-export it hidden from the `glib` crate and explicitly use that from here?

glib-macros/src/gflags_attribute.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let bitflags = gen_bitflags(name, enum_variants);
+
+    let get_type = format_ident!("{}_get_type", name.to_string().to_snake_case());
+    let (gflags_values, nb_gflags_values) = gen_gflags_values(name, enum_variants);
+
+    quote! {
+        #bitflags
+
+        impl glib::translate::ToGlib for #name {
```
sdroege: Won't the `glib::` part in the generated code fail if you add a test using the macro to the `glib` crate?

glib-macros/src/lib.rs
```diff
@@ -38,8 +40,48 @@ pub fn genum_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+///
+/// #[gflags("MyFlags")]
+/// enum MyFlags {
+///     #[gflags(name = "Flag A", nick = "nick-a")]
+///     A = 0b00000001,
+///     #[gflags(name = "Flag B")]
+///     B = 0b00000010,
+///     #[gflags(skip)]
+///     AB = Self::A.bits() | Self::B.bits(),
+///     C = 0b00000100,
```
sdroege: Maybe we can add something to the macro so that it gives a compiler error if there's an unskipped overlapping value?

glib-macros/src/gflags_attribute.rs
GuillaumeGomez: Should be replace with `$crate`, indeed.

sdroege: That's not how it works with proc macros, but yes :)

GuillaumeGomez: For proc-macros, I already provided a solution in the other PR of @gdesmott :)

glib-macros/src/lib.rs
gdesmott: Not sure how you'd check for overlapped values?

sdroege: Good question, that would require actually evaluating the code to get the bit pattern it evaluates to. So maybe not :)

glib-macros/src/gflags_attribute.rs
gdesmott: Right, we forgot to fix the other macro which I used as a base for this one. Fixed both.

gdesmott:
All fixed. I also added a property example using it, it's really easy now. :)

PR #623, opened 2020-04-05
glib-macros: Update itertools to 0.9

jplatte:


sdroege:
@GuillaumeGomez Hrm, need to also fix/update the gtk4 crates. They're failing the CI now.

jplatte:
@sdroege Same for webkit2gtk. I'm currently using all the git versions of the gtk-rs crates because it hasn't seen a release in some time and the released version doesn't yet contain the code that allows user stylesheets.

sdroege:
The webkit bindings are maintained by @antoyo, best to check that with him.

PR #624, opened 2020-04-09
Use mem::ManuallyDrop instead of mem::forget() everywhere

sdroege:
It makes the intentions clearer and potentially results in simpler
assembly, at least in debug builds.


EPashkin:
Good idea

PR #625, opened 2020-04-09
Remove unnecessary transmute calls

GuillaumeGomez:
I removed some `transmute` calls and added comments where they were missing.

cc @sdroege @EPashkin

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -343,7 +343,7 @@ pub trait Cast: ObjectType {
         if !self.is::<T>() {
             None
         } else {
-            // This transmute is safe because all our wrapper types have the
+            // This cast is safe because all our wrapper types have the
             // same representation except for the name and the phantom data
             // type. IsA<> is an unsafe trait that must only be implemented
             // if this is a valid wrapper type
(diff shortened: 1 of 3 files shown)


src/object.rs
```diff
@@ -1541,6 +1541,7 @@ impl<T: ObjectType> ObjectExt for T {
         ::signal::connect_raw(
             self.as_object_ref().to_glib_none().0,
             signal_name.as_ptr() as *const _,
+            // Two level-kind of callback, which explains why the transmute is required.
```
sdroege, line 1544: I don't understand the comment

GuillaumeGomez: I might have been a bit too lazy writing this comment, let me improve it.

sdroege: Same for the other such comments. What's "two levels" here exactly, or "two times" (below)?

src/object.rs
```diff
@@ -1541,7 +1541,7 @@ impl<T: ObjectType> ObjectExt for T {
         ::signal::connect_raw(
             self.as_object_ref().to_glib_none().0,
             signal_name.as_ptr() as *const _,
-            Some(mem::transmute(notify_trampoline::<Self, F> as usize)),
+            Some(*(&notify_trampoline::<Self, F> as *const _ as *const _)),
```
EPashkin, line 1544: Interesting how rust found type for first `_` in `as *const _ as *const _ `  😉

EPashkin:
👍

src/object.rs
GuillaumeGomez: Pointers are a weird thing in rust. You have to do it twice every time to convert between pointers of non primitives types.

sdroege: The first `as *const _` casts the `&T` into a `*const T` (there is only one option for `T` here, you can't do transitive casts from references). The second `as *const _` is from `*const T` to `*const glib_sys::GCallback` (which is what is required to be the final type).

EPashkin: Thanks

Issue #613, opened 2020-03-19
Lots of Error::description() deprecation warnings with 1.42


sdroege:
We should either move to `thiserror` like in cairo for our errors, or at least just not implement that function.

sdroege:
@GuillaumeGomez Any objections against using `thiserror`?

GuillaumeGomez:
I'll complain but you can go forward with it.

gdesmott:
Just build `glib` master with 1.42 and didn't get any warning. Do I have to use specific flags?

GuillaumeGomez:
@gdesmott You're sure you didn't allow all warnings in your environment variables by any chance? Also, try `rustc --version` just to be sure. :)

EPashkin:
Just glib or gtk/examples?

gdesmott:
Just glib, don't have any warning building https://github.com/gtk-rs/examples either.

```
$ rustc --version
rustc 1.42.0 (b8cedc004 2020-03-09)
```

EPashkin:
Then issue fixed without linking it in PRs. Ex. my cairo PR about this was before this issue.

sdroege:
It's not fixed, see for example the two `Error` impls in `glib/src/error.rs`. We need to go through all `Error` impls in our crates.

gdesmott:
ah, I get the warnings when running `cargo test`!

gdesmott:
Not sure how we'd like to change the test with this change. I suppose hardcoding source code line numbers in tests isn't ideal.

```
---- subclass::object::test::test_set_properties stdout ----
thread 'subclass::object::test::test_set_properties' panicked at 'assertion failed: `(left == right)`
  left: `"Error \"property not found\" in \"glib::object\" at src/object.rs:1365"`,
 right: `"property not found"`', src/subclass/object.rs:605:9
```

sdroege:
CC @fengalin who added this test

sdroege:
The same will be needed in other repositories.

fengalin:
I'll see what I can do, but I'd like to focus on completing https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/298

sdroege:
@gdesmott actually was looking into that (`Error::description`) before you took over :)

fengalin:
Ok :) Let's synchronize here if someone starts updating another repo for this.

gdesmott:
[...]

PR #627, opened 2020-04-10
Remove deprecated Error::description impl

fengalin:
`std::error::Error::description` is deprecated since 1.42.0. Doc says
`Display` should be used instead. Current implementation for `Display`
on `BoolError` displays all the fields, which can be obtained via the
`Debug` derive.

This commit changes the `Display` impl on `BoolError` to match
previous impl for the `Error::description`. Tests can use
`to_string` in place of `description` for the same result.

Same for `glib::Error` and `value::GetError`.

Fixes https://github.com/gtk-rs/glib/issues/613

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -92,12 +92,6 @@ impl fmt::Display for Error {
     }
 }
 
-impl error::Error for Error {
-    fn description(&self) -> &str {
-        self.message()
-    }
-}
-
 impl fmt::Debug for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Error")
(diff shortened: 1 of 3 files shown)


sdroege:
There's also `glib:Error` that still implements the description function

sdroege:
CC @gdesmott

fengalin:
> There's also `glib:Error` that still implements the description function

Oh, didn't noticed. Will take a look shortly.

EPashkin:
👍

sdroege:
Please update the commit message as this is not only about `BoolError` anymore, otherwise looks good to me :)

fengalin:
You mean the PR message? The commit message was already edited. :) Or does it need additional details?

sdroege:
The first line of the commit message still says "BoolError: remove deprecated Error::description impl "

fengalin:
Indeed

sdroege:
Thanks! :)

PR #632, opened 2020-04-14
Add test for callbacks validity

GuillaumeGomez:
With this new test alongside https://github.com/gtk-rs/gio/pull/289, we should prevent the mess we got recently to happen again. Next step: extra check on gir CI!

cc @sdroege @EPashkin

### DIFF
--- a/src/subclass/object.rs
+++ b/src/subclass/object.rs
@@ -700,6 +700,23 @@ mod test {
         assert_eq!(value.get::<String>(), Ok(Some("return value".to_string())));
     }
 
+    #[test]
+    fn test_callback_validity() {
+        use std::sync::{Arc, Mutex};
+
+        let type_ = SimpleObject::get_type();
+        let obj = Object::new(type_, &[("name", &"old-name")]).expect("Object::new failed");
+
+        let name_changed_triggered = Arc::new(Mutex::new(false));
+        let name_changed_clone = name_changed_triggered.clone();
+
+        obj.connect_notify(Some("name"), move |_, _| {
+            *name_changed_clone.lock().expect("Failed to lock") = true;
+        });
+        obj.notify("name");
+        assert!(*name_changed_triggered.lock().expect("Failed to lock"));
+    }
+
     // Note: can't test type mismatch in signals since panics accross FFI boundaries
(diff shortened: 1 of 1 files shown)


src/subclass/object.rs
```diff
@@ -692,6 +692,9 @@ mod test {
             Some("return value".to_value())
         })
         .expect("Failed to connect on 'create-string'");
+        obj.connect_notify(Some("create-string"), |_, _| {
+            println!("hello!");
```
sdroege, line 696: This is not super useful :) Instead of printing something, change some `Arc<AtomicBool>` or something from `false` to `true` here and assert that afterwards

sdroege: Also this will never be called: notify is for property changes.

src/subclass/object.rs
```diff
@@ -700,6 +703,17 @@ mod test {
         assert_eq!(value.get::<String>(), Ok(Some("return value".to_string())));
     }
 
+    #[test]
+    fn test_callback_validity() {
+        let type_ = SimpleObject::get_type();
+        let obj = Object::new(type_, &[("name", &"old-name")]).expect("Object::new failed");
+
+        obj.connect_notify(Some("name"), |_, _| {
+            println!("hello!");
```
sdroege, line 712: See https://github.com/gtk-rs/glib/pull/632/files#r408401174

src/subclass/object.rs
GuillaumeGomez: Well, it crashed for me at least? (I used a pre-fix version)

sdroege: Well that's a bug then. This callback should never be called

src/subclass/object.rs
```diff
@@ -700,6 +703,17 @@ mod test {
...
+        });
+        obj.notify("name");
```
sdroege, line 714: This one here is called as part of the `notify()` call, that's fine. Just check for it to actually happen.

src/subclass/object.rs
GuillaumeGomez: Oh my bad, I thought it was part of the test I added, this one shouldn't have been there. Removed!

PR #633, opened 2020-04-30
Return a string of any lifetime from Quark::to_string()

sdroege:
Specifying it as 'static is very specific and instead we can let the
caller chose any lifetime they want here.

CC @GuillaumeGomez @EPashkin

### DIFF
--- a/src/quark.rs
+++ b/src/quark.rs
@@ -17,7 +17,7 @@ impl Quark {
     }
 
     #[allow(clippy::trivially_copy_pass_by_ref)]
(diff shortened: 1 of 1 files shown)


heftig:
A static reference can be reborrowed into a reference with any lifetime. Why is this needed?

sdroege:
See https://github.com/rust-lang/rust/pull/42417

EPashkin:
Oh, then we have no choice 😭

GuillaumeGomez:
:+1:

PR #634, opened 2020-05-02
VariantDict support

kinnison:
This is the first step along the path to resolving https://github.com/gtk-rs/gio/issues/291


sdroege:
Looks good to me, I just don't know about the docs. @GuillaumeGomez shouldn't this be handled by rustdoc-stripper and automatically filled in from the C docs?

kinnison:
Apart from the gtk4 failures, CI is green.  As far as I can tell, we don't use rustdoc-stripper here, there's certainly to use of lgpl-docs crate that I can see, so I don't know where it would apply from.

PR #635, opened 2020-05-03
Clippy cleanliness and CI enforcement

kinnison:
This adds CI enforcement of clippy and cleans up a lot of its unhappiness.

There are a number of places where I silence lints rather than solving them, and care will need to be taken to consider unsilencing and fixing them properly in the future, but this is sufficient for now.

### DIFF
--- a/src/byte_array.rs
+++ b/src/byte_array.rs
@@ -244,9 +244,9 @@ mod tests {
         ba.remove_range(1, 2);
         ba.sort(|a, b| a.cmp(b));
         unsafe { ba.set_size(3) };
-        assert_eq!(ba, "aab".as_bytes());
+        assert_eq!(ba, b"aab" as &[u8]);
         let abc: &[u8] = b"abc";
-        assert_eq!(ByteArray::from(abc), "abc".as_bytes());
+        assert_eq!(ByteArray::from(abc), b"abc" as &[u8]);
     }
 
     #[test]
--- a/src/byte_array.rs
+++ b/src/byte_array.rs
@@ -97,6 +97,7 @@ impl ByteArray {
         }
     }
 
+    #[allow(clippy::missing_safety_doc)]
     pub unsafe fn set_size(&self, size: usize) {
         glib_sys::g_byte_array_set_size(self.to_glib_none().0, size as u32);
     }
--- a/src/gstring.rs
+++ b/src/gstring.rs
@@ -417,6 +417,7 @@ impl_from_glib_container_as_vec_string!(GString, *const c_char);
 impl_from_glib_container_as_vec_string!(GString, *mut c_char);
 
 #[cfg(test)]
+#[allow(clippy::blacklisted_name)]
 mod tests {
     use glib_sys;
     use gstring::GString;
--- a/src/gstring.rs
+++ b/src/gstring.rs
@@ -476,7 +476,7 @@ mod tests {
 
     #[test]
     fn test_vec_u8_to_gstring() {
-        let v = "foo".as_bytes();
+        let v: &[u8] = b"foo";
         let s: GString = Vec::from(v).into();
         assert_eq!(s.as_str(), "foo");
     }
--- a/src/gstring.rs
+++ b/src/gstring.rs
@@ -33,11 +33,37 @@ unsafe impl Send for GString {}
 unsafe impl Sync for GString {}
 
 impl GString {
+    /// Create a new GString from a glib-originated string, taking ownership.
+    ///
+    /// # Safety
+    ///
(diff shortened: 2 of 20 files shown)


src/byte_array.rs
```diff
@@ -244,9 +245,9 @@ mod tests {
         ba.remove_range(1, 2);
         ba.sort(|a, b| a.cmp(b));
         unsafe { ba.set_size(3) };
-        assert_eq!(ba, "aab".as_bytes());
+        assert_eq!(ba, b"aab" as &[u8]);
```
EPashkin, line 248: Maybe better use `&b"aab"[..]`

kinnison: I'm happy either way.  If you would prefer the `[..]` form over the `as &[u8]` form, then I'll change this, and other instances of this kind of change.  It's basically aesthetic so I shall defer to you.  Just say if you want me to change those up.

EPashkin: Lets leave it as is

kinnison: Ok :+1:  Thanks for reviewing.

src/gstring.rs
```diff
@@ -33,11 +33,37 @@ unsafe impl Send for GString {}
...
+    /// # Safety
+    ///
+    /// The provided string must be a valid C string (i.e. `'0'` terminated).
+    ///
+    /// It is essential that noone else free this pointer as it owned by the
+    /// returned `GString` until the borrow is dropped.
+    ///
+    /// The underlying string must not be mutated, in particular in terms of
+    /// length, underneath the `GString` instance for the duration of the borrow.
     pub unsafe fn new_borrowed(ptr: *const c_char) -> Borrowed<Self> {
```
sdroege, line 67: Why do we have this and the other constructor at all? IMHO they should be removed, the trait impls are enough and what is used everywhere else.

src/subclass/object.rs
```diff
@@ -659,7 +660,7 @@ mod test {
                 .expect("Failed to get str from args[1]");
 
             assert_eq!(name, "new-name");
-            *name_changed_clone.lock().expect("Failed to lock") = true;
+            name_changed_clone.store(true, Ordering::Relaxed);
```
sdroege, line 663: Is relaxed ordering enough here (and elsewhere)? `SeqCst` would be the safe choice.

src/gstring.rs
GuillaumeGomez: It's not supposed to be handled by users so it's strange to have them. Why did we add these functions in the first place? To not use `from_glib`?

sdroege: I think just an oversight. Let's remove them then, @kinnison :)

src/subclass/object.rs
kinnison: This is a basic test and not actually using parallelism. As far as I can tell the use of `Mutex<bool>` could just as easily have been replaced with `RefCell<bool>` as `AtomicBool`, I just went with the default conversion I've used in test suites in the past.  Your call.

src/gstring.rs
kinnison: If I delete them and nothing goes wrong in the CI then I guess it's okay, though we should remember to boost the minor to 0.10 on the next release since this is incompatible.

kinnison: To remove them I'd need to shunt stuff around a lot.  I could do that, but these are used in many traits.  Perhaps better would be to simply make the `pub(crate)` or even private since they're only used inside `gstring.rs` ?

sdroege: That's going to happen anyway, we already had many other incompatible changes. The reason why the version is still at 0.9 in git is something @GuillaumeGomez has to explain to you (I disagree and would've bumped it to 0.10 long ago)

src/subclass/object.rs
sdroege: My rule is usually that it's wrong unless someone can explain to me why anything else than `SeqCst` is correct :) You're right here with the test of course, so let's leave it.

src/gstring.rs
sdroege: Private also works for me, just not public :)

src/subclass/object.rs
kinnison: I use `Relaxed` when I know it's safe (e.g in these tests) and then whenever I see `Relaxed` I know to be aware that if I change things I may need to reassess.  I'll leave this as-is.

src/gstring.rs
kinnison: I think private is fine. I've added a commit to that effect.

kinnison:
For reference, the only CI failures were, once again, the bitrotted gtk4 stuff

PR #636, opened 2020-05-03
Early steps toward supporting lgpl-docs

kinnison:
This PR covers the early steps of supporting the lgpl-docs repository in glib.

Unfortunately there is a lot of files currently put in the ignore list, though this does mean we import at least some docs into the autogenerated-from-gir code.  Further work will be needed to work out what can be transformed from existing docs into lgpl-docs content, versus what needs to remain ignored.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,53 @@
+fn main() {
+    manage_docs();
+}
+
+#[cfg(any(feature = "embed-lgpl-docs", feature = "purge-lgpl-docs"))]
+fn manage_docs() {
+    extern crate lgpl_docs;
+    const PATH: &str = "src";
+    const IGNORES: &[&str] = &[
+        "lib.rs",
+        "prelude.rs",
+        "signal.rs",
+        "boxed.rs",
+        "byte_array.rs",
+        "bytes.rs",
+        "char.rs",
+        "clone.rs",
+        "enums.rs",
+        "error.rs",
+        "gobject/mod.rs",
+        "log.rs",
+        "main_context.rs",
+        "main_context_channel.rs",
+        "main_context_futures.rs",
+        "object.rs",
+        "send_unique.rs",
+        "shared.rs",
+        "source.rs",
+        "source_futures.rs",
+        "string.rs",
+        "subclass/boxed.rs",
+        "subclass/interface.rs",
+        "subclass/mod.rs",
+        "subclass/object.rs",
+        "subclass/simple.rs",
+        "subclass/types.rs",
+        "translate.rs",
+        "types.rs",
+        "utils.rs",
+        "value.rs",
+        "variant.rs",
+        "variant_dict.rs",
+        "variant_type.rs",
+        "wrapper.rs",
+    ];
+    lgpl_docs::purge(PATH, IGNORES);
+    if cfg!(feature = "embed-lgpl-docs") {
+        lgpl_docs::embed(lgpl_docs::Library::Glib, PATH, IGNORES);
+    }
+}
+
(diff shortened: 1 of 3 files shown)


GuillaumeGomez (approved):
:+1:

build.rs
```diff
@@ -0,0 +1,53 @@
...
+        "utils.rs",
+        "value.rs",
+        "variant.rs",
+        "variant_dict.rs",
+        "variant_type.rs",
+        "wrapper.rs",
+    ];
+    lgpl_docs::purge(PATH, IGNORES);
+    if cfg!(feature = "embed-lgpl-docs") {
+        lgpl_docs::embed(lgpl_docs::Library::Glib, PATH, IGNORES);
```
sdroege, line 48: There's also `src/auto/gobject` from `Gir_GObject.toml` btw, we have two libraries in one here

build.rs
```diff
@@ -0,0 +1,53 @@
...
+        "boxed.rs",
+        "byte_array.rs",
+        "bytes.rs",
+        "char.rs",
+        "clone.rs",
+        "enums.rs",
+        "error.rs",
+        "gobject/mod.rs",
+        "log.rs",
+        "main_context.rs",
```
sdroege, line 22: This one should be possible to fill in automatically, mostly.

build.rs
kinnison: Is it safe to call embed twice I wonder?

build.rs
kinnison: I imagine several of these can be cleared up easily, but these were all the ones where purge lost docs for them which needed more consideration than I had brainpower for at the time.

build.rs
sdroege: For different libraries and paths, probably? One would be `src/auto` and the other `src/auto/gobject`

build.rs
sdroege: Fair enough :)

build.rs
kinnison: As far as I can tell it's all relative to `src/` but maybe @GuillaumeGomez can help here?

kinnison:
The failure here appears to be related to `regen-check` which I had assumed would only affect `.rs` files in the various `src/` trees but perhaps affects `build.rs` too?  I shall try and get to it later today.

build.rs
sdroege: @GuillaumeGomez ?

GuillaumeGomez: You gave it `src`, so it shouldn't run on `build.rs`...

kinnison: It's not `build.rs` we're interested in, it's the ability to deal with the fact we have *two* Gir sets here (gobject and glib) to merge docs from.

GuillaumeGomez: Oh. I'm not sure `rustdoc-stripper` allows to give multiple doc files at once...

sdroege: But can it be called twice with different docs? It would have to handle in this case code to be in `src/*.rs` and `src/auto/gobject/*.rs` instead of `src/*.rs` and `src/auto/*.rs`.

GuillaumeGomez: It goes through throuh the folder and sub-folders and reads all the `.rs` files. I could add an option to pass files instead of a folder I guess. Would it work out for you?

kinnison: Honestly I cargo-culted this into glib, so you're best placed to work out the best approach.  I'm happy to drop this PR, or to let you take it over too.

sdroege:
https://github.com/gtk-rs/gio/pull/297 is also relevant here

sdroege:
Please rebase and create a new PR for https://github.com/gtk-rs/gtk-rs . Thanks!

PR #638, opened 2020-05-04
Only derive SetValueOptional for nullable GBoxed types

sdroege:
CC @gdesmott @GuillaumeGomez @EPashkin


EPashkin:
@sdroege Thanks, looks good.

PR #643, opened 2020-05-20
Add qdata API to ObjectExt

andy128k:


src/object.rs
```diff
@@ -1291,6 +1293,10 @@ pub trait ObjectExt: ObjectType {
     fn find_property<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<::ParamSpec>;
     fn list_properties(&self) -> Vec<::ParamSpec>;
 
+    fn set_qdata(&self, key: Quark, value: Box<dyn Any>);
+    fn get_qdata(&self, key: Quark) -> Option<&dyn Any>;
+    fn steal_qdata(&self, key: Quark) -> Option<Box<dyn Any>>;
```
sdroege, line 1298: This has to be `unsafe` unfortunately as other code might store arbitrary pointers at the same quark and then everything would fall apart. That's the main reason why I didn't add this API yet :)

src/object.rs
```diff
@@ -1291,6 +1293,10 @@ pub trait ObjectExt: ObjectType {
     fn find_property<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<::ParamSpec>;
     fn list_properties(&self) -> Vec<::ParamSpec>;
 
+    fn set_qdata(&self, key: Quark, value: Box<dyn Any>);
```
sdroege, line 1296: Instead of just `Any` this needs to be `Any + Send + Sync` as a general `glib::Object` can be used from arbitrary threads

src/object.rs
```diff
@@ -1490,6 +1496,52 @@ impl<T: ObjectType> ObjectExt for T {
         }
     }
 
+    fn set_qdata(&self, key: Quark, value: Box<dyn Any>) {
+        extern "C" fn drop_value(ptr: glib_sys::gpointer) {
+            if !ptr.is_null() {
```
sdroege, line 1501: Why would it ever be `NULL`?

src/object.rs
```diff
@@ -1490,6 +1496,52 @@ impl<T: ObjectType> ObjectExt for T {
         }
     }
 
+    fn set_qdata(&self, key: Quark, value: Box<dyn Any>) {
+        extern "C" fn drop_value(ptr: glib_sys::gpointer) {
```
sdroege, line 1500:
```suggestion
        extern "C" unsafe fn drop_value(ptr: glib_sys::gpointer) {
```

src/object.rs
```diff
@@ -1490,6 +1496,52 @@ impl<T: ObjectType> ObjectExt for T {
...
+            if !ptr.is_null() {
+                let value: Box<dyn Any> = unsafe { Box::from_raw(ptr as *mut dyn Any) };
+                std::mem::drop(value)
```
sdroege, line 1503:
```suggestion
                drop(value)
```

src/object.rs
```diff
@@ -1490,6 +1496,52 @@ impl<T: ObjectType> ObjectExt for T {
...
+        }
+    }
+
+    fn get_qdata(&self, key: Quark) -> Option<&dyn Any> {
+        unsafe {
+            let ptr = gobject_sys::g_object_get_qdata(
+                self.as_object_ref().to_glib_none().0,
+                key.to_glib(),
+            );
+            (ptr as *const Box<dyn Any>).as_ref().map(|b| &**b)
```
sdroege, line 1524: Where does the `Option` come from here? An explicit `NULL` check would probably be good for code readability.

src/object.rs
andy128k: Theoretically it wouldn't. Should I remove it?

sdroege: You could add an assertion :)

src/object.rs
andy128k: Done

src/object.rs
andy128k: Done

src/object.rs
andy128k: Done

src/object.rs
andy128k: Done

src/object.rs
andy128k: Done

src/object.rs
andy128k: Marking these functions as `unsafe` makes this PR pointless. :(
In this case I prefer even [less safe API](https://github.com/gtk-rs/glib/pull/644) but at least it is zero-cost (this one has two nested `Box`es).

sdroege: Indeed, let's go with that then :) Sorry for the trouble

PR #644, opened 2020-05-20
Add unsafe generic qdata API to ObjectExt

andy128k:


src/object.rs
```diff
@@ -1291,6 +1292,21 @@ pub trait ObjectExt: ObjectType {
     fn find_property<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<::ParamSpec>;
     fn list_properties(&self) -> Vec<::ParamSpec>;
 
+    /// # Safety
+    ///
+    /// This function doesn't store type information
+    unsafe fn set_qdata<QD: 'static + Send + Sync>(&self, key: Quark, value: QD);
```
sdroege, line 1298: If it's unsafe you can also remove the `Send` and `Sync` bounds :) For all GTK widgets it would be correct without, for example.

andy128k: Removed

sdroege:
Looks good to me, I think this is the most convenient thing we could do with qdata. You might also want to add `g_object_get_data()` and friends though (just a convenient function that doesn't require the string->quark conversion). Let me know if you want to add those too or prefer for this to be merged as is.

@EPashkin @GuillaumeGomez What do you think?

andy128k:
I am not sure **how** I want `{get,set,steal}_data` to be implemented.
They could be similar to `{get,set,steal}_qdata` and call `g_object_{get,set,steal}_data` or just be pure Rust like

```rust
unsafe fn get_data(&self, key: &str) -> Option<&QD> {
   self.get_qdata(Quark::from_stri(key))
}
```

sdroege:
I would go via Rust, seems simpler :)

andy128k:
`{get,set,steal}_data` are added also

andy128k:
I found this approach much easier than subclassing. I already use it in a few of my apps.

Together with custom [derive-macro](https://github.com/andy128k/find_duplicates/blob/master/find_duplicates_gui/src/main_window.rs#L220-L221) which generates `glib::clone::upgrade/downgrade`.

```rust
#[derive(newtype_gobject::NewTypeGObject)]
pub struct MainWindow(pub gtk::ApplicationWindow);
```

sdroege:
I see. Subclassing should also only really be used if necessary because virtual methods need to be overridden or similar things. I also don't use it otherwise :) But I never had the need for using qdata, usually I had enough information everywhere to get from the Rust struct to the widget or back as needed.

EPashkin:
:+1:

sdroege:
Only failures on CI are the known gtk4 related ones. Thanks!

PR #645, opened 2020-05-20
Disable GTK4 from the Travis CI

sdroege:
It's effectively unmaintained at this point and bitrotten. There's no
point in testing it in the GLib CI until that is fixed, it fails to
build currently.

CC @EPashkin @GuillaumeGomez


EPashkin:
@sdroege Thanks

PR #646, opened 2020-05-22
Various improvements to property handling

sdroege:
Also improve errors messages everywhere.

CC @EPashkin @GuillaumeGomez

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -1240,7 +1240,29 @@ impl Object {
         use std::ffi::CString;
 
         if !type_.is_a(&Object::static_type()) {
-            return Err(glib_bool_error!("Can't instantiate non-GObject objects"));
+            return Err(glib_bool_error!(
+                "Can't instantiate non-GObject type '{}'",
+                type_
+            ));
+        }
+
+        unsafe {
+            if gobject_sys::g_type_test_flags(
+                type_.to_glib(),
+                gobject_sys::G_TYPE_FLAG_INSTANTIATABLE,
+            ) == glib_sys::GFALSE
+            {
+                return Err(glib_bool_error!("Can't instantiate type '{}'", type_));
+            }
+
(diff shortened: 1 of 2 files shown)


src/object.rs
```diff
@@ -1949,6 +2034,88 @@ impl<T: ObjectType> ObjectExt for T {
     }
 }
 
+// Validate that the given property value has an acceptable type for the given property pspec
+// and if necessary update the value
+fn validate_property_type(
```
EPashkin, line 2039: validate_and_set_property ? Then why not just set_property_value?
`and if necessary update the value` -> `if acceptable` ?

EPashkin: Or comment is `if necessary update the value type` ?

sdroege: I don't understand the comment. This is only validating, not setting the property. Because the same function is a) also used to validate in `Object::new()` (where we can't set the property ourselves because there is no object yet), and b) in `ObjectExt::set_properties()` where we first want to validate all properties before setting one of them.

What exactly do you want changed here? :)

EPashkin: Seems I want only update commentary, that it updated only type of value, not value itself

EPashkin: and doing it almost always, without "necessary"

EPashkin: Sorry, it changes really only if "necessary"

sdroege: It's rarely updating the type. Only if the given type is a subclass of the required type instead of the type itself, and if `g_param_value_validate()` is doing changes, which also shouldn't happen often.

Can you suggest a wording that makes it more clear for you?

EPashkin: "if necessary update the type of value" or something like this

sdroege: `g_param_value_validate()` can also change the value of the value

EPashkin: Oh, this is bad :( Then current comment is good.

EPashkin:
Thanks 👍

PR #648, opened 2020-05-29
Don't require users of the glib proc macros to have gobject_sys in scope

sdroege:
CC @gdesmott @GuillaumeGomez @EPashkin @MathieuDuponchelle


EPashkin:
It not works even on nightly :(
Apveyour stable errors:
```
error[E0425]: cannot find value `crate_ident` in this scope
  --> glib-macros\src\genum_derive.rs:70:14
   |
70 |             #crate_ident::gobject_sys::GEnumValue {
   |              ^^^^^^^^^^^ not found in this scope
```

sdroege:
Indeed, fixed :)

EPashkin:
@sdroege Thanks, 👍

PR #650, opened 2020-06-05
Allow container types in Variant

zeenix:
This PR allows generic container types to be used with `Variant` API w/o breaking the API.

### DIFF
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -111,8 +111,8 @@ impl Variant {
 
     /// Returns `true` if the type of the value corresponds to `T`.
     #[inline]
-    pub fn is<T: StaticVariantType>(&self) -> bool {
-        self.type_() == T::static_variant_type()
+    pub fn is<T: DynamicVariantType>(&self) -> bool {
+        self.type_() == T::variant_type()
     }
 
     /// Tries to extract a value of type `T`.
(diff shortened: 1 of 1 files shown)


src/variant.rs
```diff
@@ -144,10 +145,10 @@ impl Variant {
     }
 
     /// Constructs a new serialised-mode GVariant instance.
-    pub fn new_from_bytes<T: StaticVariantType>(bytes: &Bytes) -> Self {
+    pub fn new_from_bytes<T: DynamicVariantType>(bytes: &Bytes) -> Self {
         unsafe {
             from_glib_none(glib_sys::g_variant_new_from_bytes(
-                T::static_variant_type().as_ptr() as *const _,
+                T::variant_type().as_ptr() as *const _,
```
sdroege, line 151: This would now everywhere cause needless allocations for simple types. Maybe that can be solved with some `Cow`? :)

src/variant.rs
```diff
@@ -378,6 +379,90 @@ impl<T: ToVariant> From<T> for Variant {
     }
 }
 
+/// Returns `VariantType` of `Self`.
+pub trait DynamicVariantType {
```
sdroege, line 383: Maybe `StaticVariantType` can be completely replaced by this, and instead of returning a `VariantType` you would return a `Cow<'static, VariantTy>`?

src/variant.rs
zeenix: not sure if that's possible w/o breaking the API but it's also that current API with lots of types is very confusing to me.

sdroege:
> w/o breaking the API.

Not breaking API is not a goal by itself :) If the API becomes weird otherwise let's rather break API. Here it seems like replacing the existing trait with something more generic would lead to better and less confusing API.

src/variant.rs
zeenix: Sure but then `VariantTy` also needs to be changed to container a ref rather.

sdroege: That's how it already works. There's `&VariantTy` (like `&str`) and `VariantType` (like `String`).

sdroege: I'd keep the name of the new trait as `StaticVariantType` btw, it's still the static signature of some type (it's not dependent on a value, it's fixed at compile time).

zeenix:
Closing this in favour of https://github.com/gtk-rs/glib/pull/651

PR #651, opened 2020-06-05
Add container Variant API

zeenix:
This PR will add container Variant API. It already adds basic array support.

### DIFF
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -65,6 +65,15 @@
 //! let map: HashMap<u16, String> = HashMap::from_variant(&variant).unwrap();
 //! assert_eq!(map[&1], "hi");
 //! assert_eq!(map[&2], "there");
+//!
+//! // And conversion to and from tuples.
+//! let variant = ("hello", 42u16, vec![ "there", "you" ],).to_variant();
+//! assert_eq!(variant.n_children(), 3);
+//! assert_eq!(variant.type_().to_str(), "(sqas)");
+//! let tuple = <(String, u16, Vec<String>)>::from_variant(&variant).unwrap();
+//! assert_eq!(tuple.0, "hello");
+//! assert_eq!(tuple.1, 42);
+//! assert_eq!(tuple.2, &[ "there", "you"]);
 //! ```
 
 use bytes::Bytes;
@@ -219,6 +228,16 @@ impl Variant {
         unsafe {
             from_glib_none(glib_sys::g_variant_new_array(
                 type_.as_ptr() as *const _,
+                children.to_glib_none().0,
+                children.len(),
+            ))
+        }
+    }
+
+    /// Creates a new GVariant tuple from children.
+    pub fn new_tuple(children: &[Variant]) -> Self {
+        unsafe {
+            from_glib_none(glib_sys::g_variant_new_tuple(
                 children.to_glib_none().0,
                 children.len(),
             ))
@@ -685,6 +704,36 @@ macro_rules! tuple_impls {
                     signature.push(')');
 
                     VariantType::new(&signature).expect("incorrect signature").into()
+                }
+            }
+
+            impl<$($name),+> FromVariant for ($($name,)+)
+            where
+                $($name: FromVariant,)+
+            {
+                fn from_variant(variant: &Variant) -> Option<Self> {
+                    Some((
+                        $(
+                            match $name::from_variant(&variant.get_child_value($n)) {
+                                Some(field) => field,
+                                None => return None,
+                            },
+                        )+
+                    ))
+                }
+            }
+
+            impl<$($name),+> ToVariant for ($($name,)+)
+            where
+                $($name: ToVariant,)+
+            {
+                fn to_variant(&self) -> Variant {
+                    let mut fields = Vec::with_capacity($len);
+                    $(
+                        let field = self.$n.to_variant();
+                        fields.push(field);
+                    )+
+                    Variant::new_tuple(&fields)
                 }
             }
         )+
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -19,6 +19,7 @@
 //! ```
 //! use glib::prelude::*; // or `use gtk::prelude::*;`
 //! use glib::{Variant, FromVariant, ToVariant};
+//! use std::collections::HashMap;
 //!
 //! // Using the `ToVariant` trait.
 //! let num = 10.to_variant();
@@ -54,6 +55,16 @@
 //! assert_eq!(variant.n_children(), 2);
 //! let vec = <Vec<String>>::from_variant(&array).unwrap();
 //! assert_eq!(vec[0], "Hello");
+//!
+//! // Conversion to and from HashMap is also possible
+//! let mut map: HashMap<u16, &str> = HashMap::new();
+//! map.insert(1, "hi");
+//! map.insert(2, "there");
+//! let variant = map.to_variant();
+//! assert_eq!(variant.n_children(), 2);
+//! let map: HashMap<u16, String> = HashMap::from_variant(&variant).unwrap();
+//! assert_eq!(map[&1], "hi");
+//! assert_eq!(map[&2], "there");
 //! ```
 
 use bytes::Bytes;
@@ -508,6 +519,48 @@ impl<T: StaticVariantType + ToVariant> ToVariant for Vec<T> {
 impl<T: StaticVariantType> StaticVariantType for Vec<T> {
     fn static_variant_type() -> Cow<'static, VariantTy> {
         <[T]>::static_variant_type()
+    }
+}
+
+impl<K, V, H> FromVariant for HashMap<K, V, H>
+where
+    K: FromVariant + Eq + Hash,
+    V: FromVariant,
+    H: BuildHasher + Default,
+{
+    fn from_variant(variant: &Variant) -> Option<Self> {
+        let mut map = HashMap::default();
+
+        for i in 0..variant.n_children() {
+            let entry = variant.get_child_value(i);
+            let key = match entry.get_child_value(0).get() {
+                Some(key) => key,
+                None => return None,
+            };
+            let val = match entry.get_child_value(1).get() {
+                Some(val) => val,
+                None => return None,
+            };
+
+            map.insert(key, val);
+        }
+
+        Some(map)
+    }
+}
+
+impl<K, V> ToVariant for HashMap<K, V>
+where
+    K: StaticVariantType + ToVariant + Eq + Hash,
+    V: StaticVariantType + ToVariant,
+{
+    fn to_variant(&self) -> Variant {
+        let mut vec = Vec::with_capacity(self.len());
+        for (key, value) in self {
+            let entry = DictEntry::new(key, value).to_variant();
+            vec.push(entry);
+        }
+        Variant::new_array::<DictEntry<K, V>>(&vec)
     }
 }
 
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -142,6 +142,25 @@ impl Variant {
         unsafe { from_glib_none(glib_sys::g_variant_get_variant(self.to_glib_none().0)) }
     }
 
+    /// Reads a child item out of a container `Variant` instance.
+    ///
+    /// # Panics
+    ///
+    /// * if `self` is not a container type.
+    /// * if given `index` is larger than number of children.
+    pub fn get_child_value(&self, index: usize) -> Variant {
+        assert!(index < self.n_children());
+        let ty = self.type_().to_str();
+        assert!(ty.starts_with("a") || ty.starts_with("{") || ty.starts_with("("));
+
+        unsafe {
+            from_glib_none(glib_sys::g_variant_get_child_value(
+                self.to_glib_none().0,
+                index,
+            ))
+        }
+    }
+
     /// Tries to extract a `&str`.
     ///
     /// Returns `Some` if the variant has a string type (`s`, `o` or `g` type
@@ -197,6 +216,20 @@ impl Variant {
     /// Returns the serialised form of a GVariant instance.
     pub fn get_data_as_bytes(&self) -> Bytes {
         unsafe { from_glib_full(glib_sys::g_variant_get_data_as_bytes(self.to_glib_none().0)) }
+    }
+
+    /// Determines the number of children in a container GVariant instance.
+    pub fn n_children(&self) -> usize {
+        let type_ = self.type_().to_str();
+        assert!(
+            type_.starts_with("a")
+                || type_.starts_with("m")
+                || type_.starts_with("(")
+                || type_.starts_with("{")
+                || type_.starts_with("v")
+        );
+
+        unsafe { glib_sys::g_variant_n_children(self.to_glib_none().0) }
     }
 }
 
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -36,6 +36,11 @@
 //! // `get_str` tries to borrow a string slice.
 //! assert_eq!(hello.get_str(), Some("Hello!"));
 //! assert_eq!(num.get_str(), None);
+//!
+//! // Variant carrying a Variant
+//! let variant = Variant::new_variant(&hello);
+//! let variant = variant.get_variant().unwrap();
+//! assert_eq!(variant.get_str(), Some("Hello!"));
 //! ```
 
 use bytes::Bytes;
@@ -121,6 +126,20 @@ impl Variant {
     #[inline]
     pub fn get<T: FromVariant>(&self) -> Option<T> {
         T::from_variant(self)
+    }
+
+    /// Boxes value.
+    #[inline]
+    pub fn new_variant(value: &Variant) -> Self {
+        unsafe { from_glib_full(glib_sys::g_variant_new_variant(value.to_glib_full())) }
+    }
+
+    /// Unboxes self.
+    ///
+    /// Returns `Some` if self contains a `Variant`.
+    #[inline]
+    pub fn get_variant(&self) -> Option<Variant> {
+        unsafe { from_glib_none(glib_sys::g_variant_get_variant(self.to_glib_none().0)) }
     }
 
     /// Tries to extract a `&str`.
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -287,6 +287,12 @@ pub trait StaticVariantType {
     fn static_variant_type() -> Cow<'static, VariantTy>;
 }
 
+impl StaticVariantType for Variant {
+    fn static_variant_type() -> Cow<'static, VariantTy> {
+        unsafe { VariantTy::from_str_unchecked("v").into() }
+    }
+}
+
 impl<'a, T: ?Sized + ToVariant> ToVariant for &'a T {
     fn to_variant(&self) -> Variant {
         <T as ToVariant>::to_variant(self)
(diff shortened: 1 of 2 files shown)


zeenix:
@sdroege Marked as WIP as I'm sure I'll be adding more API soon. But have a look already if you like the direction or not.

src/variant.rs
```diff
@@ -36,6 +36,18 @@
 //! // `get_str` tries to borrow a string slice.
 //! assert_eq!(hello.get_str(), Some("Hello!"));
 //! assert_eq!(num.get_str(), None);
+//!
+//! // Variant carying a Variant
+//! let variant = Variant::new_variant(&hello);
+//! let variant = variant.get_variant().unwrap();
+//! assert_eq!(variant.get_str(), Some("Hello!"));
+//!
+//! // Variant carying an array
```
sdroege, line 45:
```suggestion
//! // Variant carrying an array
```

src/variant.rs
```diff
@@ -143,6 +190,18 @@ impl Variant {
         }
     }
 
+    /// Creates a new GVariant array from children.
+    pub fn new_array<T: StaticVariantType>(children: &[Variant]) -> Self {
+        let children = children.as_ref();
```
sdroege, line 195: Why `as_ref()`?

src/variant.rs
```diff
@@ -378,6 +448,78 @@ impl<T: ToVariant> From<T> for Variant {
...
+
+impl<T: StaticVariantType> StaticVariantType for Vec<T> {
+    fn static_variant_type() -> Cow<'static, VariantTy> {
+        <[T]>::static_variant_type()
+    }
+}
+
+macro_rules! map_impls {
+    ($name:ident) => {
+        impl<K: StaticVariantType, V: StaticVariantType> StaticVariantType for std::collections::$name<K, V> {
```
sdroege, line 468: `HashSet` doesn't have a `V`, that's the random state :)

src/variant.rs
```diff
@@ -267,6 +331,12 @@ pub trait StaticVariantType {
     fn static_variant_type() -> Cow<'static, VariantTy>;
```
sdroege, line 331: Oh it already returned a `Cow` before, completely missed that :)

sdroege:
Goes into the right direction IMHO, thanks :)

sdroege:
So only thing missing here is the `FromVariant` / `ToVariant` impls for Vec/slice/tuples/hashmap or also something else?

zeenix:
> So only thing missing here is the `FromVariant` / `ToVariant` impls for Vec/slice/tuples/hashmap or also something else?

Yeah, I'll try to add those but for my own needs, the only thing that I know is missing is `g_variant_new_dict_entry` wrapper.

sdroege:
> Yeah, I'll try to add those

Well without these, the new trait impls are not really useful yet :)

zeenix:
> Well without these, the new trait impls are not really useful yet :)

They are, you can create the Variant from bytes or manually create it (like the array example in my patches). But yeah, w/o those conversions, the API will still be very incomplete.

sdroege:
Any updates here?

zeenix:
> Any updates here?

no. :( If you want to merge what we already have here, let's unmark it as WIP and merge. I can always add additional PRs.

sdroege:
Without the remaining trait impls it's not really useful yet, so let's just wait :) Too late for this release now anyway, it happened already.

zeenix:
> Without the remaining trait impls it's not really useful yet, so let's just wait :) Too late for this release now anyway, it happened already.

Sure. :)

sdroege:
Any updates here?

zeenix:
> Any updates here?

No, for me it's not an isolated effort but rather something I've been doing as part of the gvariant addition in zvariant (so it gets some real testing) but there has been more important zbus bugs/issues that I've been looking into, in my limited sparetime before I could continue finishing this work.

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+        VariantType::new(&signature).expect("incorrect signature").into()
+    }
+}
+
+impl<T: FromVariant> FromVariant for Vec<T> {
+    fn from_variant(variant: &Variant) -> Option<Self> {
+        let mut vec = vec![];
+
+        let n_children = variant.n_children();
+        if n_children != 0 {
```
sdroege, line 492: This `if` here seems redundant

src/variant.rs
```diff
@@ -174,10 +250,25 @@ impl Variant {
         ))
     }
 
+    /// Creates a new dictionary entry GVariant.
+    pub fn new_dict_entry(key: &Variant, value: &Variant) -> Self {
```
sdroege, line 254: What would one do with the return value, how would it be inserted into a dict? Might be worth adding a short doc example :)

src/variant.rs
```diff
@@ -143,6 +208,17 @@ impl Variant {
         }
     }
 
+    /// Creates a new GVariant array from children.
+    pub fn new_array<T: StaticVariantType>(children: &[Variant]) -> Self {
```
sdroege, line 212: Some kind of `append_to_array()` API could also be useful

sdroege: That would make `impl ToVariant for Vec<T>` a bit less wasteful (no need for an intermediate vec)

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+where
+    K: FromVariant + Eq + Hash,
+    V: FromVariant,
+    H: BuildHasher + Default,
+{
+    fn from_variant(variant: &Variant) -> Option<Self> {
+        let mut map = HashMap::default();
+
+        let n_children = variant.n_children();
+        if n_children != 0 {
```
sdroege, line 531: Redundant `if`

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+                &key.to_variant(),
+                &value.to_variant(),
+            );
+            vec.push(entry);
+        }
+        Variant::new_array::<DictEntry<K, V>>(&vec)
+    }
+}
+
+struct DictEntry<K, V>(K, V);
```
sdroege, line 569: Should this be public?

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+    7 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6)
+    8 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7)
+    9 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8)
+    10 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9)
+    11 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10)
+    12 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10 11 T11)
+    13 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10 11 T11 12 T12)
+    14 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10 11 T11 12 T12 13 T13)
+    15 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10 11 T11 12 T12 13 T13 14 T14)
+    16 => (0 T0 1 T1 2 T2 3 T3 4 T4 5 T5 6 T6 7 T7 8 T8 9 T9 10 T10 11 T11 12 T12 13 T13 14 T14 15 T15)
```
sdroege, line 629: Tests for this would be useful :)

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+    }
+}
+
+impl<T: StaticVariantType> StaticVariantType for Vec<T> {
+    fn static_variant_type() -> Cow<'static, VariantTy> {
+        <[T]>::static_variant_type()
+    }
+}
+
+impl<K, V, H> FromVariant for HashMap<K, V, H>
```
sdroege, line 521: Tests for this would be useful :)

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+impl<T: StaticVariantType> StaticVariantType for [T] {
+    fn static_variant_type() -> Cow<'static, VariantTy> {
+        let child_type = T::static_variant_type();
+        let signature = format!("a{}", child_type.to_str());
+
+        VariantType::new(&signature).expect("incorrect signature").into()
+    }
+}
+
+impl<T: FromVariant> FromVariant for Vec<T> {
```
sdroege, line 487: Tests for this would be useful :)

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+    }
+}
+
+impl<K, V> ToVariant for HashMap<K, V>
+where
+    K: StaticVariantType + ToVariant + Eq + Hash,
+    V: StaticVariantType + ToVariant,
+{
+    fn to_variant(&self) -> Variant {
+        let mut vec = vec![];
```
sdroege, line 557: See above for append API or at least allocating with right capacity

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+        let signature = format!("a{{{}{}}}", key_type.to_str(), value_type.to_str());
+
+        VariantType::new(&signature).expect("incorrect signature").into()
+    }
+}
+
+macro_rules! tuple_impls {
+    ($($len:expr => ($($n:tt $name:ident)+))+) => {
+        $(
+            impl<$($name),+> StaticVariantType for ($($name,)+)
```
sdroege, line 594: The from/to traits are missing here still

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+
+        VariantType::new(&signature).expect("incorrect signature").into()
+    }
+}
+
+impl<T: FromVariant> FromVariant for Vec<T> {
+    fn from_variant(variant: &Variant) -> Option<Self> {
+        let mut vec = vec![];
+
+        let n_children = variant.n_children();
```
sdroege, line 491: If there's no `append_to_array()` API possible, you could do `Vec::with_capacity(variant.n_children())` at least to prevent reallocations

src/variant.rs
```diff
@@ -378,10 +475,164 @@ impl<T: ToVariant> From<T> for Variant {
...
+            }
+        }
+
+        Some(vec)
+    }
+}
+
+impl<T: StaticVariantType + ToVariant> ToVariant for Vec<T> {
+    fn to_variant(&self) -> Variant {
+        let mut vec = vec![];
```
sdroege, line 507: Could pre-allocate the vec with the right capacity

sdroege (commented):
Looks like good progress :)

src/variant.rs
zeenix: A `Variant` is supposed to be non-mutable so there is no `_append` possible. If we want this, we need bindings for `VariantBuilder`. Hopefully basic bindinds for that are not hard.

src/variant.rs
zeenix: I added docs that run as tests.

src/variant.rs
sdroege: Ah I see. Something for another PR then unless you want to do it now :)

src/variant.rs
zeenix: ah right, `0..0` range isn't an issue in Rust.

src/variant.rs
zeenix: Hmm.. maybe yeah and `Variant::new_dict_entry` could take this as arg.

src/variant.rs
zeenix: I know. I didn't say I'm done. :)

src/variant.rs
sdroege: In which language is it?

src/variant.rs
zeenix: Another PR seems good. I want this to be done ASAP. :) Let's create an issue for that and assign it to me.

src/variant.rs
zeenix: I don't know actually. Forgot it, brain fart.. :)

sdroege: Ok, I was wondering what language I should try to avoid :)

src/variant.rs
zeenix: In "GVariant" format (unlike D-Bus), a dict-entry can be used on its own and doesn't need to be in a dict. I'm just provide a binding for the C API, which doesn't document it's use either. I'll add a bit of info. :)

src/variant.rs
zeenix: Created https://github.com/gtk-rs/glib/issues/676 for this. I think the PR is now good enough to be merged.

src/variant.rs
```diff
@@ -36,16 +37,54 @@
...
 //! assert_eq!(num.get_str(), None);
+//!
+//! // Variant carrying a Variant
+//! let variant = Variant::new_variant(&hello);
+//! let variant = variant.get_variant().unwrap();
+//! assert_eq!(variant.get_str(), Some("Hello!"));
+//!
+//! // Variant carrying an array
+//! let array = [Variant::from("Hello"), Variant::from("there!")];
+//! let variant = Variant::new_array::<&str>(&array);
```
sdroege, line 48: What happens if the array does not contain string variants (i.e. if the generic parameter to `new_array()` does not match what's in the array)? Shouldn't this return some kind of error then?

src/variant.rs
```diff
@@ -122,6 +162,40 @@ impl Variant {
...
+    }
+
+    /// Reads a child item out of a container `Variant` instance.
+    ///
+    /// # Panics
+    ///
+    /// * if `self` is not a container type.
+    /// * if given `index` is larger than number of children.
+    pub fn get_child_value(&self, index: usize) -> Variant {
+        assert!(index <= self.n_children());
```
sdroege, line 190:
```suggestion
        assert!(index < self.n_children());
```

Or not? It's zero-index I assume?

src/variant.rs
sdroege: The docs in C say

> All items in the array must have the same type, which must be the same as child_type , if given.

So we should probably make that an assertion on our side and document that

src/variant.rs
```diff
@@ -178,6 +273,11 @@ impl Variant {
     pub fn get_data_as_bytes(&self) -> Bytes {
         unsafe { from_glib_full(glib_sys::g_variant_get_data_as_bytes(self.to_glib_none().0)) }
     }
+
+    /// Determines the number of children in a container GVariant instance.
+    pub fn n_children(&self) -> usize {
+        unsafe { glib_sys::g_variant_n_children(self.to_glib_none().0) }
```
sdroege, line 279: What does this do on non-container variants? C docs say

> It is an error to call this function on any other type of GVariant.

So we should probably check that here?

src/variant.rs
```diff
@@ -36,16 +37,54 @@
...
+//! assert_eq!(map[&2], "there");
+//!
+//! // And conversion to and from tuples.
+//! let variant = ("hello", 42u16, vec![ "there", "you" ],).to_variant();
+//! assert_eq!(variant.n_children(), 3);
+//! assert_eq!(variant.type_().to_str(), "(sqas)");
+//! let tuple = <(String, u16, Vec<String>)>::from_variant(&variant).unwrap();
+//! assert_eq!(tuple.0, "hello");
+//! assert_eq!(tuple.1, 42);
+//! assert_eq!(tuple.2, &[ "there", "you"]);
```
sdroege, line 76: Another trivial container that would be useful to add here would be maybes (`m`)

src/variant.rs
```diff
@@ -378,10 +484,259 @@ impl<T: ToVariant> From<T> for Variant {
...
+            ))
+        }
+    }
+}
+
+impl<K: StaticVariantType, V: StaticVariantType> StaticVariantType for DictEntry<K, V> {
+    fn static_variant_type() -> Cow<'static, VariantTy> {
+        let key_type = K::static_variant_type();
+        let value_type = V::static_variant_type();
+        let signature = format!("{{{}{}}}", key_type.to_str(), value_type.to_str());
```
sdroege, line 649: `"{{{}{}}}"` it's looking at us :rofl:

src/variant.rs
```diff
@@ -453,5 +808,14 @@ mod tests {
...
         assert!(!set.contains(&v3));
+
+        assert_eq!(<HashMap<&str,(&str,u8,u32)>>::static_variant_type().to_str(), "a{s(syu)}");
+    }
+
+    #[test]
+    fn test_array() {
+        // Test just the signature for now.
+        assert_eq!(<Vec<&str>>::static_variant_type().to_str(), "as");
+        assert_eq!(<Vec<(&str,u8,u32)>>::static_variant_type().to_str(), "a(syu)");
```
sdroege, line 819: Are these tests not covered by the doctests too?

sdroege (commented):
Looks good to me otherwise, only trivial stuff left :)

src/variant.rs
zeenix: ah good point. I will need to add that on zvariant side too and will need it here for testing the serializer against.

src/variant.rs
sdroege: I assume it's a `g_return_val_if_fail()` in C that causes a) 0 to be returned and not showing the caller that things are wrong, and b) generally `g_return_val_if_fail()` and similar guards are leaving everything behind in an inconsistent state.

If it was a proper assertion then a panic would still be better as that can at least be caught at isolation boundaries if needed.

src/variant.rs
zeenix: It wasn't super "trivial" in the end for me but I added good bindings after all. :)

sdroege: I expected it to be basically the same as for arrays/vecs, just with a single item / `Option` :)

sdroege:
Now just needs to merge cleanly :) Can you remove those "Revert Revert" commits and the gir update commit?

zeenix:
> Now just needs to merge cleanly :) Can you remove those "Revert Revert" commits and the gir update commit?

Rebased. Although I failed to test the results because:

```
$ cargo test
error: failed to select a version for `glib-sys`.
    ... required by package `glib v0.10.0 (/home/zeenix/checkout/gnome/glib-rs)`
versions that meet the requirements `=0.9.1` are: 0.9.1

the package `glib` depends on `glib-sys`, with features: `v2_64` but `glib-sys` does not have these features.


failed to select a version for `glib-sys` which could resolve this conflict

```

Cargo clean didn't help, neither did removal of selected folders under `~.cargo`.

zeenix:
> Rebased. Although I failed to test the results because:

Nm, needed `cargo update`. All tests pass still.

sdroege:
Needs a run of `cargo fmt`, seems otherwise the CI is happy :)

PR #652, opened 2020-06-06
Ignore g_variant_get_gtype()

sdroege:
It should be used by a StaticType impl, which we have, and not be
available as a standalone function.

----

CC @EPashkin @GuillaumeGomez


EPashkin:
👍

PR #655, opened 2020-06-11
Revert "Revert "Update to focal gir""

GuillaumeGomez:
Reverts gtk-rs/glib#654

This is becoming so funny XD


sdroege:
Good to go after sys is in

PR #657, opened 2020-07-01
Use g_type_register_static_simple() instead of g_type_register_static…

sdroege:
…() for simplicity

----

CC @GuillaumeGomez @EPashkin

### DIFF
--- a/src/subclass/interface.rs
+++ b/src/subclass/interface.rs
@@ -8,7 +8,6 @@ use gobject_sys;
 use std::borrow::Borrow;
 use std::marker;
 use std::mem;
-use std::ptr;
 use translate::*;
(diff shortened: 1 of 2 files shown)


sdroege:
@GuillaumeGomez The clone tests are failing somehow, which is unrelated to my changes: https://travis-ci.org/github/gtk-rs/glib/jobs/703869975

Any guesses?

GuillaumeGomez:
Let's check once if it fails again. But I wonder if the problem doesn't come from the compiler error output...

sdroege:
> Let's check once if it fails again. But I wonder if the problem doesn't come from the compiler error output...

Oh maybe the compiler warning broke it

sdroege:
That was it apparently. Now just the mac runner exploded :man_shrugging:

PR #662, opened 2020-07-05
Keep the scope of clone! macro arms in the crate

aledomu:
When doing recursive calls in the macro, it calls itself by its raw name. This is a problem when importing and renaming it. This fix makes that possible.

### DIFF
--- a/src/clone.rs
+++ b/src/clone.rs
@@ -457,7 +457,7 @@ macro_rules! clone {
         }
     );
     ($($(@ $strength:ident$(-$var:ident-$var2:ident)?)? $($variables:ident).+ $(as $rename:ident)?),+ => @default-panic, move || $body:expr ) => (
-        clone!($($(@ $strength$(-$var-$var2)?)? $($variables).+ $(as $rename)?),* => @default-panic, move || { $body })
+        $crate::clone!($($(@ $strength$(-$var-$var2)?)? $($variables).+ $(as $rename)?),* => @default-panic, move || { $body })
     );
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Nice catch! Could you add a test to enforce it please?

aledomu:
> Nice catch! Could you add a test to enforce it please?

Done. I tested that without the modification it fails as expected.

sdroege:
Looks good to me but needs indentation to be fixed up: https://travis-ci.org/github/gtk-rs/glib/jobs/705543841

aledomu:
I think it'd be better if there was a pre-commit hook for preventing contributors from sending unvalidated code like I did, but I don't know how the checks are setup in the CI/CD.

sdroege:
Indeed, I also dislike the current way of code style checking where you only get the results long after having submitted the changes. I don't know how to do that better with Travis/GitHub though (with GitLab that's relatively easy), so if you have any suggestions please let us know :)

Note the pre-commit hooks are not part of the repository but would have to be manually installed into each clone.

For the changes themselves, looks good to me now. Thanks a lot!

aledomu:
Isn't a standard local git hook enough?

sdroege:
> Isn't a standard local git hook enough?

Everybody would have to add that into their own clone manually then. Is that more likely to happen than people running `cargo fmt` manually?

aledomu:
If people try to commit their changes without doing a pass with `cargo fmt` and `cargo clippy`, the commit will fail. That forces people to run the check. That's how we do it in Fractal and I think it does the job. If you think so, I'll leave you another PR with those changes, I only need to know what are the commands that you are running in the CI/CD for doing the format and lint checks.

sdroege:
How do you add the pre-commit hook into the local git clones of everybody?

aledomu:
See #663

PR #663, opened 2020-07-07
Add pre-commit hook for format checks

aledomu:
It lacks lint checks. I need to know what commands you are running for this.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+if ! which rustup &> /dev/null; then
+    curl https://sh.rustup.rs -sSf  | sh -s -- -y
+    export PATH=$PATH:$HOME/.cargo/bin
+    if ! which rustup &> /dev/null; then
+        echo "Failed to install rustup"
+    fi
+fi
+
+if ! rustup component list|grep rustfmt &> /dev/null; then
+    echo "Installing rustfmt.."
+    rustup component add rustfmt
+fi
+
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Sorry but I'm strongly against this. :-/ We're about to switch to github actions and there it'll be very quick to get errors for formatting. I don't want anything to be done on the user's side if possible.

aledomu:
Just curious, when forking the repo as a contributor, will actions be imported? Or will actions only work when doing the PR as it does right now?

hooks/pre-commit
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
sdroege, line 1: This won't work on Windows

hooks/pre-commit
```diff
@@ -0,0 +1,25 @@
...
+
+if ! which rustup &> /dev/null; then
+    curl https://sh.rustup.rs -sSf  | sh -s -- -y
```
sdroege, line 4: Not everybody uses rustup :)

hooks/pre-commit
aledomu: :facepalm:

aledomu:
It seems clear this isn't a good idea at all. Closing

sdroege:
> It seems clear this isn't a good idea at all. Closing

It's a good idea to have something like this but there's a reason why I asked how you want to do that in the issue :) All ways of doing this that I could think of are going to break in various situations.

PR #665, opened 2020-07-09
glib-macros: symlink the parent LICENSE

cuviper:
A symlink to `../LICENSE` is usually enough to ensure the license is
included in the package of published sources.


sdroege:
Oops, thanks for noticing! @GuillaumeGomez can you do a little bugfix release with this?

sdroege:
> @GuillaumeGomez can you do a little bugfix release with this?

Of glib-macros only, not glib too

cuviper:
Thank you both!

PR #666, opened 2020-07-21
Clone to proc macro

GuillaumeGomez:
I ported the clone macro to proc-macro, which allows to get rid of the very difficult to read/maintain code we had.

cc @sdroege @EPashkin


sdroege:
PR number 666, very ominous 😱

I'll review this next week, thanks for doing this. Do you like the code better now?

EPashkin:
Not sure error in example or macro:
```
error: missing `in` in `for` loop
   --> src/bin/progress_tracker.rs:62:13
    |
62  | /             clone!(@weak self.widgets as widgets => move |_| {
63  | |                 if active.get() {
64  | |                     return;
65  | |                 }
...   |
108 | |                 });
109 | |             }),
    | |______________^ help: try adding `in` here
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0425]: cannot find value `in1` in this scope
   --> src/bin/progress_tracker.rs:62:13
    |
62  | /             clone!(@weak self.widgets as widgets => move |_| {
63  | |                 if active.get() {
64  | |                     return;
65  | |                 }
...   |
108 | |                 });
109 | |             }),
    | |______________^ not found in this scope
    |
```

GuillaumeGomez:
In generated code very likely. I'll need to add some more tests. :)

> PR number 666, very ominous 😱

😆 

> I'll review this next week, thanks for doing this. Do you like the code better now?

Now it's pretty straight-forward and it's much easier to follow what's going on. So I definitely prefer it this way.

GuillaumeGomez:
I used this occasion to extend the macro a bit more: you can now have blocks as value to `@default-return`.

GuillaumeGomez:
I think travis is completely out of date (I updated the minimum version to 1.41 but it doesn't care...) and the stable version is working on appveyor but not on travis. Maybe it's time to move this CI too?

glib-macros/src/clone.rs
```diff
@@ -0,0 +1,562 @@
...
+        match (self.borrow_kind, wrapper_kind) {
+            (BorrowKind::Weak, WrapperKind::DefaultPanic) => {
+                let name = if let Some(ref a) = self.alias {
+                    a
+                } else {
+                    &self.name
+                };
+                format!(
+                    "\
+let {0} = match {1}::clone::Upgrade::upgrade(&{0}) {{
```
sdroege, line 90: We use syn and quote in this crate elsewhere, can you check if using them makes things simpler / less verbose here? Also CC @gdesmott, who probably has an opinion on that.

glib-macros/src/clone.rs
```diff
@@ -0,0 +1,562 @@
...
+        },
+        Some(TokenTree::Group(g)) => return group_to_string(&g),
+        None => panic!("Unexpected end after `@default-return`"),
+    };
+    loop {
+        match parts.peek() {
+            Some(TokenTree::Punct(p)) => {
+                let p_s = p.to_string();
+                if p_s == "{" || p_s == "(" || p_s == "[" {
+                    total += 1;
```
sdroege, line 361: For example this kind of parsing should already be taken care of by syn. It parses the token tree into a proper Rust AST. I worry that if we parse Rust code ourselves here that we end up missing various special cases and have problems maintaining this over time (e.g. if the language is extended).

glib-macros/src/clone.rs
gdesmott: I'm far from being a macro expert but `quote` and `syn` makes things much easier and readable from my (small) experience. I'd advise to keep using them.

sdroege: Using syn makes no sense here as the macro does not really contain Rust but only something based on Rust tokens, and the actual closure block we take as-is.

quote might make sense and I'll look at that next week, together with improving the parsing a bit. But for now this is all ok, I don't think there are bugs :)

sdroege:
@GuillaumeGomez I would merge now but what's the CI failure about?

```
error[E0658]: procedural macros cannot be expanded to expressions
```

That only works with nightly and beta it seems, not even with 1.45. So probably have to wait until 1.46?

GuillaumeGomez:
Yes, let's wait until stable version has it as well. So in four weeks.

GuillaumeGomez:
CI kinda passed.

sdroege:
> [Update minimum rust version to 1.46](/gtk-rs/glib/pull/666/commits/9d276485ffab57b0ee40fc36691cfa047e826d57)

Do we want that at this point?

GuillaumeGomez:
Is it an issue? We need to upgrade the minimum rust version for the next release in any case. Or maybe I missed something?

sdroege:
> Is it an issue?

Maybe for some people it is :) The version required by Firefox is 1.41.1 currently and that's what the enterprise Linux distros are shipping.

> We need to upgrade the minimum rust version for the next release in any case. Or maybe I missed something?

Why do we need to, other than this issue? Are there other things we need to update for? If there are good reasons then let's go for it, I like using newer APIs ;)

sdroege:
Please rebase and create a new PR for https://github.com/gtk-rs/gtk-rs . Thanks!

Issue #668, opened 2020-07-23
Try to get rid of ObjectImpl::get_type_data and corresponding macro


sdroege:
If I'm not missing anything this is actually unneeded in the current design and ObjectSubclass::type_data could be used everywhere. 

Will check in a few days

PR #669, opened 2020-07-25
Get rid of ObjectImpl::get_type_data() and glib_object_impl!() macro

sdroege:
And as a side-effect clean up trait bounds in various places.

Fixes https://github.com/gtk-rs/glib/issues/668

----

CC @EPashkin @GuillaumeGomez


EPashkin:
@sdroege Thanks
👍 for this and followups

PR #673, opened 2020-08-12
doc: fix documentation for usage of Error::kind<T>() with match expression

takaswie:
Although glib::Error cannot be converted into Option::<T> directly,
its documentation describe it's possible. It seems to be typo for
Error::kind::<T>().


sdroege:
Thanks!

PR #522, opened 2019-09-07
Fully qualify inner macros for exported macros...

fengalin:
... otherwise they can't be resolved in dependent crates compiled
for edition 2018.

PR #675, opened 2020-08-28
Use $crate when invoking glib_object_wrapper! in a macro

ids1024:
This was done elsewhere in #522.


GuillaumeGomez (approved):

GuillaumeGomez:
Normally, gtk-rs is kinda supposed to support the 2015 edition "only", but I guess it's fine. Thanks for the PR!

sdroege:
This works with both editions though (or the CI would've failed already).

PR #677, opened 2020-09-03
fix: Do not transfer ownership in Variant::new_variant()

ids1024:
Trying to port some code to the new variant API, this was causing errors.

I believe this is correct now.


ids1024:
Travis failure appears to be spurious: `failed to resolve address for github.com: nodename nor servname provided, or not known; class=Net (12)`

sdroege:
CC @zeenix

sdroege:
Thanks, the CI failure is indeed completely unrelated :)

Issue #678, opened 2020-09-03
Implement ToVariant and FromVariant for Variant (not currently possible)


ids1024:
With the new methods and trait implementations provided in https://github.com/gtk-rs/glib/pull/651, it seems like it would be helpful if `Variant` implemented `ToVariant` and `FromVariant`.

As a motivating example, consider this snippet trying to set a property with gdbus:

```rust
properties_proxy.call_sync::<gio::Cancellable>(
    "Set",
    Some(&("some-interface", "some-property", Variant::new_variant(&value)).to_variant()),
    gio::DBusCallFlags::NONE,
    60000,
    None,
)?;
```

This almost works, except the fact `Variant::new_variant(&value)` returns a `Variant`, which doesn't implement `ToVariant`. It's possible by passing a vector of variants to `Variant::new_tuple`, but it would be nice it the somewhat more ergonomic API were available.

Unfortunately, implementing `ToVariant` results in a compiler error:

```rust
error[E0119]: conflicting implementations of trait `std::convert::From<variant::Variant>` for type `variant::Variant`:
   --> src/variant.rs:526:1
    |
526 | impl<T: ToVariant> From<T> for Variant {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T> std::convert::From<T> for T;
```

I can send a PR implementing `FromVariant`, since that doesn't pose an issue. But I'm not sure there's any way to support `ToVariant` on `Variant`, without new/unstable functionality in Rustc (specialization, negative trait bounds). Or some breaking changes to glib-rs which in some way would involve not having a  `impl<T: ToVariant> From<T> for Variant`  implementation.

ids1024:
While `FromVariant` allows things like this:

```rust
let ret: (HashMap<String, Variant>,) = proxy
    .call_sync::<gio::Cancellable>(
        "GetManagedObjects",
        None,
        gio::DBusCallFlags::NONE,
        60000,
        None,
    )
    .unwrap()
    .get()
    .unwrap();
```

Not sure if it would be strange for `Variant` to implement `FromVariant` but not `ToVariant`.

sdroege:
> Or some breaking changes to glib-rs which in some way would involve not having a `impl<T: ToVariant> From<T> for Variant` implementation.

That's what I would suggest. The `From` impl doesn't make much sense, that's why there's a separate trait already.

PR #679, opened 2020-09-03
Implement FromVariant for Variant

ids1024:
This is useful when combined with the container variant functions and trait implementations from https://github.com/gtk-rs/glib/pull/651.

Implementing `ToVariant` would be useful as well, but isn't currently possible: https://github.com/gtk-rs/glib/issues/678

### DIFF
--- a/src/variant.rs
+++ b/src/variant.rs
@@ -31,10 +31,8 @@
 //! assert_eq!(num.get::<i32>(), Some(10));
 //! assert_eq!(num.get::<u32>(), None);
 //!
-//! // `Variant` implements `From`
-//! let hello = Variant::from("Hello!");
-//!
 //! // `get_str` tries to borrow a string slice.
+//! let hello = "Hello!".to_variant();
 //! assert_eq!(hello.get_str(), Some("Hello!"));
 //! assert_eq!(num.get_str(), None);
 //!
@@ -44,7 +42,7 @@
 //! assert_eq!(variant.get_str(), Some("Hello!"));
 //!
 //! // Variant carrying an array
-//! let array = [Variant::from("Hello"), Variant::from("there!")];
+//! let array = ["Hello".to_variant(), "there!".to_variant()];
 //! let variant = Variant::array::<&str>(&array);
 //! assert_eq!(variant.n_children(), 2);
 //! assert_eq!(variant.get_child_value(0).get_str(), Some("Hello"));
@@ -529,12 +527,6 @@ impl ToVariant for str {
     }
 }
 
-impl<T: ToVariant> From<T> for Variant {
-    fn from(value: T) -> Variant {
-        value.to_variant()
-    }
-}
-
 impl<T: StaticVariantType> StaticVariantType for Option<T> {
     fn static_variant_type() -> Cow<'static, VariantTy> {
         let child_type = T::static_variant_type();
@@ -731,6 +723,18 @@ where
     }
 }
 
+impl ToVariant for Variant {
+    fn to_variant(&self) -> Variant {
+        Variant::variant(self)
+    }
+}
+
+impl FromVariant for Variant {
+    fn from_variant(variant: &Variant) -> Option<Self> {
+        variant.get_variant()
+    }
+}
+
 impl<K: StaticVariantType, V: StaticVariantType> StaticVariantType for DictEntry<K, V> {
     fn static_variant_type() -> Cow<'static, VariantTy> {
         let key_type = K::static_variant_type();
@@ -842,7 +846,7 @@ mod tests {
             fn $name() {
                 let mut n = $ty::max_value();
                 while n > 0 {
-                    let v = Variant::from(n);
+                    let v = n.to_variant();
                     assert_eq!(v.get(), Some(n));
                     n /= 2;
                 }
@@ -856,9 +860,9 @@ mod tests {
             fn $name() {
                 let mut n = $ty::max_value();
                 while n > 0 {
-                    let v = Variant::from(n);
+                    let v = n.to_variant();
                     assert_eq!(v.get(), Some(n));
-                    let v = Variant::from(-n);
+                    let v = (-n).to_variant();
                     assert_eq!(v.get(), Some(-n));
                     n /= 2;
                 }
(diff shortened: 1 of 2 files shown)


sdroege:
See https://github.com/gtk-rs/glib/issues/678#issuecomment-686967071 . Can you also add that here for symmetry? :)

src/variant.rs
```diff
@@ -725,6 +725,12 @@ where
     }
 }
 
+impl FromVariant for Variant {
+    fn from_variant(variant: &Variant) -> Option<Self> {
+        Some(variant.clone())
```
sdroege, line 730: Why this and not `Variant::new_variant()`? I would expect this to box the variant in another variant (and that the behaviour for `glib::Value` is currently the same as what you implement here is a bug that I'm going to fix in the near future).

ids1024: Hm. `new_variant()` creates a variant of type `v` instead of the original type. My thought was that it could be useful to use the abstractions in the new container API when one already has a variant of the type one wants, but needs to put it in a tuple or some such.

In any case, I don't think it's `new_variant()` that should be used here, but rather `get_variant()`. (`new_variant()` would be used in `ToVariant`.)

ids1024: And I do like how in the code I showed in https://github.com/gtk-rs/glib/issues/678#issuecomment-686813911, I could use this implementation of `FromVariant` to partially deconstruct a complicated `Variant` type.

I don't know if there's a clean way to support both.

sdroege: > And I do like how in the code I showed in #678 (comment), I could use this implementation of FromVariant to partially deconstruct a complicated Variant type.

While I agree that this is convenient, this seems more like a hack then a nice solution :) By having this implemented the API is asymmetric and the `Variant` wrapping is not possible to express anymore.

> In any case, I don't think it's new_variant() that should be used here, but rather get_variant(). (new_variant() would be used in ToVariant.)

Indeed, sorry

sdroege (changes requested):
See https://github.com/gtk-rs/glib/pull/679/files#r483437430

@zeenix @GuillaumeGomez Do you have any opinions?

myfreeweb:
Trying to use the current implementation to access D-Bus property change notifications with their `a{sv}`:

```rust
    dd.connect_local("g-properties-changed", true, move |args| {
        let new_props = args[1]
            .get::<glib::Variant>()
            .unwrap()
            .unwrap()
            .get::<std::collections::HashMap<String, glib::Variant>>()
            .unwrap();
        None
    })
```

It does work now:

```
{"TimeToEmpty": Variant { ptr: 0x1eaedcbfd580, type: VariantTy { inner: "v" }, value: "<int64 19200>" }, "UpdateTime": Variant { ptr: 0x1eaed6ac0490, type: VariantTy { inner: "v" }, value: "<uint64 1599419489>" }, "EnergyRate": Variant { ptr: 0x1eaed6a8c260, type: VariantTy { inner: "v" }, value: "<5.6056000000000008>" }, "Energy": Variant { ptr: 0x1eaedcbd5890, type: VariantTy { inner: "v" }, value: "<29.9222>" }}
```

But I get the aforementioned "boxed" variants and there doesn't seem to be a way to access the inner variant in these, since the instances just return the outer variant itself :/
i.e. I can infinitely chain `.get::<glib::Variant>().unwrap().get::<glib::Variant>().unwrap()` uselessly

zeenix:
> But I get the aforementioned "boxed" variants and there doesn't seem to be a way to access the inner variant in these, since the instances just return the outer variant itself :/

You need to use the `get_child_value()` method.

myfreeweb:
Oh. Thanks. idk why I keep staring at docs.rs for the release version expecting it to show something new from a PR.. :D

myfreeweb:
hmm..

```
thread 'main' panicked at 'assertion failed: ty.starts_with("a") || ty.starts_with("{") || ty.starts_with("(")', /home/greg/.local/share/cargo/git/checkouts/glib-53370c4288e47ffa/633b61a/src/variant.rs:197:9
```

myfreeweb:
(using `get_child_value` to walk the `a{sv}` also gets me to these same variants with `v` inside)

src/variant.rs
ids1024: I've rebased on master and changed to using `.variant()` and `.get_variant()`.

For my immediate use case, I think the [`.iter()` method](https://github.com/gtk-rs/glib/pull/683) is probably a better way to go through a collection variant without fully deconstructing it, so I'm mostly happy with this implementation, supposing the gtk-rs maintainers all agree it should work this way.

src/variant.rs
```diff
@@ -529,12 +527,6 @@ impl ToVariant for str {
     }
 }
 
-impl<T: ToVariant> From<T> for Variant {
-    fn from(value: T) -> Variant {
-        value.to_variant()
```
sdroege, line 534: This impl also had the problem that it took its argument by value, i.e. took ownership of it and consumed it, but then only used it by reference. This is rather inefficient...

sdroege (approved):
Looks good to me. @EPashkin @zeenix @GuillaumeGomez ?

src/variant.rs
sdroege: Also for having a type that allows deconstructing via a "wildcard", you could add some dummy type that implements the trait and just does nothing, or not?

zeenix:
> `thread 'main' panicked at 'assertion failed: ty.starts_with("a") || ty.starts_with("{") || ty.starts_with("(")', /home/greg/.local/share/cargo/git/checkouts/glib-53370c4288e47ffa/633b61a/src/variant.rs:197:9`

Already fixed in master.

> (using `get_child_value` to walk the `a{sv}` also gets me to these same variants with `v` inside)
> 
> (why do they exist?)

What?

zeenix:
@sdroege sounds good to me.

FWIW, this is what I was doing in zvariant as well but @elmarco insisted that use of std traits rather than our own, was more important than a uniform API to deal with transformation of all types (including Variant itself) to/from Variant type. The huge drawback of not having a uniform API is that user has to be conscious of Variant boxing. So what I'm saying is that I believe this (PR) and custom traits is the best way forward!

myfreeweb:
@zeenix the `v` variants were also fixed with the last push (`ids1024 force-pushed the ids1024:from_variant branch from 633b61a to f6b61f3 11 hours ago`) :+1:

Issue #682, opened 2020-09-04
Add wrappers for GVariantIter


ids1024:
This seems like it could be useful, with a call like `Variant::iter`.

Implementing this seems to run into an issue:

```rust
glib_wrapper! {
    pub struct VariantIter(Boxed<glib_sys::GVariantIter>);

    match fn {
        copy => |ptr| glib_sys::g_variant_iter_copy(ptr),
        free => |ptr| glib_sys::g_variant_iter_free(ptr),
    }
}
```

This fails to compile since `g_variant_iter_copy()` takes a `*mut`, not a `*const`. This is easily fixed with `ptr as *mut _`, but of course Rust considers that undefined behavior.

Not sure what the best solution is. Change the `glib_wrapper!` macro?

sdroege:
Cast or use the mut_override function. it's not UB on pointers, only on references.

PR #683, opened 2020-09-04
Add a VariantIter type, and Variant::iter() method

ids1024:
https://github.com/gtk-rs/glib/issues/682.

### DIFF
--- a/src/variant_iter.rs
+++ b/src/variant_iter.rs
@@ -1,0 +1,113 @@
+// Copyright 2020, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+// This is similar to the GVariantIter provided by glib, but that would
+// introduce a heap allocation and doesn't provide a way to determine how
+// many items are left in the iterator.
+
+use std::iter::{DoubleEndedIterator, ExactSizeIterator, Iterator};
+
+use variant::Variant;
+
+/// Iterator over items in a variant.
+#[derive(Debug)]
+pub struct VariantIter {
+    variant: Variant,
+    head: usize,
+    tail: usize,
+}
+
+impl VariantIter {
+    pub(crate) fn new(variant: Variant) -> Self {
+        let tail = variant.n_children();
+        Self {
+            variant,
+            head: 0,
+            tail,
+        }
+    }
+}
+
+impl Iterator for VariantIter {
+    type Item = Variant;
+
+    fn next(&mut self) -> Option<Variant> {
+        if self.head == self.tail {
+            None
+        } else {
+            let value = self.variant.get_child_value(self.head);
+            self.head += 1;
+            Some(value)
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let size = self.tail - self.head;
+        (size, Some(size))
+    }
+}
+
+impl DoubleEndedIterator for VariantIter {
+    fn next_back(&mut self) -> Option<Variant> {
+        if self.head == self.tail {
+            None
+        } else {
+            self.tail -= 1;
+            Some(self.variant.get_child_value(self.tail))
+        }
+    }
+}
+
+impl ExactSizeIterator for VariantIter {}
+
+#[cfg(test)]
+mod tests {
+    use prelude::*;
+    use std::collections::HashMap;
+    use variant::DictEntry;
+    use variant::Variant;
+
+    #[test]
+    fn test_variant_iter_variant() {
+        let v = Variant::variant(&"foo".to_string().to_variant());
+        let vec: Vec<String> = v.iter().map(|i| i.get().unwrap()).collect();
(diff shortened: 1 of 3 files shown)


sdroege (approved):

src/variant_iter.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn next(&mut self) -> Option<Variant> {
+        unsafe {
+            let value = glib_sys::g_variant_iter_next_value(mut_override(self.to_glib_none().0));
+            if !value.is_null() {
+                Some(from_glib_full(value))
+            } else {
+                None
+            }
+        }
+    }
```
sdroege, line 32: You probably want to implement `Iterator::size_hint()` and `ExactSizeIterator` here around `g_variant_iter_n_children()`. That would make things like `Iterator::collect()` work better :)

Note that the Rust functions are about the remaining elements, while the `GVariantIter` function is about the overall elements.

ids1024: Yeah, I thought about that.

> Note that the Rust functions are about the remaining elements, while the GVariantIter function is about the overall elements.

This part is a bit of a pain, since it means it would have to maintain a count of how many elements it has gone through so far. I guess the overhead of that wouldn't be too significant. But I suppose `glib_wrapper!` wouldn't support adding an additional member to the wrapper struct.

sdroege: Indeed, that sucks :) Maybe instead we can implement an iterator ourselves around counting? Instead of using the C `GVariantIter`. That seems equally simple and there doesn't seem to be any advantage in using the C iterator, it just adds another heap allocation.

ids1024: Hm. I wasn't sure if it the C implementation had any advantages over attempting to reimplement it (if variants used linked lists for instance, it could iterate in a more efficient manner).

But looking at the glib source code, it seems like `g_variant_iter_next_value()` is just implemented with `g_variant_get_child_value()`, and stores a `i` and `n` with the length and position in the container, and a reference to the `Variant`.

I don't see any reason that couldn't be reimplemented in Rust without an extra heap allocation.

ids1024:
I've updated to an implementation using `g_variant_get_child_value()`, which now implements `ExactSizeIterator`.

src/variant_iter.rs
```diff
@@ -0,0 +1,93 @@
+// Copyright 2020, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use std::iter::{ExactSizeIterator, Iterator};
+
+use variant::Variant;
+
+/// Iterator over items in a variant.
+pub struct VariantIter {
```
sdroege, line 10:
```suggestion
#[derive(Debug)]
pub struct VariantIter {
```

src/variant_iter.rs
```diff
@@ -0,0 +1,93 @@
...
+        Self { variant, n, i: 0 }
+    }
+}
+
+impl Iterator for VariantIter {
+    type Item = Variant;
+
+    fn next(&mut self) -> Option<Variant> {
+        if self.i < self.n {
+            let value = self.variant.get_child_value(self.i);
```
sdroege, line 28: Could of course also implement `DoubleEndedIterator` now :man_shrugging:

src/variant_iter.rs
```diff
@@ -0,0 +1,93 @@
+// Copyright 2020, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use std::iter::{ExactSizeIterator, Iterator};
+
+use variant::Variant;
+
```
sdroege, line 8: Maybe add a small comment here why we don't just add bindings for `GVariantIter` but do our own simple version of it

sdroege (commented):
Looks good to me apart from those little details :) Thanks!

src/variant_iter.rs
```diff
@@ -0,0 +1,113 @@
...
+// many items are left in the iterator.
+
+use std::iter::{DoubleEndedIterator, ExactSizeIterator, Iterator};
+
+use variant::Variant;
+
+/// Iterator over items in a variant.
+#[derive(Debug)]
+pub struct VariantIter {
+    variant: Variant,
```
sdroege, line 16: Why not `&'a Variant`? We don't need to clone above, especially not because variants are immutable :)

ids1024: The clone just increases the reference count, right? This matches the behavior of `GVariantIter`, which calls `g_variant_ref`.

If `VariantIter` just takes a reference to the `Variant`, that should work the same way for most purposes. But it can sometimes be a pain to have the lifetime of one struct tied to that of another. I think avoiding that is probably worth the cost of just incrementing a reference count, while also being more consistent with the fact that most gtk-rs types (like gtk itself) wrap reference counting.

sdroege: Ok, works for me

sdroege (approved):
Will merge once the CI is through

PR #684, opened 2020-09-07
Borrow some values from ParamSpec instead of copying

sdroege:
CC @EPashkin @GuillaumeGomez

### DIFF
--- a/src/param_spec.rs
+++ b/src/param_spec.rs
@@ -10,6 +10,8 @@ use translate::*;
 use ParamFlags;
 use Value;
 
+use std::ffi::CStr;
+
 glib_wrapper! {
     #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
     pub struct ParamSpec(Shared<gobject_sys::GParamSpec>);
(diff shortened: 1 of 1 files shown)


src/param_spec.rs
```diff
@@ -37,20 +39,26 @@ impl ParamSpec {
         unsafe { from_glib((*self.to_glib_none().0).flags) }
     }
 
-    pub fn get_blurb(&self) -> String {
-        unsafe { from_glib_none(gobject_sys::g_param_spec_get_blurb(self.to_glib_none().0)) }
+    pub fn get_blurb(&self) -> &str {
```
heftig, line 42: These functions should use the `pub fn get_blurb<'a>(&self) -> &'a str` pattern to free the output lifetime since we know the strings come from GQuarks and are valid for `'static`.

heftig: I was wrong, this only applies to `get_name`. Nick and blurb are stored in the pspec.

heftig: We could have a `pub fn get_static_blurb<'a>(&self) -> Option<&'a str>` which checks the `flags` for `G_PARAM_STATIC_BLURB` but I'm not sure that's worth the effort.

sdroege: That doesn't seem very useful, but yes let's do that for name.

GuillaumeGomez (approved):
Since we can't clone the type and can't modify it when the returned "get content" is still held, then the data is "safe" so all good for me. Thanks!

EPashkin (approved):
@sdroege Thanks, LGFM

PR #685, opened 2020-09-08
Refactor property handling in subclasses

sdroege:


sdroege:
@GuillaumeGomez Let's get this in before more merge conflicts happen :) I have more plans with this but that can come iteratively.

sdroege:
https://github.com/gtk-rs/examples/pull/335 is needed in addition for the examples to pass

src/subclass/mod.rs
```diff
@@ -141,25 +106,52 @@
 //!
 //!     // Trait that is used to override virtual methods of glib::Object.
 //!     impl ObjectImpl for SimpleObject {
+//!         // Called once in the very beginning to list all properties of this class.
+//!         fn properties(_klass: &Self::Class) -> Vec<glib::ParamSpec> {
```
sdroege, line 110: The main question here is whether this here (a method on the `ObjectImpl` trait) than a `klass.install_properties()` inside `ObjectSubclass::class_init()`.

What do you think @GuillaumeGomez?

Maybe @ids1024 also has an opinion on this :)

GuillaumeGomez: I don't really have one. I'm not using glib much so might be better to have opinion from someone using glib more often (like you for example? :p).

bilelmoussaoui: Pinging @hfiguiere as well as they wrote a bunch of custom widgets using Rust

ids1024: I'm probably not an authority given I haven't even (yet) implemented a GLib class in Rust that had properties. (At the moment, my interest is having as little boilerplate as possible to create a simple subclass).

I suppose placing it as a method of the `ObjectImpl` trait seems more Rustic. But `klass.install_properties` would make it be more consistent with the C API and make it clear when it is run.

Another option, though theoretically limited, would be an associated constant of type `&'static [glib::ParamSpec]`. Can't think of any reason a `ParamSpec` couldn't be provided in a constant context, though that could be a concern.

sdroege: It can't be an associated constant because we can't create a `glib::ParamSpec` at compile-time, otherwise I can think of a few cases where you would only know the properties at runtime but I don't think those would happen in Rust and also you wouldn't be able to use the `ObjectSubclass` trait there anyway.

I'll remove the `class` parameter to the function though, that seems not very useful :)

ids1024: > we can't create a `glib::ParamSpec` at compile-time

Ah, I was thinking about the example in `glib::subclass` which uses a static, but looking at that again I see that uses `subclass::Property`, which is kind of an ugly type (and I see you just removed).

Yeah, definitely can't be constant given it takes an FFI call to generate it. So I guess that is eliminated.

ids1024: Oh, or maybe it could be a constant if it's possible/valid to generate a paramspec in Rust without FFI calls. It's probably not too complicated of a C struct, if glib considers its representation fixed and something language bindings are allowed to assume.

sdroege: It needs to be allocated by GLib, so not really any choice here :)

sdroege: I guess let's get this in then later today unless there are any objections. So far all feedback was positive, but I'll also have to update the GStreamer repos before merging this so they don't fail compiling :)

sdroege:
I'll change things a bit more first. For properties I think this can be improved a bit further, and for signals I also have an idea.

sdroege:
Superseded by https://github.com/gtk-rs/gtk-rs/pull/31

PR #688, opened 2020-09-09
Improve ParamSpec bindings

sdroege:
* [x] Missing param spec subclasses
* [x] At least some tests

### DIFF
--- a/src/param_spec.rs
+++ b/src/param_spec.rs
@@ -143,7 +143,7 @@ impl ParamSpec {
         flags: ParamFlags,
     ) -> ParamSpec {
         unsafe {
-            from_glib_full(gobject_sys::g_param_spec_boolean(
+            from_glib_none(gobject_sys::g_param_spec_boolean(
                 name.to_glib_none().0,
                 nick.to_glib_none().0,
                 blurb.to_glib_none().0,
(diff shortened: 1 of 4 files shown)


src/param_spec.rs
```diff
@@ -547,3 +605,230 @@ impl ParamSpec {
...
+    StaticType + FromGlibPtrNone<*mut gobject_sys::GParamSpec> + 'static
+{
+}
+
+extern "C" {
+    pub static g_param_spec_types: *const glib_sys::GType;
+}
+
+macro_rules! define_param_spec {
+    ($typ:ident, $ffi_typ:path, $mod_name:ident, $typ_offset:expr) => {
```
GuillaumeGomez, line 619: You could rename `$ffi_typ` into `$ffi_type` and `$typ_offset` into `$type_offset`.

sdroege: Indeed, just seems inconsistent with `$typ` then :) Any suggestions for making it the same?

GuillaumeGomez: `$rust_type`?

sdroege: Thanks, better :)

sdroege:
All but the last commit should go into a bugfix release :)

GuillaumeGomez:
You still need to add some tests before that. ;) Otherwise no issue for the bugfix release.

sdroege:
> You still need to add some tests before that. ;)

Tests for? The existing tests found these bugs, just had to run in valgrind :)

GuillaumeGomez:
You listed that in your first comment here. Just quoting. :)

sdroege:
Ah, those are in the last commit.

GuillaumeGomez (approved):

GuillaumeGomez:
Oh indeed, went too fast and didn't see them. Looks good to me! :+1:

Issue #569, opened 2020-01-06
Trait implementation of ToGlibContainerFromSlice for GArray is not initialized


gwutz:
Right now the Trait implementation initializes the GArray to a ptr::null_mut() which will crash for Slice to GArray.

As this is only implemented for Pointer Types maybe initialize the GArray with a size of a pointer. In my specific usecase i needed to please an API for int64_t in a GArray which is not possible with that trait as only pointers are covered.

Relevant code: https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L994
```rust
#[inline]
fn to_glib_full_from_slice(t: &[T]) -> *mut glib_sys::GArray {
    let mut arr: *mut glib_sys::GArray = ptr::null_mut();
    unsafe {
        for ptr in t.iter().map(ToGlibPtr::to_glib_full) {
            arr = glib_sys::g_array_append_vals(arr, Ptr::to(ptr), 1);
        }
    }
    arr
}
```

sdroege:
As the current impls don't work at all we can also remove them for now until someone has a use-case for them, and then figure out how to have them correctly for that use-case.

EPashkin:
Seems now we have has use-case ;)
@gwutz can you try replace first line with this?
```
        let mut arr: *mut glib_sys::GArray = unsafe {
            glib_sys::g_array_new(false.to_glib(), false.to_glib(), mem::size_of::<T>() as u32)
        };
```

EPashkin:
As we know size seems better use `g_array_sized_new` instead.
IMHO Main problem in this issue: how to know is array need be zero_terminated or not.

sdroege:
Also how to distinguish pointer types from fundamental types.

sdroege:
>             glib_sys::g_array_new(false.to_glib(), false.to_glib(), mem::size_of::<T>() as u32)

This is only correct for fundamental types (the size).

EPashkin:
Yes, I missed that implementation only for `ToGlibPtr` and it really need size of pointer,
Also I used `T` instead of `<T as GlibPtrDefault>::GlibType`, so it not works too.

EPashkin:
@gwutz seems better for you just add custom wrapper type on `vec` or slice and implement this trait or only  function on it

PR #692, opened 2020-09-14
Remove GArray FFI translation impls

sdroege:
They can't possibly work right now and would just crash with a NULL
pointer dereference.

Fixes https://github.com/gtk-rs/glib/issues/569

----

CC @GuillaumeGomez @EPashkin


GuillaumeGomez (approved):
:+1:

PR #693, opened 2020-09-14
Use Duration instead of u32 when representing milliseconds

### LINKED ISSUE #691: glib::timeout_add and friends should probaby take a Duration instead
It seems to me that it makes more sense to have `gtk::timeout_add` and `glib::timeout_add{,_local}` take a [`Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html) and remove their `_seconds` counterparts.

This would make it clearer on the meaning of the parameter, and matches better to what the standard library does, thus feels more idiomatic.

GuillaumeGomez:
Fixes #691.

cc @sdroege @EPashkin

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -10,6 +10,7 @@ use std::mem::transmute;
 use std::num::NonZeroU32;
 #[cfg(unix)]
 use std::os::unix::io::RawFd;
+use std::time::Duration;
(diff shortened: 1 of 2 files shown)


sdroege (approved):

sdroege:
CI is unhappy

sdroege (approved):

sdroege:
CI still unhappy. The examples need fixes

Issue #658, opened 2020-07-02
Support for logging through the log crate


xanathar:
Hi,
searching for a solution to integrate a Rust project with glib logging, I stumbled on the [glib_logger project](https://github.com/bboozzoo/glib_logger/), which bridges glib logging with the [log crate](https://docs.rs/log/0.4.8/log/).

Since it didn't have full support for domains and other features I was interested in, I sent a [PR (still pending)](https://github.com/bboozzoo/glib_logger/pull/1) which adds the missing bits.

However, I was wondering if there is an interest in adopting such a solution in the main gtk-rs sources, given that logging is quite a basic functionality which pretty much all glib applications need, and that this is a fair solution at being as idiomatic as possible while still integrating with glib log handlers.

Would a PR for this be welcome (in which project) or not? 
Or, in general, your thoughts?

sdroege:
I think having integration into the log crate into both directions would be useful to have in the glib bindings (`glib` is the repo you want to create the PR for), but it should be hidden behind a feature flag.

xanathar:
Cool thanks, I'll send a contribution as soon as I can.
(you can close this issue or keep it open until the pr, as you prefer)

sdroege:
Now the only other thing that would be useful to have is a GLib logging function that directly calls into the `log` crate instead of doing its own stuff. Then we have integration in either direction, and that one is actually more useful for most of my code :)

xanathar:
I can do that, but it will take a little more time.

sdroege:
No worries :)

xanathar:
I've done this... but it's very very minimal:

```rust
fn default_glib_log_handler(string1: &str, level: glib::LogLevel, string2: &str) {
    let level = match level {
        glib::LogLevel::Error | glib::LogLevel::Critical => log::Level::Error,
        glib::LogLevel::Warning => log::Level::Warn,
        glib::LogLevel::Message | glib::LogLevel::Info => log::Level::Info,
        glib::LogLevel::Debug => log::Level::Debug,
    };

    log!(target: string1, level, "{}", string2);
}

pub fn setup() {
    glib::log_set_default_handler(default_glib_log_handler);
}
```

I don't even know if it's worth adding to the glib-rs library, given that it's such small snippet (though I can easily send a PR for that).

Did you have in mind something with a bigger scope?

sdroege:
> Did you have in mind something with a bigger scope?

[...]

PR #694, opened 2020-09-15
Proposal: support for logging through the log crate

xanathar:
This PR is (as now) more meant as a base for discussion rather than something meant to be merged (though, in case, it's tested and somewhat ready).

This is a follow-up to this issue: https://github.com/gtk-rs/glib/issues/658

The changes contain:
- a `GlibLogger` struct implementing the `log::Log` trait, so that code using the [log crate](https://crates.io/crates/log) will log to glib automatically (with customizable behavior)
- a series of optional macros which override the macros defined in the [log crate](https://crates.io/crates/log), which allow to specify the logging domain in a `G_LOG_DOMAIN` static/const

The two parts above are "activated" by two different features in `Cargo.toml`: `bridged_logging` and `bridged_logging_domain_macros`.

I understand these are probably a bit "opinionated" in their implementation, so feedback is welcome; probably the most controversial part could be the overridden macros: if those are an obstacle they can be removed and only the adapter kept.

I think a sizable part of people interfacing glib code with Rust code using gtk-rs will want to have the two logging systems cooperating; currently this is done by custom code or external third party crates, so this might be helpful.

Example code using this, would be:

```rust
static glib_logger: glib::log_bridge::GlibLogger = glib::log_bridge::GlibLogger {
    format: glib::log_bridge::GlibLoggerFormat::Structured,
    domain: glib::log_bridge::GlibLoggerDomain::CrateTarget,
};

fn main() {
    log::set_logger(&glib_logger);
    log::set_max_level(log::LevelFilter::Debug);

    info!("This line will get logged by glib");

    error!("This is a Rust error");

    info!("This is from Rust %s and doesn't crash");

    g_info!("domain", "This is using glib logging");
}
```


sdroege:
>     * a `GlibLogger` struct implementing the `log::Log` trait, so that code using the [log crate](https://crates.io/crates/log) will log to glib automatically (with customizable behavior)

I like this in principle, will review tomorrow :) I wanted to have this in glib since a while.

>     * a series of optional macros which override the macros defined in the [log crate](https://crates.io/crates/log), which allow to specify the logging domain in a `G_LOG_DOMAIN` static/const

This is confusing me a bit. How does that differ from [`g_info!`](https://gtk-rs.org/docs/glib/macro.g_info.html) etc?

xanathar:
> This is confusing me a bit. How does that differ from g_info! etc?

Two ways mainly:
- they are named like the log crate macros (so they are quite a "plug&play" replacement, changing the `macro_use` in `lib.rs`/`main.rs` is usually enough)
- they default to use a variable named `G_LOG_DOMAIN ` (which must be in scope) if no context is provided, thus sort of emulating what `G_LOG_DOMAIN ` does in C.

As I said, those are the most opinionated part.

src/bridged_logging.rs
```diff
@@ -0,0 +1,226 @@
...
+        func: Option<&str>,
+        message: &str
+    ) {
+        let line_str = match line {
+            None => None,
+            Some(l) => Some(l.to_string()),
+        };
+
+        unsafe {
+            crate::glib_sys::g_log_structured_standard(
```
sdroege, line 99: The structured logging feature is new in GLib 2.50, but as we don't actually log any custom structured information just going via `g_log` should be sufficient or not?

sdroege: Oh nevermind, actually we do. And this function is new in 2.56 :) So needs some condition compilation around the `v2_56` feature.

src/bridged_logging.rs
```diff
@@ -0,0 +1,226 @@
...
+#[cfg(any(feature = "dox", feature = "bridged_logging_domain_macros"))]
+/// A macro which behaves exactly as `log::trace!` except that it sets the
+/// current log target to the contents of a `G_LOG_DOMAIN` constant (and fails
+/// to build if not defined).
+macro_rules! trace {
+    (target: $target:expr, $($arg:tt)+) => (
+        log::log!(target: $target, log::Level::Trace, $($arg)+);
+    );
+    ($($arg:tt)+) => (
+        log::log!(target: G_LOG_DOMAIN, log::Level::Trace, $($arg)+);
```
sdroege, line 224: I didn't know that the macro hygiene rules allowed this, interesting.

But can you put these macros into a separate commit in the same PR?

src/lib.rs
```diff
@@ -198,6 +199,9 @@ pub use log::{
     unset_print_handler, unset_printerr_handler, LogHandlerId, LogLevel, LogLevels,
 };
 
+#[cfg(any(feature = "dox", all(feature = "v2_46", feature = "bridged_logging")))]
+pub mod bridged_logging;
```
sdroege, line 203: Maybe `pub use bridged_logging::GlibLogger`?

src/bridged_logging.rs
```diff
@@ -0,0 +1,226 @@
...
+/// fn main() {
+///     log::set_logger(&glib_logger);
+///     log::set_max_level(log::LevelFilter::Debug);
+///
+///     info!("This line will get logged by glib");
+/// }
+/// ```
+pub struct GlibLogger {
+    pub format: GlibLoggerFormat,
+    pub domain: GlibLoggerDomain,
```
sdroege, line 59: Maybe make these private and private a constructor

src/bridged_logging.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// Logs will use the `target` of the log crate as a domain; this allows
+    /// Rust code like `warn!(target: "my-domain", "...");` to log to the glib
+    /// logger using the specified domain.
+    CrateTarget,
+    /// Logs will use the crate path as the log domain.
+    CratePath,
+    /// Logs will use the specified domain for logging. Note that this option
+    /// due to the way the `log` crate is implemented, is global for all crates
+    /// in the current process.
+    Custom(&'static str),
```
sdroege, line 36: Is this really useful?

sdroege:
Looks good to me generally

src/bridged_logging.rs
xanathar: Sure, I'll do one forced commit without and another with.

src/bridged_logging.rs
xanathar: I'm on the fence here.
On one side I assume there will be somebody who wants to add a Rust library to an existing glib based project and is happy to specify a single domain for all Rust logging; on the other they can get a better more future-proof result with a little more effort specifying a domain everywhere.

xanathar:
Sorry I mistakenly closed the PR.

src/bridged_logging.rs
sdroege: I don't know, @GuillaumeGomez ? :)

xanathar:
CI failures seem to be related to formatting. I will fix them by today, along with a couple of docs comments which are not accurate after the latest changes, and I will undraft this.

Open questions:
- What do we want to do with `GlibLoggerDomain::Custom(&'static str)`
- What to do with the override macros (is a separate commit enough)

sdroege:
> What do we want to do with `GlibLoggerDomain::Custom(&'static str)`

I don't see its point personally, but maybe someone else does? :)

> What to do with the override macros (is a separate commit enough)

I like them and just want a second opinion.

src/bridged_logging.rs
GuillaumeGomez: I think it's a good idea. However, a static `str` might be too restrictive.

sdroege: Must be either that or a `String`

GuillaumeGomez: A `string` would be preferable indeed.

xanathar: I fear a `String` cannot be done (at the very least not without some dirty tricks), because the logger is a `&'static` and, as far as I can tell, there is no way to build a `static String` (all its constructors are not `const fn`).

Possible workarounds are with a `lazy_static` or with a `RefCell` inside the logger, but at that point I fear it becomes dirtier and harder to use correctly vs the potential gains.

*(disclaimer: limited knowledge of Rust here, so I might be missing something)

sdroege: You could use [`set_boxed_logger()`](https://docs.rs/log/0.4.11/log/fn.set_boxed_logger.html) but I agree that it's a bit ugly.

So what would be the use-case for this custom string, how would one use it and isn't it confusing to just have everything going into the `log` crate showing up the same way in the glib logs?

xanathar: My 2 cents given the current state: let's remove this option, move forward with the rest and eventually add it in the future in another pr if and when any reasonable need+solution comes up.

sdroege: Sounds good to me :)

GuillaumeGomez: Sounds good to me as well.

GuillaumeGomez:
I like the macros too. :)

sdroege:
It looks like the macros disappeared from the changes somehow?

src/bridged_logging.rs
```diff
@@ -0,0 +1,175 @@
+// Copyright 2020, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+#![cfg(any(feature = "dox", all(feature = "v2_46", feature = "bridged_logging")))]
```
sdroege, line 5: We usually put those above the `mod` keyword

Cargo.toml
```diff
@@ -48,7 +49,8 @@ v2_58 = ["v2_56", "glib-sys/v2_58", "gobject-sys/v2_58"]
 v2_60 = ["v2_58", "glib-sys/v2_60"]
 v2_62 = ["v2_60", "glib-sys/v2_62", "gobject-sys/v2_62"]
 v2_64 = ["v2_62", "glib-sys/v2_64"]
-dox = ["glib-sys/dox", "gobject-sys/dox"]
+bridged_logging = ["v2_46", "log"]
```
sdroege, line 52: Which part needs GLib 2.46?

src/bridged_logging.rs
```diff
@@ -0,0 +1,175 @@
...
+    Plain,
+    /// A simple format, writing file, line and message on output.
+    LineAndFile,
+    /// A logger using glib structured logging.
+    #[cfg(any(feature = "v2_56", feature = "dox"))]
+    Structured,
+}
+
+/// Enumeration of the possible domain handling behaviours for a [`GlibLogger`]
+pub enum GlibLoggerDomain {
```
sdroege, line 25: Derive Debug, Clone, Copy, PartialEq, Eq :)

src/bridged_logging.rs
```diff
@@ -0,0 +1,175 @@
...
+/// );
+///
+/// fn main() {
+///     log::set_logger(&glib_logger);
+///     log::set_max_level(log::LevelFilter::Debug);
+///
+///     info!("This line will get logged by glib");
+/// }
+/// ```
+pub struct GlibLogger {
```
sdroege, line 54: Derive Debug at least

sdroege:
Looks good to me otherwise

Cargo.toml
xanathar: Ooops. My eyes had been tricked by 

```rust
mod log;
#[cfg(any(feature = "v2_46", feature = "dox"))]
pub use log::log_set_handler;
```

You are right, there's no need for v2_46 here.

xanathar:
> It looks like the macros disappeared from the changes somehow?

You told me to do a second separate commit... so I removed them and adding them later in a new commit. Have I misunderstood what you wanted ?

sdroege:
> You told me to do a second separate commit... so I removed them and adding them later in a new commit. Have I misunderstood what you wanted ?

A separate commit in this PR. It's a separate functional change, so should be a separate commit but it can be part of the same PR as it's related.

Personally I would also clean up the commit history of this PR so it doesn't contain all the back and forth but just a clean history but I guess we can forget that for the gtk repository anyway :)

xanathar:
> Personally I would also clean up the commit history of this PR so it doesn't contain all the back and forth but just a clean history but I guess we can forget that for the gtk repository anyway :)

My github-fu is a bit lacking, but I might have done it without accidentally closing the PR like I did last time 😅

sdroege (approved):

xanathar:
Thanks to you :)

Cargo.toml
```diff
@@ -48,7 +49,9 @@ v2_58 = ["v2_56", "glib-sys/v2_58", "gobject-sys/v2_58"]
 v2_60 = ["v2_58", "glib-sys/v2_60"]
 v2_62 = ["v2_60", "glib-sys/v2_62", "gobject-sys/v2_62"]
 v2_64 = ["v2_62", "glib-sys/v2_64"]
-dox = ["glib-sys/dox", "gobject-sys/dox"]
+bridged_logging = ["log"]
+bridged_logging_domain_macros = ["bridged_logging", "log"]
```
GuillaumeGomez, line 53: The features' name aren't great... :-/

Cargo.toml
```diff
@@ -48,7 +49,9 @@ v2_58 = ["v2_56", "glib-sys/v2_58", "gobject-sys/v2_58"]
...
+dox = ["glib-sys/dox", "gobject-sys/dox", "log"]
```
GuillaumeGomez, line 54:
```suggestion
dox = ["glib-sys/dox", "gobject-sys/dox", "bridged_logging_domain_macros"]
```

src/bridged_logging.rs
```diff
@@ -0,0 +1,245 @@
+// Copyright 2020, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+extern crate log as rs_log;
```
GuillaumeGomez, line 5: It's weird to import a crate outside of the `lib.rs` file. What do you think @sdroege ?

src/bridged_logging.rs
```diff
@@ -0,0 +1,245 @@
...
+///
+/// Example:
+///
+/// ```no_run
+/// static glib_logger: glib::GlibLogger = glib::GlibLogger::new(
+///     glib::GlibLoggerFormat::Structured,
+///     glib::GlibLoggerDomain::CrateTarget,
+/// );
+///
+/// fn main() {
```
GuillaumeGomez, line 46: Just a nit but you don't need to put this code in the `main` function.

src/bridged_logging.rs
```diff
@@ -0,0 +1,245 @@
...
+    ///
+    /// Example:
+    ///
+    /// ```no_run
+    /// static glib_logger: GlibLogger = glib::GlibLogger::new(
+    ///     glib::GlibLoggerFormat::Structured,
+    ///     glib::GlibLoggerDomain::CrateTarget,
+    /// );
+    ///
+    /// fn main() {
```
GuillaumeGomez, line 70: nit: no need for `main` function or any function.

src/bridged_logging.rs
```diff
@@ -0,0 +1,245 @@
...
+            rs_log::Level::Error => glib_sys::G_LOG_LEVEL_CRITICAL,
+            rs_log::Level::Warn => glib_sys::G_LOG_LEVEL_WARNING,
+            rs_log::Level::Info => glib_sys::G_LOG_LEVEL_INFO,
+            rs_log::Level::Debug => glib_sys::G_LOG_LEVEL_DEBUG,
+            // There is no equivalent to trace level in glib
+            rs_log::Level::Trace => glib_sys::G_LOG_LEVEL_DEBUG,
+        }
+    }
+
+    fn write_log(domain: Option<&str>, level: rs_log::Level, message: &str) {
```
GuillaumeGomez, line 93: Just wondering: isn't this function supposed to be public?

src/bridged_logging.rs
```diff
@@ -0,0 +1,245 @@
...
+            crate::glib_sys::g_log(
+                domain.to_glib_none().0,
+                GlibLogger::level_to_glib(level),
+                message.replace("%", "%%").to_glib_none().0,
+            );
+        }
+    }
+
+    #[cfg(any(feature = "v2_56", feature = "dox"))]
+    fn write_log_structured(
```
GuillaumeGomez, line 104: Just wondering: isn't this function supposed to be public?

sdroege: No, it's an internal function called from the public one.

src/bridged_logging.rs
sdroege: No, it's an internal function called from the public one.

src/bridged_logging.rs
sdroege: I don't see the problem :)

Cargo.toml
sdroege: Indeed but I don't have a better suggestion. Do you? :)

GuillaumeGomez, line 53:
```suggestion
bridged_logging_domain_macros = ["bridged_logging"]
```

GuillaumeGomez: No indeed. Doesn't mean we shouldn't try to find better ones though. :p

sdroege: Other crates seem to just call the feature `log`, so we could do the same. `log` and `log_macros`

GuillaumeGomez: Sounds good to me!

xanathar:
Changed the code with all the suggestions collected so far.

src/lib.rs
```diff
@@ -198,6 +199,12 @@ pub use log::{
     unset_print_handler, unset_printerr_handler, LogHandlerId, LogLevel, LogLevels,
 };
 
+#[cfg(any(feature = "log", feature = "dox"))]
+#[macro_use]
+mod bridged_logging;
+#[cfg(any(feature = "log", feature = "dox"))]
```
sdroege, line 205: Does this work without declaring a feature for this in `Cargo.toml`? The optional dependency implicitly declares a feature?

Cargo.toml
```diff
@@ -48,7 +49,8 @@ v2_58 = ["v2_56", "glib-sys/v2_58", "gobject-sys/v2_58"]
 v2_60 = ["v2_58", "glib-sys/v2_60"]
 v2_62 = ["v2_60", "glib-sys/v2_62", "gobject-sys/v2_62"]
 v2_64 = ["v2_62", "glib-sys/v2_64"]
-dox = ["glib-sys/dox", "gobject-sys/dox"]
+log_macros = ["log"]
```
sdroege, line 52: Can you change `.travis.yml` to to also enable both new features in the cases it builds for the latest GLib version currently?

src/lib.rs
xanathar: From what I could get: dependencies and features share the same namespace in Cargo.toml, so any optional dependency is implicitly also something you can specify as a feature.

If you are worried about discoverability, I think we can rename the `log` dependency giving it an alias, and then have a separate feature with the `log` name.

sdroege: As long as `cargo build --features log` works, this is ok for me. Discoverability of features is a whole other story and also suboptimal for other features already. For now this just has to be clear from the documentation.

Cargo.toml
xanathar: I can do it, but looking at `.travis.yml` it seems that it does two builds, one against the empty feature set, and the other against `v2_48`.

Do you want me to add `log_macros` beside `v2_48` or upgrade to the latest, while we are at it (considering that structured logging requires `v2_56`)?

sdroege: Add it besides the other one. I don't think we can update to latest because travis runs an ancient version of Ubuntu, but maybe that changed in the meantime.

src/lib.rs
xanathar: I added some notes in the documentation, now it should be clearer.

xanathar:
I've clarified the usage in documentation, fixed the code examples and added the new features to Travis' CI (hoping it gets green :) ).
This should address all feedback received so far.

sdroege:
CI seems happy :) Thanks!

PR #695, opened 2020-09-30
Fix for g_param_spec_types symbol linking error with MSVC

seungha-yang:
Regression from https://github.com/gtk-rs/glib/pull/688

cc: @sdroege

### DIFF
--- a/src/param_spec.rs
+++ b/src/param_spec.rs
@@ -611,6 +611,7 @@ pub trait ParamSpecType:
 {
 }
 
+#[link(name = "gobject-2.0")]
 extern "C" {
     pub static g_param_spec_types: *const glib_sys::GType;
 }


GuillaumeGomez (approved):

seungha-yang:
Sorry for this unclear commit message. (I don't fully understand why `#[no_mangle]` is needed here.

I've seen this build error with [gst-plugins-rs](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs)
```
libglib-20259afa4dec16ec.rlib(glib-20259afa4dec16ec.glib.8h6me5m7-cgu.8.rcgu.o) : error LNK2019: unresolved external symbol g_param_spec_types referenced in function _ZN92_$LT$glib..param_spec..param_spec_char..ParamSpecChar$u20$as$u20$glib..types..StaticType$GT$11static_type17h09c13f9e266f0171E
libglib-20259afa4dec16ec.rlib(glib-20259afa4dec16ec.glib.8h6me5m7-cgu.11.rcgu.o) : error LNK2001: unresolved external symbol g_param_spec_types
...
```

src/param_spec.rs
```diff
@@ -612,6 +612,7 @@ pub trait ParamSpecType:
 }
 
 extern "C" {
+    #[no_mangle]
```
sdroege, line 615: It's unclear to me why this is actually required here but not for functions in `extern "C"` blocks and only on Windows/MSVC. The [nomicon](https://doc.rust-lang.org/nomicon/ffi.html#accessing-foreign-globals) also does not mention it.

A stackoverflow answer from @kornelski also mentions it though: https://stackoverflow.com/questions/45550387/how-can-i-access-a-c-global-variable-constant-in-rust-ffi

kornelski: I'm not sure if that code is necessary, or just happens to parse.

sdroege: Well here it fixes a linker error :) So some more digging is needed for what `rustc` is actually doing here.

sdroege:
Curious:

```
warning: attribute should be applied to a function or static

   --> /home/travis/build/gtk-rs/glib/src/param_spec.rs:615:5

    |

615 |     #[no_mangle]

    |     ^^^^^^^^^^^^

616 |     pub static g_param_spec_types: *const glib_sys::GType;

    |     ------------------------------------------------------ not a function or static

    |

    = note: `#[warn(unused_attributes)]` on by default

    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
```

src/param_spec.rs
seungha-yang: `#[no_mangle]` wasn't right fix. we need `#[link(name = "gobject-2.0")]` for the fix

sdroege (approved):

PR #696, opened 2020-10-04
Use https for all links

dns2utf8:
Hi all

This is a followup to https://github.com/gtk-rs/gtk/pull/1054

Cheers,
Stefan

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@ Alongside GLib bindings, this crate also includes the infrastructure to
 subclass `GObject` for extending libraries like GTK+. A code example can
 be found in the documentation of this crate in the `subclass` module.
 
-- [Gtk-rs project site](http://gtk-rs.org/)
+- [Gtk-rs project site](https://gtk-rs.org/)
 
-- [Online documentation](http://gtk-rs.org/docs/)
+- [Online documentation](https://gtk-rs.org/docs/)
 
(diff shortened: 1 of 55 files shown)


README.md
```diff
@@ -6,9 +6,9 @@ Alongside GLib bindings, this crate also includes the infrastructure to
 subclass `GObject` for extending libraries like GTK+. A code example can
 be found in the documentation of this crate in the `subclass` module.
 
-- [Gtk-rs project site](http://gtk-rs.org/)
+- [Gtk-rs project site](https://gtk-rs.org/)
 
-- [Online documentation](http://gtk-rs.org/docs/)
+- [Online documentation](https://gtk-rs.org/docs/)
```
GuillaumeGomez, line 11: Actually, the good link is this one: https://gtk-rs.org/docs-src/

dns2utf8: Ah, there are a lot of links pointing to the old folder

sdroege: @GuillaumeGomez Can we maybe have `docs` forward to `docs-src`? Lots of search results with Google also point to the wrong location.

GuillaumeGomez: It's worth a try at least.

GuillaumeGomez: Apparently, from what I can understand, it requires access to wherever the domain name was registered to be able to add a new redirection. So it's up to @gkoz, I can't do anything about it. :-/

PR #697, opened 2020-10-09
groovy update: regenerate

bilelmoussaoui:
this also generates glib::Uri as a replacement of the removed functions


sdroege:
> this also generates glib::Uri as a replacement of the removed functions

The functions were not removed, but because gir is not very clever it moved them under the new `GUri` API. Because the functions started with `g_uri` :)

src/auto/functions.rs
```diff
@@ -514,6 +514,11 @@ pub fn file_set_contents<P: AsRef<std::path::Path>>(
     }
 }
 
+//#[cfg(any(feature = "v2_66", feature = "dox"))]
+//pub fn file_set_contents_full<P: AsRef<std::path::Path>>(filename: P, contents: &[u8], flags: /*Ignored*/FileSetContentsFlags, mode: i32) -> Result<(), Error> {
```
sdroege, line 518: This can easily be generated if you add the new enum to the `generate` list in `Gir.toml`

src/auto/uri.rs
```diff
@@ -0,0 +1,530 @@
...
+use GString;
+use Quark;
+#[cfg(any(feature = "v2_66", feature = "dox"))]
+use UriFlags;
+#[cfg(any(feature = "v2_66", feature = "dox"))]
+use UriHideFlags;
+
+glib_wrapper! {
+    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct Uri(Shared<glib_sys::GUri>);
```
sdroege, line 25: As it's immutable, you can impl `Send` / `Sync` on this :)

src/auto/uri.rs
```diff
@@ -0,0 +1,530 @@
...
+        }
+    }
+
+    #[cfg(any(feature = "v2_66", feature = "dox"))]
+    fn to_string(&self) -> GString {
+        unsafe { from_glib_full(glib_sys::g_uri_to_string(self.to_glib_none().0)) }
+    }
+
+    #[cfg(any(feature = "v2_66", feature = "dox"))]
+    pub fn to_string_partial(&self, flags: UriHideFlags) -> Option<GString> {
```
sdroege, line 114: @elmarco Can this return `NULL`? Probably not? Should get a `(not nullable)` annotation in GLib then and the corresponding configuration in `Gir.toml`

bilelmoussaoui:
> > this also generates glib::Uri as a replacement of the removed functions
> 
> The functions were not removed, but because gir is not very clever it moved them under the new `GUri` API. Because the functions started with `g_uri` :)

Right, that's what I meant. Sorry, my english is confusing sometimes

src/auto/enums.rs
```diff
@@ -494,3 +494,113 @@ impl FromGlib<glib_sys::GTimeType> for TimeType {
...
+            9 => UriError::BadFragment,
+            value => UriError::__Unknown(value),
+        }
+    }
+}
+
+#[cfg(any(feature = "v2_66", feature = "dox"))]
+impl ErrorDomain for UriError {
+    fn domain() -> Quark {
+        unsafe { from_glib(glib_sys::g_uri_quark()) }
```
sdroege, line 584: This looks wrong. That should be `g_uri_error_quark()`. The above function does not exist.

It's also wrong in the GLib annotations (CC @elmarco )

src/auto/uri.rs
```diff
@@ -0,0 +1,530 @@
...
+                host.to_glib_none().0,
+                port,
+                path.to_glib_none().0,
+                query.to_glib_none().0,
+                fragment.to_glib_none().0,
+            ))
+        }
+    }
+
+    pub fn error_quark() -> Quark {
```
sdroege, line 177: This is would go away once the `UriError` enum has to correct error domain attached to it

src/auto/uri.rs
```diff
@@ -0,0 +1,530 @@
...
+            if error.is_null() {
+                Ok(from_glib_full(ret))
+            } else {
+                Err(from_glib_full(error))
+            }
+        }
+    }
+
+    //#[cfg(any(feature = "v2_66", feature = "dox"))]
+    //pub fn parse_params(params: &str, separators: &str, flags: UriParamsFlags) -> Result</*Unknown conversion*//*Unimplemented*/HashTable TypeId { ns_id: 0, id: 28 }/TypeId { ns_id: 0, id: 28 }, Error> {
```
sdroege, line 298: That's strange. We support specifically `HashTable<String, String>` (and nothing else)

sdroege (commented):
Looks ok otherwise

src/auto/enums.rs
bilelmoussaoui: Can we handle this from our side until a fix lands upstream?

sdroege: I think only be changing the `.gir` file unfortunately.

sdroege: I.e. something in `fix.sh` in `gir-files`. No need to regenerate everything else for that, just glib is enough.

elmarco: It's the error quark string. I don't think we can change it now.

sdroege:
```xml
    <enumeration name="UriError"
                 version="2.66"
                 c:type="GUriError"
                 glib:error-domain="g-uri-quark">
```

This should match with the function for getting the quark AFAIU. There should be an annotation to override that in the enum definition.

sdroege: I think this is actually a bug in `gir`. This is the quark string, not a function. So far there was also always a function with that same name though.

I would suggest adjusting the .gir file for now and then creating an issue for fixing `gir` to handle this correctly. Should maybe just create a `once_cell::Sync::Lazy<glib::Quark>` from that static string instead.

src/auto/uri.rs
bilelmoussaoui: It doesn't seems to be or did I do something wrong?

sdroege: Maybe just put it into the ignore list for now then :) Can be fixed properly while gir is fixed

src/auto/uri.rs
sdroege: See https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1687

src/auto/enums.rs
```diff
@@ -494,3 +494,113 @@ impl FromGlib<glib_sys::GTimeType> for TimeType {
         }
     }
 }
+
+#[cfg(any(feature = "v2_66", feature = "dox"))]
```
sdroege, line 498: Need to add `v2_66` feature also to `Cargo.toml`.

sdroege: This was also forgotten for other crates I guess. Can you create PRs for that? For gio I created https://github.com/gtk-rs/gio/pull/319 already.

src/auto/uri.rs
bilelmoussaoui: So this method shouldn't return an Option<T>?

src/auto/uri.rs
bilelmoussaoui: Putting it in the ignore list doesn't seem to do anything, should I mark the enum as manual and fix the code for now?

src/auto/enums.rs
bilelmoussaoui: AH somehow I missed this one. I will update the relevant crates

src/auto/uri.rs
sdroege: I guess just edit the manual code here now so we can finally get this in, and I'll work on fixing gir this weekend or tomorrow so we don't have this problem anymore in the future. Easy enough to fix there, just need to find 20 minutes of free time.

src/auto/uri.rs
sdroege: Yes, `nullable = false` for the return value in `Gir.toml` is needed here. Also see the various other functions marked as not nullable in that MR.

src/auto/uri.rs
GuillaumeGomez: Manual until fixed (if that was the question :) ). Or fix gir then regen this one (I tend to prefer doing it this way).

sdroege: Then let's wait until the weekend unless @bilelmoussaoui wants to fix gir

bilelmoussaoui: I can look at fixing gir today if I have +/- an idea of how the enum configuration would be? At least for this specific use case

sdroege: See my comments in the issue. It's not about configuration, it's about using that field in the XML file correctly

bilelmoussaoui: Actually, my bad. I was trying to ignore this function in the UriError enum instead of Uri itself. I've ignored the function for now until I (or someone else) get the time to look at gir issue

PR #699, opened 2020-10-13
Added register_glib_to_logcrate_logger function to do glib-to-rust logging

xanathar:
This adds a new `register_glib_to_logcrate_logger` function which routes all logging happening into glib to the log crate commonly used in Rust.

This is basically the reverse of `GlibLogger` (which routes the log crate to glib).

I found no trivial way (due to the fact that GlibLogger has a `const fn` constructor to simplify integration etc.) to prevent the user to use both `GlibLogger` and `register_glib_to_logcrate_logger` at the same time, which causes an obvious stack overflow (luckily, very early in the application lifetime). 

I clarified the potential stack overflow in the documentation. If preferred, different cargo features can be used to "xor" the two functionalities; I felt that to be more complicated than the problem it solved but it's personal opinion :)

Related issue: https://github.com/gtk-rs/glib/issues/658

### DIFF
--- a/src/bridged_logging.rs
+++ b/src/bridged_logging.rs
@@ -2,8 +2,7 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>
 
-extern crate log as rs_log;
-
+use crate::log as glib_log;
 use glib_sys;
 use translate::*;
 
@@ -47,6 +46,16 @@ pub enum GlibLoggerDomain {
 /// In order to use this type, `glib` must be built with the `log` feature
 /// enabled.
 ///
+/// Use this if you want to use glib as the main logging output in your application,
+/// and want to route all logging happening through the log crate to glib logging.
+/// If you want the opposite, see
+/// [`rust_log_handler`](fn.rust_log_handler.html).
+///
+/// NOTE: This should never be used when
+/// [`rust_log_handler`](fn.rust_log_handler.html) has
+/// been registered as a default glib log handler, otherwise a stack overflow
+/// will occur.
+///
 /// Example:
 ///
 /// ```no_run
@@ -70,6 +79,8 @@ pub struct GlibLogger {
 
 impl GlibLogger {
     /// Creates a new instance of [`GlibLogger`](struct.GlibLogger.html).
+    /// See documentation of [`GlibLogger`](struct.GlibLogger.html) for more
+    /// information.
     ///
     /// Example:
     ///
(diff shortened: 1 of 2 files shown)


src/bridged_logging.rs
```diff
@@ -186,6 +198,37 @@ impl rs_log::Log for GlibLogger {
     fn flush(&self) {}
 }
 
+/// Registers a glib log handler which routes all logging messages to the
+/// [`log crate`](https://crates.io/crates/log).
+///
+/// In order to use this type, `glib` must be built with the `log` feature
```
sdroege, line 204:
```suggestion
/// In order to use this function, `glib` must be built with the `log` feature
```

src/bridged_logging.rs
```diff
@@ -186,6 +198,37 @@ impl rs_log::Log for GlibLogger {
...
+///
+/// Use this function if you want to use the log crate as the main logging
+/// output in your application, and want to route all logging happening in
+/// glib to the log crate. If you want the opposite, use [`GlibLogger`](struct.GlibLogger.html).
+///
+/// NOTE: This should never be used when [`GlibLogger`](struct.GlibLogger.html) is
+/// registered as a logger, otherwise a stack overflow will occur.
+///
+/// ```no_run
+/// register_glib_to_logcrate_logger();
```
sdroege, line 215:
```suggestion
/// glib::register_glib_to_logcrate_logger();
```

src/bridged_logging.rs
```diff
@@ -186,6 +198,37 @@ impl rs_log::Log for GlibLogger {
...
+/// ```
+pub fn register_glib_to_logcrate_logger() {
```
sdroege, line 217: The name is a bit weird with the `crate` in there. Not sure what would be a better name :)

Or maybe we could only expose the actual logging function and let users call `log_set_default_handler()` with it?

xanathar: I'm also definitely not happy with the name (long and clumsy).
Sadly there is an overload on the "log" / "logger" terms which is not easy to untangle, and the crate approach means that there is no official Rust logging facility, while at the same time the logging crate is called just... "log", adding to the overload and confusion.
Suggestions are definitely welcome.

(s far as the rest goes, previous comments have been addressed and CI is green)

xanathar: Throwing some options:
- `register_rust_log_handler` / `set_rust_log_handler` / `set_default_rust_log_handler`
- `rust_log_handler` (and the application code calls `log_set_default_handler(rust_log_handler)`
- `redirect_rust_logs_to_glib`
- `setup_glib_logging` (might create confusion with `GlibLogger`)

sdroege: I think I prefer the second option. @GuillaumeGomez ?

GuillaumeGomez: Fine by me!

sdroege: Let's go with that then or what do you think @xanathar ? :)

xanathar: Sure! I'll push it asap

xanathar:
PR is (in theory) ready.

CI fails (on Rust stable targets) with this:

```
git diff -R --exit-code
diff --git b/src/auto/versions.txt a/src/auto/versions.txt
index ff9347c..0f6b6a9 100644
--- b/src/auto/versions.txt
+++ a/src/auto/versions.txt
@@ -1,2 +1,2 @@
-Generated by gir (https://github.com/gtk-rs/gir @ ad40c01)
+Generated by gir (https://github.com/gtk-rs/gir @ ad40c01+)
 from gir-files (https://github.com/gtk-rs/gir-files @ 9f47916)
make: *** [regen_check] Error 1
The command "if [ "$TRAVIS_RUST_VERSION" == "stable" ] && ! [ "$ARM" == "1" ]; then rustup component add rustfmt; make regen_check; fi" exited with 2.
```

Not really sure if it's a problem of the commit (can't see why) or with CI itself.

sdroege:
> Not really sure if it's a problem of the commit (can't see why) or with CI itself.

Problem with the CI, don't worry about that :)

sdroege:
Thanks, looks good :)

PR #702, opened 2020-10-13
Fix .get_instance() assert failure with zero-sized type

### LINKED ISSUE #701: Assertion error on ObjectSubclass::get_instance() with zero sized type
If a zero-sized type implements `ObjectSubclass`, calling `get_instance()` errors:

```
thread 'main' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`', /home/ian/src/glib-rs/src/subclass/types.rs:300:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This assert is failing:

https://github.com/gtk-rs/glib/blob/0cc0a46c6da1cd7c853cef8f6851d80d4a0de202/src/subclass/types.rs#L300

Removing the assertion fixes it. I don't know if there's a good reason to have the assertion. If it's to verify the offset has been set, perhaps it should use a sentinel value other than 0.

I came across this just testing the API, and the struct I actually was going to use wasn't zero-sized. But there are probably some sane uses for a GObject with no data members (except those in a parent type).

ids1024:
Also adds a call to `.get_instance()` to a test (which fails without the other change).

Fixes #701

### DIFF
--- a/src/subclass/object.rs
+++ b/src/subclass/object.rs
@@ -577,6 +577,7 @@ mod test {
         // itself. No private/impl data is allocated for zero-sized types.
         let imp = ChildObject::from_instance(&obj);
         assert_eq!(imp as *const _ as *const (), obj.as_ptr() as *const _);
+        assert_eq!(obj, imp.get_instance());
     }
 
     #[test]
--- a/src/subclass/types.rs
+++ b/src/subclass/types.rs
@@ -297,7 +297,6 @@ pub trait ObjectSubclass: Sized + 'static {
             assert_ne!(type_, Type::Invalid);
 
             let offset = -data.as_ref().private_offset;
-            assert_ne!(offset, 0);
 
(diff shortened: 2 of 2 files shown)


src/subclass/object.rs
```diff
@@ -577,6 +577,7 @@ mod test {
         // itself. No private/impl data is allocated for zero-sized types.
         let imp = ChildObject::from_instance(&obj);
         assert_eq!(imp as *const _ as *const (), obj.as_ptr() as *const _);
+        imp.get_instance();
```
sdroege, line 580:
```suggestion
        assert_eq!(obj, imp.get_instance());
```

Maybe? :)

sdroege (commented):
Looks good to me otherwise, thanks!

sdroege (approved):

PR #704, opened 2020-10-14
Use :ty rather than :path for ffi_name and ffi_class_name

ids1024:
I believe this is more correct, and should allow more things to work without breaking anything. But it's hard to be sure with complicated macros...

This allows code like this to compile:

```rust
glib_wrapper! {
    pub struct SimpleObject(
        Object<(<SimpleObjectPrivate as ObjectSubclass>::Instance),
        (<SimpleObjectPrivate as ObjectSubclass>::Class), SimpleObjectClass>)
        @extends gtk::Widget;

    match fn {
        get_type => || SimpleObjectPrivate::get_type().to_glib(),
    }
}
```

### DIFF
--- a/src/boxed.rs
+++ b/src/boxed.rs
@@ -16,7 +16,7 @@ use translate::*;
 /// Wrapper implementations for Boxed types. See `glib_wrapper!`.
 #[macro_export]
 macro_rules! glib_boxed_wrapper {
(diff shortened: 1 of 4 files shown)


sdroege:
As everything apparently still compiles, let's get this in. The reason why this was needed is apparently not valid anymore :) Thanks for noticing this!

PR #705, opened 2020-10-15
Add a ObjectSubclass version of glib_wrapper!

ids1024:
This makes it easier to define a wrapper for a type defined in Rust using `ObjectSubclass`. It also updates the `glib::subclass` documentation to use this.

I'm not all that fond of the `glib_wrapper!` macro, and would be happy to see something better, but this seems like a good way to improve subclassing ergonomics for now.

There's certainly more room for improvement in both the API and documentation for subclassing.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -758,7 +758,7 @@ macro_rules! glib_object_wrapper {
             }
 
             fn as_ptr(&self) -> *mut Self::GlibType {
-                self.0.to_glib_none().0 as *mut _
+                $crate::translate::ToGlibPtr::to_glib_none(&self.0).0 as *mut _
             }
         }
 
@@ -1072,7 +1072,7 @@ macro_rules! glib_object_wrapper {
                 }
 
                 // And take the reference to the object from above to pass it to the caller
-                Option::<$name>::from_glib_full(obj as *mut $ffi_name).map(|o| $crate::object::Cast::unsafe_cast(o))
+                <Option::<$name> as $crate::translate::FromGlibPtrFull<*mut $ffi_name>>::from_glib_full(obj as *mut $ffi_name).map(|o| $crate::object::Cast::unsafe_cast(o))
             }
         }
 
--- a/src/subclass/mod.rs
+++ b/src/subclass/mod.rs
@@ -91,7 +91,7 @@
 //! // This is the struct containing all state carried with
 //! // the new type. Generally this has to make use of
 //! // interior mutability.
-//! pub struct SimpleObject {
+//! pub struct SimpleObjectPrivate {
 //!     name: RefCell<Option<String>>,
 //!     animal: Cell<Animal>,
 //!     flags: Cell<MyFlags>,
@@ -100,7 +100,7 @@
 //! // ObjectSubclass is the trait that defines the new type and
 //! // contains all information needed by the GObject type system,
 //! // including the new type's name, parent type, etc.
-//! impl ObjectSubclass for SimpleObject {
+//! impl ObjectSubclass for SimpleObjectPrivate {
 //!     // This type name must be unique per process.
 //!     const NAME: &'static str = "SimpleObject";
 //!
@@ -137,7 +137,7 @@
 //! }
 //!
 //! // Trait that is used to override virtual methods of glib::Object.
-//! impl ObjectImpl for SimpleObject {
+//! impl ObjectImpl for SimpleObjectPrivate {
 //!     // Called whenever a property is set on this instance. The id
 //!     // is the same as the index of the property in the PROPERTIES array.
 //!     fn set_property(&self, _obj: &glib::Object, id: usize, value: &glib::Value) {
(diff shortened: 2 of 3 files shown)


ids1024:
CI passing except for `warning: spurious network error (1 tries remaining): [6] Couldn't resolve host name (Could not resolve host: static.crates.io)` on one of the macOS builds.

src/object.rs
```diff
@@ -758,6 +758,7 @@ macro_rules! glib_object_wrapper {
             }
 
             fn as_ptr(&self) -> *mut Self::GlibType {
+                use $crate::translate::ToGlibPtr;
                 self.0.to_glib_none().0 as *mut _
```
sdroege, line 762:
```suggestion
                $crate::translate::ToGlibPtr::to_glib_none(&self.0).0 as *mut _
```

And the same below. Otherwise one could get conflicts if for some reason another trait with `to_glib_none()` was in scope :) Unlikely but easy to get around

src/subclass/mod.rs
```diff
@@ -91,7 +91,7 @@
 //! // This is the struct containing all state carried with
 //! // the new type. Generally this has to make use of
 //! // interior mutability.
-//! pub struct SimpleObject {
+//! pub struct SimpleObjectPrivate {
```
sdroege, line 94: I'm not a fan of this pattern. We have namespaces in Rust, so let's use those :)

I usually put this into an internal `imp` module, and the public part outside of it. Like in the [`listbox_model.rs`](https://github.com/gtk-rs/examples/blob/pending/src/bin/listbox_model.rs) example, which you could update with these macro changes too

src/subclass/mod.rs
```diff
@@ -189,9 +189,23 @@
 //!     }
 //! }
 //!
-//! pub fn main() {
+//! // Optionally, define a wrapper type to make it more ergononic to use from Rust
```
sdroege, line 192:
```suggestion
//! // Optionally, define a wrapper type to make it more ergonomic to use from Rust
```

src/wrapper.rs
```diff
@@ -428,6 +428,47 @@ macro_rules! glib_wrapper {
...
+            @extends [$($extends),+], @implements []);
+    };
+
+    // ObjectSubclass, parents and interfaces
+    (
+        $(#[$attr:meta])*
+        pub struct $name:ident(ObjectSubclass<$subclass:ty, $rust_class_name:ident>) @extends $($extends:path),+, @implements $($implements:path),+;
+    ) => {
+        $crate::glib_object_wrapper!(@object [$($attr)*] $name, <$subclass as $crate::subclass::types::ObjectSubclass>::Instance, <$subclass as $crate::subclass::types::ObjectSubclass>::Class, $rust_class_name,
+            @get_type $crate::wrapper::object_subclass_get_type::<$subclass>(),
```
sdroege, line 468: Why not
```suggestion
            @get_type || $crate::translate::ToGlib::to_glib(<$subclass as $crate::subclass::types::ObjectSubclass>::get_type()),
```

That way you don't need to introduce private hidden API

src/subclass/mod.rs
ids1024: I used the same style as https://github.com/gtk-rs/examples/blob/master/src/bin/basic_subclass.rs (and some other places), but I suppose that would work too.

src/object.rs
ids1024: Good point; I hadn't though of such a conflict. Probably a good idea to prevent that, though it indeed seems unlikely.

src/subclass/mod.rs
```diff
@@ -189,9 +189,23 @@
...
+//! glib_wrapper! {
+//!     pub struct SimpleObject(ObjectSubclass<SimpleObjectPrivate, SimpleObjectClass>);
```
sdroege, line 194: I think this is indeed an improvement. The main problem that is left is that users will have to list all the parent classes and implemented interfaces nonetheless, but I guess that's just how it is :)

Instead of `ObjectSubclass` there might be a better name, not sure. The same thing would also work for new interfaces defined in Rust. I can't think of anything better though :)

src/subclass/mod.rs
sdroege: Yeah I meant to change that one for a long time already. Thanks for the reminder :)

src/subclass/mod.rs
ids1024: The implementation assumes the type implements `ObjectSubclass`, so it seems like an accurate enough name. But if anyone things of something better, that's fine.

I don't really like the `glib_wrapper!` macro. Of course complicated `macro_rules!` macros end up with macro hygiene issues, an incoherent implementation, and confusing error messages. But I also don't like the fact the macro is used by defining a tuple struct containing a certain 'type', which isn't actually a type and isn't actually the representation of the resulting struct...

But coming up with a better API isn't exactly simple either...

sdroege: > The implementation assumes the type implements ObjectSubclass, so it seems like an accurate enough name. But if anyone things of something better, that's fine.

That's a good point, let's keep it like that then. For interfaces we can add another variant later :)

> I don't really like the glib_wrapper! macro.

Me neither but I couldn't come up with anything better yet... so if you have any ideas go ahead

src/subclass/mod.rs
```diff
@@ -55,143 +55,161 @@
...
+//! }
 //!
-//!     // Called right after construction of the instance.
-//!     fn constructed(&self, obj: &glib::Object) {
-//!         // Chain up to the parent type's implementation of this virtual
-//!         // method.
-//!         self.parent_constructed(obj);
+//! // Optionally, define a wrapper type to make it more ergonomic to use from Rust
+//! glib_wrapper! {
+//!     pub struct SimpleObject(ObjectSubclass<imp::SimpleObject, SimpleObjectClass>);
```
sdroege, line 198:
```suggestion
//!     pub struct SimpleObject(ObjectSubclass<imp::SimpleObject, imp::SimpleObjectClass>);
```

shouldn't it be like that?

sdroege: No, because this is the name that is given to the wrapper (confusingly). The actual class struct is taken from `ObjectSubclass::Class`

PR #708, opened 2020-10-17
Regenerate with latest gir and trust upstream nullable annotations for return values

sdroege:
See https://github.com/gtk-rs/gir/pull/970 and https://github.com/gtk-rs/gir/pull/972

CC @sophie-h and @GuillaumeGomez


sdroege:
gio is blocked by https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1708 , I'm not going to add overrides in `Gir.toml` for all these.

GuillaumeGomez (approved):

PR #710, opened 2020-10-19
Remove glib::TimeVal

sdroege:
It's equivalent to std::time::SystemTime except that it's not year-2038-safe


GuillaumeGomez (approved):
Thanks!

PR #713, opened 2020-10-22
Remove PhantomData from struct created in glib_object_wrapper!

ids1024:
I don't believe these any reason to use `PhantomData` in a struct like this that isn't generic.

`glib` itself, at least, still compiles and passes tests with this simply removed.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -724,7 +724,7 @@ macro_rules! glib_object_wrapper {
         // types. Due to inheritance and up/downcasting we must implement these by pointer or
         // otherwise they would potentially give differeny results for the same object depending on
         // the type we currently know for it
-        #[derive(Clone, Hash, Ord)]
+        #[derive(Clone, Hash, PartialOrd, Ord, PartialEq, Eq, Debug)]
         pub struct $name($crate::object::ObjectRef);
 
         #[doc(hidden)]
@@ -1031,30 +1031,6 @@ macro_rules! glib_object_wrapper {
             fn static_type() -> $crate::types::Type {
                 #[allow(unused_unsafe)]
                 unsafe { $crate::translate::from_glib($get_type_expr) }
-            }
-        }
-
-        impl<T: $crate::object::ObjectType> ::std::cmp::PartialEq<T> for $name {
-            #[inline]
-            fn eq(&self, other: &T) -> bool {
-                $crate::translate::ToGlibPtr::to_glib_none(&self.0).0 == $crate::translate::ToGlibPtr::to_glib_none($crate::object::ObjectType::as_object_ref(other)).0
-            }
-        }
-
-        impl ::std::cmp::Eq for $name { }
-
-        impl<T: $crate::object::ObjectType> ::std::cmp::PartialOrd<T> for $name {
-            #[inline]
-            fn partial_cmp(&self, other: &T) -> Option<::std::cmp::Ordering> {
-                $crate::translate::ToGlibPtr::to_glib_none(&self.0).0.partial_cmp(&$crate::translate::ToGlibPtr::to_glib_none($crate::object::ObjectType::as_object_ref(other)).0)
-            }
-        }
-
-        impl ::std::fmt::Debug for $name {
-            fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-                f.debug_struct(stringify!($name))
-                    .field("inner", &self.0)
-                    .finish()
             }
         }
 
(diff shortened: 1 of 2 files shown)


ids1024:
Am I missing something, or would deriving `PartialEq`, `Eq`, `PartialOrd`, and `Ord` have the desired behavior, since it will pass them on to the wrapped `ObjectRef` type? It would be cleaner if they don't have to be explicitly implemented.

Similarly, it looks like the main reason the explict `Debug` is needed is so it won't display the `PhantomData` member.

src/object.rs
```diff
@@ -725,7 +725,7 @@ macro_rules! glib_object_wrapper {
         // otherwise they would potentially give differeny results for the same object depending on
         // the type we currently know for it
         #[derive(Clone, Hash, Ord)]
-        pub struct $name($crate::object::ObjectRef, ::std::marker::PhantomData<$ffi_name>);
+        pub struct $name($crate::object::ObjectRef);
```
sdroege, line 728: That would also apply to `src/shared.rs` and `src/boxed.rs` then and maybe other uses of `PhantomData` here.

I can't remember exactly why this was added (the commit message that added it should have the details) but it was something about variance.

sdroege:
> Similarly, it looks like the main reason the explict `Debug` is needed is so it won't display the `PhantomData` member.

That's also needed to additional print the type name :)

src/object.rs
ids1024: It looks like this `PhantomData` dates from https://github.com/gtk-rs/glib/commit/15578ea0fafa53515463f86f104d1d0e5a59f15c.

Looking at the side by side diff, it seems `glib_object_wrapper!` replaced a generic `pub struct GenericObject<T>(Ref, PhantomData<T>);` type.

As I understand, the `PhantomData` no longer serves any purpose now that it's not a generic type, but I suppose the author of that commit didn't consider this at the time the change was made.

ids1024: `glib_shared_wrapper!` and `glib_boxed_wrapper!` don't include `PhantomData` in the the structs they define. `PhantomData` is used by `shared.rs` and `boxed.rs` for a couple generic types, but it's probably actually appropriate there.

So I think this change should be correct, and I don't see other obviously unnecessary uses of `PhantomData`.

ids1024:
> That's also needed to additional print the type name :)

Current `Debug` implementation:

```
SimpleObject { inner: ObjectRef { inner: 0x55bf0f179000, type: SimpleObject } }
```

Derived implementation:

```
SimpleObject(ObjectRef { inner: 0x56013db81000, type: SimpleObject })
```

So it looks like deriving debug is fine here unless printing it as a tuple struct is undesirable. (And if the first is preferable, the struct should probably just be defined as a struct with a `inner` member).

ids1024:
Since it appears to be correct, I've pushed a commit deriving `PartialOrd`, `PartialEq`, `Eq`, and `Debug`, as I suggested earlier.

src/object.rs
sdroege: In shared/boxed that phantom data could be removed by changing the memory manager to a ZST instead of an uninhabitated/uninstantiable type. But doesn't really matter :)

This one can go away though, it's basically the same case as here or not?

https://github.com/gtk-rs/glib/blob/master/src/value.rs#L137

sdroege (approved):

src/object.rs
ids1024: Given the comment above that line:

```
// TODO: Should use impl !Send for Value {} once stable
```

Perhaps the `PhantomData` is meant as a hack to prevent `Value` from implementing `Send`? But testing it, it seems it doesn't even with `PhantomData` removed (due to the type of `GValue`) so it shouldn't be needed.

sdroege: Yeah that comment is unrelated. Because `GValue` contains a pointer it won't implement `Send` anyway.

sdroege:
The relevant documentation is btw https://doc.rust-lang.org/nomicon/phantom-data.html

src/object.rs
```diff
@@ -1035,30 +1034,6 @@ macro_rules! glib_object_wrapper {
             }
         }
 
-        impl<T: $crate::object::ObjectType> ::std::cmp::PartialEq<T> for $name {
```
sdroege, line 1038: This was actually needed :) It allows comparing e.g. a `gtk::Widget` to a `gtk::Window`. Now it's only possible to compare if they're of the same type.

Same thing below with `PartialOrd`.

This breaks applications now.

PR #717, opened 2020-10-24
Don't derive PartialOrd/PartialEq implementations in glib_wrapper!

ids1024:
Should fix https://github.com/gtk-rs/glib/pull/713/files#r511494658

By implementing it to invoke the traits on the wrapped `ObjectRef` the implementation shorter and clearer than before.


ids1024:
I was going to also add something like `trait ObjectType: for<T: ObjectType> PartialEq<T>`, instead of the current `trait ObjectType: PartialEq`. But that doesn't appear to be valid syntax.

sdroege:
Thanks!

PR #718, opened 2020-10-25
translate: take advantage of Rust Option & Result types

fengalin:
This commit introduces `OptionToGlib` and `TryFromGlib` to help with translations from and to `Glib` types which encode a value with a sentinel to indicate an undefined value and/or for which some values are invalid.

These traits allow auto-implementing `Option<T>::to_glib`, `Option<T>::from_glib` or `Result::<Option<T>, _>::from_glib` from dependent crates when appropriate despite the orphan rule.

Required for https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/issues/234.

Note: I removed the existing fallible implementations of `FromGlib` for `Option<T>` (e.g. `i32` to `u32`). I believe we should now use the auto-implemented `Result<T, InvalidError>` and update the caller where appropriate instead of silently returning `None`, which is not the same IMHO. See also the comment for `impl FromGlib<u32> for char`. If you agree, I'll update code in the affected crates in the `gtk-rs` and `gstreamer` organizations. Meanwhile, CI will fail.

### DIFF
--- a/src/translate.rs
+++ b/src/translate.rs
@@ -20,6 +20,71 @@
 //!     }
 //! ```
 //!
+//! Implementing `OptionToGlib` on a Rust type `T` allows specifying a sentinel to indicate
+//! a `None` value and auto-implementing `FromGlib` for `Option<T>`, which would not be
+//! possible in dependent crates due to the [orphan rule](https://doc.rust-lang.org/book/ch10-02-traits.html#implementing-a-trait-on-a-type).
+//! In the example below, `ToGlib` is auto-implemented for `Option<SpecialU32>`.
+//!
+//! ```
+//! # use glib::translate::*;
+//! struct SpecialU32(u32);
+//! impl ToGlib for SpecialU32 {
+//!     type GlibType = libc::c_uint;
+//!     fn to_glib(&self) -> libc::c_uint {
+//!         self.0 as libc::c_uint
+//!     }
+//! }
+//! impl OptionToGlib for SpecialU32 {
+//!     const GLIB_NONE: Self::GlibType = 0xFFFFFF;
+//! }
+//! ```
+//!
+//! In order to auto-implement `FromGlib` for `Option<SpecialU32>`, proceed as follows:
+//!
+//! ```
+//! # use glib::translate::*;
+//! # struct SpecialU32(u32);
+//! # impl ToGlib for SpecialU32 {
+//! #     type GlibType = libc::c_uint;
+//! #     fn to_glib(&self) -> libc::c_uint {
+//! #         self.0 as libc::c_uint
+//! #     }
+//! # }
+//! # impl OptionToGlib for SpecialU32 {
+//! #     const GLIB_NONE: Self::GlibType = 0xFFFFFF;
+//! # }
+//! impl TryFromGlib<libc::c_uint> for SpecialU32 {
+//!     type Error = NoneError;
+//!     fn try_from_glib(val: libc::c_uint) -> Result<Self, NoneError> {
+//!         if val == SpecialU32::GLIB_NONE {
+//!             return Err(NoneError);
+//!         }
+//!         Ok(SpecialU32(val as u32))
+//!     }
+//! }
+//! ```
+//!
+//! The `TryFromGlib` trait can also be implemented when the Glib type range is larger than the
+//! target Rust type's range. In the example below, the Rust type `U32` can be built from a signed
+//! `libc::c_long`, which means that the negative range is not valid.
+//!
+//! ```
+//! # use std::convert::TryFrom;
+//! # use std::num::TryFromIntError;
+//! # use glib::translate::*;
+//! struct U32(u32);
+//! impl TryFromGlib<libc::c_long> for U32 {
+//!     type Error = TryFromIntError;
+//!     fn try_from_glib(val: libc::c_long) -> Result<Self, TryFromIntError> {
+//!         Ok(U32(u32::try_from(val)?))
+//!     }
+//! }
+//! ```
+//!
+//! Finally, you can define `TryFromGlib` with both `None` and `Invalid` alternatives by setting
+//! the associated `type Error = NoneOrInvalidError<I>` (where `I` is the `Error` type when the
+//! value is invalid), which results in auto-implementing `FromGlib` for `Result<Option<T>, I>`.
+//!
 //! `ToGlibPtr`, `FromGlibPtrNone`, `FromGlibPtrFull` and `FromGlibPtrBorrow` work on `gpointer`s
 //! and ensure correct ownership of values
 //! according to [Glib ownership transfer rules](https://gi.readthedocs.io/en/latest/annotations/giannotations.html).
@@ -68,10 +133,12 @@
 use glib_sys;
 use libc::{c_char, size_t};
 use std::char;
-use std::cmp::Ordering;
+use std::cmp::{Eq, Ordering, PartialEq};
 use std::collections::HashMap;
+use std::error::Error;
 use std::ffi::{CStr, CString};
 use std::ffi::{OsStr, OsString};
+use std::fmt;
 use std::mem;
 #[cfg(not(windows))]
 use std::os::unix::prelude::*;
@@ -273,6 +340,31 @@ impl ToGlib for Ordering {
             Ordering::Less => -1,
             Ordering::Equal => 0,
             Ordering::Greater => 1,
+        }
+    }
+}
+
+/// A Rust type `T` for which `Option<T>` translates to the same glib type as T.
+// The copy bound on `Self::GlibType` is requested by clippy by virtue of:
+// https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
+pub trait OptionToGlib: ToGlib
+where
+    Self::GlibType: Copy,
+{
+    const GLIB_NONE: Self::GlibType;
+}
+
+impl<T: OptionToGlib> ToGlib for Option<T>
+where
+    T::GlibType: Copy,
+{
+    type GlibType = T::GlibType;
+
+    #[inline]
+    fn to_glib(&self) -> Self::GlibType {
+        match self {
+            Some(t) => t.to_glib(),
+            None => T::GLIB_NONE,
         }
     }
 }
@@ -1081,13 +1173,6 @@ impl FromGlib<glib_sys::gboolean> for bool {
     }
 }
 
-impl FromGlib<u32> for char {
-    #[inline]
-    fn from_glib(val: u32) -> char {
-        char::from_u32(val).expect("Valid Unicode character expected")
-    }
-}
-
 impl FromGlib<i32> for Ordering {
     #[inline]
     fn from_glib(val: i32) -> Ordering {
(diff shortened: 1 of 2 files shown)


src/translate.rs
```diff
@@ -1081,56 +1158,119 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
-        }
+    fn try_from_glib(val: i32) -> Result<Self, InvalidError> {
+        u32::try_from(val).map_err(|_| InvalidError)
     }
 }
 
-impl FromGlib<i64> for Option<u64> {
+// FIXME should we implement this here?
+// Maybe we should force the caller to use try_from_glib instead.
+impl FromGlib<u32> for char {
```
GuillaumeGomez: If it can panic, then I think it'd be better to force the usage of `try_from_glib` as you suggested. Maybe let's wait for @sdroege's opinion too before changing anything.

GuillaumeGomez:
Looks like a nice improvement overall!

src/translate.rs
sdroege: I think this will need changes in the code generator and possibly manual code calling it though. The code generator does not know about fallible `from_glib()` and there is no such information in the .gir files, so that seems a bit annoying but we can hardcode that information for `u32` -> `char`

fengalin:
Additional proposal: for signed to unsigned conversions, we could use [`TryFromIntError`](https://doc.rust-lang.org/std/num/struct.TryFromIntError.html) instead of `InvalidError`. Maybe `InvalidError` is too broad and not necessary since it doesn't allow for auto-implementations anyway.

src/translate.rs
fengalin: Yeah, that's planned :) 
I intend to detect the `try_from_glib` trait impl. I didn't look into it yet though.

sdroege:
> Additional proposal: for signed to unsigned conversions, we could use [`TryFromIntError`](https://doc.rust-lang.org/std/num/struct.TryFromIntError.html) instead of `InvalidError`. Maybe `InvalidError` is too broad and not necessary since it doesn't allow for auto-implementations anyway.

Why do we have `from_glib()` conversions for those anyway (and I doubt they're used widely)? I would remove those and only keep `from_glib()` / `to_glib()` for actual GLib types where this makes sense.

src/translate.rs
sdroege: For `char` it could be hardcoded, for other types (`gst::ClockTime`) you'd have to add a configuration for it

fengalin:
> Why do we have `from_glib()` conversions for those anyway (and I doubt they're used widely)? I would remove those and only keep `from_glib()` / `to_glib()` for actual GLib types where this makes sense.

I guess that was implemented to handle the cases where the glib type was a signed so that negatives would indicate that the value is invalid (e.g. an error code in a return value) but the actual usable value was positive.

sdroege:
> I guess that was implemented to handle the cases where the glib type was a signed so that negatives would indicate that the value is invalid (e.g. an error code in a return value) but the actual usable value was positive.

I believe those can go away. I can't remember a single usage of that :)

fengalin:
> I believe those can go away. I can't remember a single usage of that :)

Good! :)

src/translate.rs
```diff
@@ -277,6 +337,23 @@ impl ToGlib for Ordering {
     }
 }
 
+/// A Rust type `T` for which `Option<T>` translates to the same glib type as T.
+pub trait OptionToGlib: ToGlib {
+    fn glib_none() -> Self::GlibType;
```
sdroege: Why not an associated constant?

src/translate.rs
```diff
@@ -1081,56 +1158,119 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
+    fn try_from_glib(val: i32) -> Result<Self, InvalidError> {
+        u64::try_from(val).map_err(|_| InvalidError)
+    }
+}
+
+/// Error type for `TryFromGlib` when the Glib value can be undefined or invalid.
+#[derive(Debug, PartialEq, Eq)]
+pub enum UndefinedOrInvalidError {
+    Invalid,
+    Undefined,
```
sdroege: In the other direction you call it "none". That should be consistent. So either `None` here, other `UndefinedToGlib` for the trait. I prefer the "none" naming :)

sdroege (changes requested):
Looks good to me apart from those comments, thanks!

src/translate.rs
fengalin: Yes! This is obviously the intent here.

src/translate.rs
fengalin: I'm going with None everywhere!

fengalin:
I'm considering removing the `InvalidError` type and let the user define their own type for `TryFromGlib` in this particular cases (e.g.: [`TryFromIntError`](https://doc.rust-lang.org/std/num/struct.TryFromIntError.html), [`CharTryFromError`](https://doc.rust-lang.org/std/char/struct.CharTryFromError.html), ...). I would also remove `UndefineOrInvalidError` in favor of an `Either<NoneError, E>` where `E: Error`.

(More context on the above: initially, I thought of `InvalidError` to auto-implement a `Result<T, InvalidError>` (in a similar way as `NoneError` (formerly `UndefinedError`)), but it didn't make sense since `try_from_glib` already returns `Result<T, InvalidError>`. IMO auto-implementing `FromGlib` for `Result<Option<T>, E>` when `TryFromGlib::<T>::Error` matches `Either<NoneError, E: Error>` is still valuable since that is different from `try_from_glib` which returns `Result<T, Either<NoneError, E>>`.)

For the `u32` to `char` conversion, there is [TryFrom<u32> for char](https://doc.rust-lang.org/std/primitive.char.html#impl-TryFrom%3Cu32%3E) in `std` since rustc 1.34.0. There's also a nightly [`from_u32(i: u32) -> Option<char>`](https://doc.rust-lang.org/std/primitive.char.html#method.from_u32), but I don't really see the point in returning an `Option` rather than a `Result`. So in `glib::translate` I would remove:

- `impl TryFromGlib<u32> for char`.
- `impl FromGlib<u32> for char`, which either forces functions which use this to return a `Result<_, CharTryFromError>` (or an `Option`, but I don't like it) or `expect` in the wrapper body, which is the current one size fit all behaviour.

What do you think?

sdroege:
Sounds good to me but @GuillaumeGomez will probably be unhappy if you add `either` as a dependency (fine by me though).

fengalin:
> @GuillaumeGomez will probably be unhappy if you add either as a dependency (fine by me though).

In that case, I can make a new `enum` for this.

GuillaumeGomez:
I confirm: not happy! :< (but debatable depending on the amount of required code)

fengalin:
> I confirm: not happy! :< (but debatable depending on the amount of required code)

Ok! I'll push an update shortly with a local `enum` so you can tell.

fengalin:
No need for `Either`: `NoneOrInvalidError` seems simple enough IMHO.

I'll update gir and take a look at the affected crates tomorrow: Pango & GTK showed up from a rapid check.

GuillaumeGomez:
Hey @fengalin ! We migrated the glib crate into gtk-rs/gtk-rs repository. Could you please reopen your PR there? :)

src/translate.rs
```diff
@@ -1081,56 +1173,113 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
-impl FromGlib<u32> for Option<char> {
+/// Translate from a Glib type which can result in an undefined and/or invalid value.
+pub trait TryFromGlib<G>: Sized {
+    type Error;
+    fn try_from_glib(val: G) -> Result<Self, Self::Error>;
+}
+
+/// Error type for `TryFromGlib` when the Glib value is None.
+#[derive(Debug, PartialEq, Eq)]
+pub struct NoneError;
```
sdroege, line 1191: These generic error names might be problematic with code using `use glib::translate::*` as exists quite a lot. There are `NoneError` also elsewhere.

But maybe this is not a problem in practice.

src/translate.rs
```diff
@@ -1081,56 +1173,113 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
+
+/// Error type for `TryFromGlib` when the Glib value can be None or invalid.
+#[derive(Debug)]
+pub enum NoneOrInvalidError<I> {
+    Invalid(I),
+    None,
+}
+
+impl<I> NoneOrInvalidError<I> {
+    /// Builds the `None` variante.
```
sdroege, line 1216:
```suggestion
    /// Builds the `None` variant.
```

src/translate.rs
```diff
@@ -1081,56 +1173,113 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
+    pub fn none() -> Self {
+        NoneOrInvalidError::None
+    }
+
+    /// Returns `true` if `self` is the `None` variante.
```
sdroege, line 1221:
```suggestion
    /// Returns `true` if `self` is the `None` variant.
```

src/translate.rs
```diff
@@ -1081,56 +1173,113 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
+    // FIXME uncomment when CI can upgrade to 1.47.1
+    //#[allow(clippy::match_like_matches_macro)]
+    pub fn is_none(&self) -> bool {
+        match self {
+            NoneOrInvalidError::None => true,
+            _ => false,
+        }
+    }
+
+    /// Returns `true` if `self` is the `Invalid` variante.
```
sdroege, line 1232:
```suggestion
    /// Returns `true` if `self` is the `Invalid` variant.
```

src/translate.rs
```diff
@@ -1081,56 +1173,113 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
+    // FIXME `matches!` was introduced in rustc 1.42.0, current MSRV is 1.40.0
+    // FIXME uncomment when CI can upgrade to 1.47.1
+    //#[allow(clippy::match_like_matches_macro)]
+    pub fn is_invalid(&self) -> bool {
+        match self {
+            NoneOrInvalidError::None => false,
+            _ => true,
```
sdroege, line 1239: Would seem more future-proof to check for the `Invalid` variant and have the `_` case be `false`

src/translate.rs
```diff
@@ -1081,56 +1173,113 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
-impl FromGlib<i64> for Option<u64> {
-    #[inline]
-    fn from_glib(val: i64) -> Option<u64> {
-        if val >= 0 {
-            Some(val as u64)
-        } else {
-            None
+impl<I: Error> Error for NoneOrInvalidError<I> {}
+
+impl<I: PartialEq> PartialEq for NoneOrInvalidError<I> {
```
sdroege, line 1261: Isn't this the same as the derived `PartialEq`?

sdroege: And `Eq` below

sdroege (commented):
Looks good to me otherwise

src/translate.rs
fengalin: Spoiled by french :)

src/translate.rs
fengalin: Sure!

src/translate.rs
fengalin: I tried prefixing with `TryFromGlib`. See https://github.com/fengalin/gtk-rs/commit/c47eae5a1d1910c30788343fc7c718e7fdb3cf69

src/translate.rs
fengalin: The intention was to only implement these if `I` also implemented them. I didn't want to force those traits on `I`.

src/translate.rs
fengalin: Maybe a bit verbose. `Translate` would be shorter, but less informative.

sdroege: Or we just leave it unprefixed until it actually becomes a problem

src/translate.rs
sdroege: That's what the default derive for them is doing :)

fengalin: Oh, I didn't know that! :) I thought it would have forced them on `I`. Thanks!

src/translate.rs
fengalin: Hum... I think we should stick with a name now.

sdroege: See above in the initial comment "But maybe this is not a problem in practice." :) If all the names seem ugly, let's just keep it and worry about conflicts if they actually happen. Seems very unlikely in practice.

fengalin: I saw your comment :) But, I also agree it would be better to use a prefix. I also think it's the usual way of naming errors. Here are my proposals (taking the longest affected `struct`):

- `NoneOrInvalidError`
- `TryFromGlibNoneOrInvalidError`
- `TranslateNoneOrInvalidError`
- `GlibNoneOrInvalidError`

Maybe the last one :) What do you think? Also @GuillaumeGomez.

sdroege: I would vote for the last one too, @GuillaumeGomez ?

GuillaumeGomez: No preference so let's go for the last one. :)

fengalin:
Transferred to https://github.com/gtk-rs/gtk-rs/pull/11

Issue #707, opened 2020-10-17
Use the paste crate for building the name of the class struct in glib_wrapper!


sdroege:
See https://crates.io/crates/paste

Currently we require to pass in `FooClass` (or `FooInterface`) to `glib_wrapper!` only for being able to define a struct with that name. That's rather redundant and can be solved by using the `paste` crate.

For that to work we will also not to (hidden) re-export the `paste` macro from glib but that shouldn't be a problem really.

@ids1024 Maybe something you'd be interested to look into?

sdroege:
Also when doing that it would probably make sense to keep the old variants of the macro around for a while to not immediately break all code, and remove them in a few weeks.

It would also need a small update to [gir](https://github.com/gtk-rs/gir/).

sdroege:
@GuillaumeGomez Wanted to write a proc-macro inside glib-macros for this specific use-case, i.e. appending a `Class` to an existing identifier.

ids1024:
It would be nice to not have to specify the class struct name, while also not silently bringing another type into scope. Not that I can think of a good way to do both...

ids1024:
Hm, is there any reason the class struct couldn't be a generic type?

`glib_object_wrapper!` has this case:

```rust
    (@class_impl $name:ident, $ffi_class_name:ty, $rust_class_name:ident) => {
        #[repr(C)] 
        #[derive(Debug)]
        pub struct $rust_class_name($ffi_class_name);
              
        unsafe impl $crate::object::IsClassFor for $rust_class_name {
            type Instance = $name;
        }
                
        unsafe impl Send for $rust_class_name { }
        unsafe impl Sync for $rust_class_name { }
    };
```

Could this be generic with something like `pub struct ClassWrapper<T>(T)`, instantiated as `ClassWrapper<$ffi_class_name>`? I suppose the trouble is that `IsClassFor` couldn't be implemented generically, and the compiler would reject `impl IsClassFor for ClassWrapper<Object>` from another crate... But there's probably some way to adjust the use of types and traits to something that will work.

If this could be made to work, it wouldn't be necessary to specify a new identify (it would just be an instance of a generic).

sdroege:
[...]

PR #720, opened 2020-10-26
In glib_object_wrapper!, use generic Class type instead defining a struct

### LINKED ISSUE #37: Use the paste crate for building the name of the class struct in glib_wrapper!
See https://crates.io/crates/paste

Currently we require to pass in `FooClass` (or `FooInterface`) to `glib_wrapper!` only for being able to define a struct with that name. That's rather redundant and can be solved by using the `paste` crate.

For that to work we will also not to (hidden) re-export the `paste` macro from glib but that shouldn't be a problem really.

@ids1024 Maybe something you'd be interested to look into?

ids1024:
This makes the `$rust_class_name` argument in `glib_wrapper!` optional. If specified, it defines a `type` alias for the class.

This removes the `ObjectType::RustClassType` associated type, since it can't be enforced/assumed by the type system that that is an instance of this generic struct.

The `IsClassFor` trait is also removed, superseded by this generic type.

This also changes the `IsSubclassable` trait to be implemented on the object struct rather than the class struct, since there otherwise would be foreign trait on foreign type errors.

**TODO**: I still need to determine the naming and where this should be exported (should `Class` be called something else? should it be in the prelude?). I'll also have to try updating gtk to use this (if nothing else, the `IsSubclassable` implementations will need to be changed).

Fixes https://github.com/gtk-rs/glib/issues/707.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -52,8 +52,6 @@ pub unsafe trait ObjectType:
     type GlibType: 'static;
     /// type of the FFI Class structure.
     type GlibClassType: 'static;
-    /// type of the Rust Class structure.
-    type RustClassType: 'static;
 
     fn as_object_ref(&self) -> &ObjectRef;
     fn as_ptr(&self) -> *mut Self::GlibType;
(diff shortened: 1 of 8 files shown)


sdroege:
Thanks, I'll go through this tomorrow :)

src/object.rs
```diff
@@ -2570,3 +2429,116 @@ impl<'a> BindingBuilder<'a> {
         }
     }
 }
+
+#[repr(C)]
```
sdroege, line 2433:
```suggestion
#[repr(transparent)]
```

Because that's how you use it :)

src/object.rs
```diff
@@ -2570,3 +2429,116 @@ impl<'a> BindingBuilder<'a> {
...
+            }
+        }
+    }
+}
+
+unsafe impl<T: ObjectType> Send for Class<T> { }
+unsafe impl<T: ObjectType> Sync for Class<T> { }
+
+// This should require Self: IsA<Self::Super>, but that seems to cause a cycle error
+pub unsafe trait SuperClassIs: ObjectType {
```
sdroege, line 2522:
```suggestion
pub unsafe trait ParentClassIs: ObjectType {
```

Parent is the terminology we use everywhere else

src/object.rs
```diff
@@ -1109,92 +1016,43 @@ macro_rules! glib_object_wrapper {
...
-            fn deref(&self) -> &Self::Target {
-                $crate::object::IsClassFor::upcast_ref(self)
-            }
-        }
-
-        impl ::std::ops::DerefMut for $rust_class_name {
-            fn deref_mut(&mut self) -> &mut Self::Target {
-                $crate::object::IsClassFor::upcast_ref_mut(self)
-            }
+        unsafe impl $crate::object::SuperClassIs for $name {
```
sdroege, line 1029: This is now always only implemented for the direct parent class, or for the whole hierarchy?

src/subclass/simple.rs
```diff
@@ -56,7 +56,7 @@ unsafe impl<T: ObjectSubclass> super::types::ClassStruct for ClassStruct<T> {
 }
 
 impl<T: ObjectSubclass> ops::Deref for ClassStruct<T> {
```
sdroege, line 58: Is this still needed? That deref would just go via `Class<T>` now or not?

src/wrapper.rs
```diff
@@ -329,144 +329,156 @@ macro_rules! glib_wrapper {
...
-        pub struct $name:ident(Object<$ffi_name:ty, $rust_class_name:ident>);
+        pub struct $name:ident(Object<$ffi_name:ty $(, $rust_class_name:ident)?>);
 
         match fn {
             get_type => || $get_type_expr:expr,
         }
     ) => {
-        $crate::glib_object_wrapper!(@object [$($attr)*] $name, $ffi_name, ::std::os::raw::c_void, $rust_class_name, @get_type $get_type_expr, @extends [], @implements []);
+        $crate::glib_object_wrapper!(@object [$($attr)*] $name, $ffi_name, ::std::os::raw::c_void, @get_type $get_type_expr, @extends [], @implements []);
+        $( pub type $rust_class_name = $crate::object::Class<$name>; )?
```
sdroege, line 339: I assume we can omit that now and just go everywhere via `Class`? I.e. leave that additional parameter of the macro unused now but leave it there so everything continues to compile for now until the code generator is updated.

sdroege (commented):
Thanks, generally looks good to me

GuillaumeGomez:
Hey @ids1024  ! We migrated the glib crate into gtk-rs/gtk-rs repository. Could you please reopen your PR there? :)

src/wrapper.rs
ids1024: Then code that actually relies on the class type existing will break (until it's updated to use `Class<_>`. For instance, `ObjectClass` is referenced in a couple places.

src/object.rs
ids1024: Only for the direct parent class, which then allows the type to `deref` recursively up to any parent type. Which seems to be how it worked before.

src/object.rs
ids1024: Everywhere except the `$super_name` of the macro, which I was naming it after. :laughing: 

But yeah, that's probably a better term given it refers to the direct parent, not any ancestor type.

src/subclass/simple.rs
ids1024: Hm. Well `ClassStruct` is different from `Class`. I don't see anywhere else `Deref` is implemented for `ClassStruct`, so unless I'm missing something it won't deref without this.

src/wrapper.rs
sdroege: Well let's update those at the same time then :)

src/subclass/simple.rs
sdroege: But where is `simple::ClassStruct` used if not as part of `klass_init()` where you would now pass a `Class<simple::ClassStruct<_>>` or not?

src/wrapper.rs
sdroege: Will be easier on https://github.com/gtk-rs/gtk-rs as that has all repos in one place

src/subclass/simple.rs
sdroege: Basically `simple::ClassStruct` is an FFI class struct currently, so it would seem more consistent to use the `Class` wrapper around it in all code that uses it as such

ids1024: I'm not entirely familar with how the `ClassStruct` and `InstanceStruct` are used.

But `ObjectSubclass::class_init()` seems to take `Self::Class` as an argument, which is something implementing the `ClassStruct` trait, usually `simple::ClassStruct`. Not the new `Class` type.

I don't believe `simple::ClassStruct` implements `ObjectType`, so `Class<simple::ClassStruct<_>>` isn't a valid type anyway.

sdroege: Indeed. Let's keep that for later then to clean this up, but can you create an issue for this?

PR #721, opened 2020-10-28
Fix double registration of GEnums

TobiasMorell:
The `GEnum` derive macro did not check whether an enum exists before registering it, leading to a panic if the same enum is registered twice:

```
GLib-GObject-WARNING **: 08:04:50.392: cannot register existing type 'EnumName'
thread '<unnamed>' panicked at 'assertion failed: `(left != right)`
  left: `<invalid>`,
  right: `<invalid>`', /path/to/source.rs
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This PR makes sure that enums are only registered if `gobject_sys::g_type_from_name` returns `0`.

### DIFF
--- a/glib-macros/src/genum_derive.rs
+++ b/glib-macros/src/genum_derive.rs
@@ -165,7 +165,12 @@ pub fn impl_genum(input: &syn::DeriveInput) -> TokenStream {
 
                 let name = std::ffi::CString::new(#gtype_name).expect("CString::new failed");
                 unsafe {
-                    let type_ = #crate_ident::gobject_sys::g_enum_register_static(name.as_ptr(), VALUES.as_ptr());
+                    // Lookup the type ID or return 0 if it has not yet been registered under the specific name
+                    let mut type_ = #crate_ident::gobject_sys::g_type_from_name(name.as_pr());
+                    if type_ == 0 {
+                        // Register the type ONLY if not done before
+                        type_ = #crate_ident::gobject_sys::g_enum_register_static(name.as_ptr(), VALUES.as_ptr());
+                    }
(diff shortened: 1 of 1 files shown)


glib-macros/src/genum_derive.rs
```diff
@@ -165,7 +165,12 @@ pub fn impl_genum(input: &syn::DeriveInput) -> TokenStream {
 
                 let name = std::ffi::CString::new(#gtype_name).expect("CString::new failed");
                 unsafe {
-                    let type_ = #crate_ident::gobject_sys::g_enum_register_static(name.as_ptr(), VALUES.as_ptr());
+                    // Lookup the type ID or return 0 if it has not yet been registered under the specific name
+                    let mut type_ = #crate_ident::gobject_sys::g_type_from_name(name.as_pr());
+                    if type_ == 0 {
+                        // Register the type ONLY if not done before
```
sdroege, line 171: This assumes that it's exactly the same enum, which is not guaranteed. Might be different versions or other conflicts.

The only safe solution here is to panic.

In your case I assume this happens because the enum registration code is statically linked multiple times into different pieces of code, so it gets registered multiple times. You must only have it exactly once by putting it into a shared library that is linked to all users of it.

TobiasMorell: Thank you for your quick response. I'll try to determine where this static linking happens.

sdroege: If you have this in a crate that is a dependency of multiple GStreamer plugins plus maybe an application, then this would happen. Or if you link multiple different versions of the code with the enum into the same final result.

PR #725, opened 2020-12-06
Implement derive macro for Downgrade/Upgrade traits

andy128k:


glib-macros/src/lib.rs
```diff
@@ -85,3 +88,17 @@ pub fn gflags(attr: TokenStream, item: TokenStream) -> TokenStream {
     let gen = gflags_attribute::impl_gflags(&input, &gtype_name);
     gen.into()
 }
+
+#[proc_macro_derive(Downgrade)]
```
sdroege, line 92: Can you add some documentation here and explain what it's good for?

glib-macros/src/downgrade_struct.rs
```diff
@@ -0,0 +1,44 @@
...
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{Generics, Ident};
+
+pub fn derive_downgrade_for_struct(
+    ident: Ident,
+    generics: Generics,
+    data_struct: syn::DataStruct,
+) -> TokenStream {
+    let weak_ref = format_ident!("{}WeakRef", ident);
```
sdroege, line 11: I'd call it `{}Weak` without the `Ref` as it isn't really a reference in the Rust sense (`&x`)

tests/clone.rs
```diff
@@ -170,3 +171,27 @@ mod import_rename {
         closure(0, 0);
     }
 }
+
+#[test]
+fn derive_downgrade() {
+    #[derive(Downgrade)]
+    pub struct NewType(Object);
```
sdroege, line 178: What's the advantage of the macro over putting your struct into an `Rc`/`Arc`?

glib-macros/src/lib.rs
andy128k: Sure, I just wanted to draft a PR and receive a feedback first.

tests/clone.rs
andy128k: Easier to add methods to your structure.

sdroege: But more expensive as each field has to be handled separately. But for the `NewType` pattern above I would use it quite often personally, the others should probably get a mention about the performance impact in the documentation :)

glib-macros/src/lib.rs
sdroege: I didn't look at the code in detail but generally looks like a good idea and I wanted to have such a thing for the newtype pattern since quite a while.

sdroege:
Can you also submit it to https://github.com/gtk-rs/gtk-rs and close the one here?

tests/clone.rs
andy128k: I also started with `NewType` pattern with single gtk widget in it. Later I found that I want a `struct MyListView<T: ToString>(gtk::Widget, PhantomData<T>)`. So, I decided to generalize macro.

andy128k:
Closing in favor of https://github.com/gtk-rs/gtk-rs/pull/200

PR #728, opened 2021-02-24
Type: Derive Ord and Hash

heftig:
They're useful when `Type`s need to be used in sets or as map keys.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -17,7 +17,7 @@ use std::mem;
 use std::ptr;
 
 /// A GLib or GLib-based library type
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Type {
     /// An invalid `Type` used as error return value in some functions
     Invalid,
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -17,7 +17,7 @@ use std::mem;
 use std::ptr;
 
 /// A GLib or GLib-based library type
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Type {
     /// An invalid `Type` used as error return value in some functions
     Invalid,
```
sdroege, line 23: Looking at this, we should probably define `Type` as `struct Type(ffi::GType)` and have the basic types as associated constants on it instead of making the whole thing an enum. By making it an enum it takes twice the space (due to padding) and most operations have to do two comparisons instead of one (discriminant and value)...

heftig: That would be quite the API change. Do you want it in this PR?

sdroege: No, separate PR please. This one here is fine as is, I just remembered it now :)

sdroege:
@heftig Wrong repository btw. Can you submit a new PR to https://github.com/gtk-rs/gtk-rs/ ? Thanks!

PR #729, opened 2021-02-24
Type: Make leaner, replace enum with assoc consts 

heftig:
As suggested in https://github.com/gtk-rs/glib/pull/728#discussion_r582044041.

(Depends on #728)


sdroege:
@heftig Wrong repository. Can you submit a new PR to https://github.com/gtk-rs/gtk-rs/ ? Thanks!

sdroege:
Looks good to me otherwise

